Issue #79, opened 2026-06-28
Benchmark Phase 1: semi-automated runner over a 3-5 task set


suleimansh:
Child of #75. Builds on Phase 0 (#78). Turn the manual method into a semi-automated runner over a small task set.

## Goal
3 to 5 tasks, run through a runner that launches the agent, detects acceptance, times the run, and logs structured human-intervention events.

## Deliverables
- `runner`: checkout starting commit -> launch agent (GemStack | Next.js adapter) -> stream events -> poll acceptance script -> emit `report.json` `{ framework, task, runIndex, seconds, interventions, status }`.
- Two thin adapters (GemStack / Next.js) so the runner stays framework-agnostic.
- A task set of 3 to 5 tasks from the categories in the #75 design (feature add, schema change, bug fix, AI integration, refactor), each with an acceptance script.
- N runs per task (start at 5); raw `report.json` per run.

## Out of scope
Aggregation/reporting polish and the committed baseline land in Phase 2.

suleimansh:
Scope note: the 3-5 task set this runs is now the **ecosystem** set (#84) over the reframed app pair (#83) — Vike extension family vs idiomatic Next.js. See the direction-change comment on #75. The runner design here is unchanged.

suleimansh:
Relocating the benchmark to its real home: https://github.com/suleimansh/vike-data/issues/330 (suleimansh/vike-data), where the extension family it now measures actually lives, and where it's a joint artifact with Rom. Closing here; tracking continues there. (Issues can't transfer across orgs, hence close-and-recreate.) The Phase 0 method learnings carry over. The stranded `benchmarks/` + `bench-app-*` code on this repo gets removed in a follow-up when the vike-data harness exists.

Issue #78, opened 2026-06-28
Benchmark Phase 0: one task, both apps, manual tally (prove the rubric)


suleimansh:
Child of #75. First concrete step of the "our AI vs Next.js" benchmark: prove the method and the intervention rubric by hand before any automation.

## Goal
Run **one** task on both a GemStack reference app and a functionally equivalent Next.js app, with the same AI agent, and record:
1. Time-to-task (wall clock to acceptance pass).
2. Human-intervention count (per the rubric in the #75 design comment).

## Deliverables
- Two minimal but functionally-equivalent reference apps (or agreement on existing ones to reuse): GemStack side with the orchestration layer in reach, Next.js side vanilla.
- One task: prompt + starting commit + acceptance script (exit 0 = pass) + hard timeout + max-intervention cap.
- A manual run log for each side: seconds + intervention tally + status (pass/DNF).
- A short write-up: did the rubric hold up, what was ambiguous, what to automate next.

## Out of scope
No runner, no aggregator yet. Manual stopwatch and tally only. This phase de-risks the methodology, not the tooling.

Issue #75, opened 2026-06-28
Benchmark: GemStack AI orchestration vs Next.js (task speed + human interventions)


suleimansh:
Idea from the Vike collab (Rom). Measure how "our AI" performs compared to Next.js: the same AI coding agent achieving the same real development tasks, with the GemStack orchestration layer (autopilot + skills + mcp) available versus a vanilla Next.js baseline that has none of it.

## Metrics
1. Time for the AI to achieve a given task.
2. Number of human interventions required to complete it.

## Scope to define
- A fixed task set (representative dev tasks) and a Next.js comparison harness.
- Reproducible runner + reporting (time + intervention count per task).
- A baseline run we can track over time.

## Out of scope
This is not the self-healing loop (an app that watches its own tests/audits/errors and fixes itself). That is a separate direction. This benchmark only measures an AI agent building and changing apps.

Priority is medium; this is a strong story for what the orchestration layer is for, not blocking shipping. See the design spec in the comments.

## Phases
- [ ] #78 Phase 0: one task, both apps, manual tally (prove the rubric)
- [ ] #79 Phase 1: semi-automated runner over a 3-5 task set
- [ ] #80 Phase 2: full suite, aggregator, committed baseline

suleimansh:
## Benchmark design: "our AI" vs Next.js (time-to-task + human interventions)

Scoping the "scope to define" section. This is the design spec only; the runnable harness is a follow-up.

> Note: this benchmark is **not** the self-healing loop. It measures an AI coding agent building and changing apps. Self-healing is a separate direction and is out of scope here.

### What we are actually measuring

Rom's framing: "how our AI performs compared to Next.js." Next.js has no AI orchestration layer, so the honest comparison is **the same AI coding agent achieving the same tasks, with the GemStack orchestration layer available versus not**:

- **GemStack side** ("our AI") - the agent operates with the orchestration layer in reach: `@gemstack/ai-skills` (invokable capabilities), `@gemstack/ai-mcp` / `@gemstack/mcp` (tools and resources the app exposes to the agent), and `@gemstack/ai-autopilot` (supervisor topology for multi-step plan -> dispatch -> synthesize).
- **Next.js baseline** - the same agent, same model, operating on a functionally equivalent Next.js app with none of that layer (vanilla framework + generic file editing).

Two headline metrics, both Rom's:

1. **Time-to-task** - wall-clock from task start to acceptance criteria passing.
[...]

PR #82, opened 2026-06-28
bench: record Phase 0 run for task-001-tags (both apps PASS)

suleimansh:
First end-to-end Phase 0 run of the "our AI vs Next.js" benchmark (#75, #78).

## What this is
The actual measurement run on top of the merged Phase 0 harness. Not new product code, just recorded results plus the agents' solution patches. The baseline apps stay pristine (the Phase 0 starting point); only `benchmarks/runs/` is added.

## Method
- One autonomous AI coding agent per app, same agent type and model both sides (fair-harness rule), same task spec verbatim, no human in the loop.
- Harness validated first: both committed baselines fail the gate identically (same 5 checks) -> a real symmetric gap.
- Independently re-verified by hand against fresh DBs + restarted servers after the agents reported PASS.

## Results
| App | Status | Time | Interventions |
|---|---|---|---|
| bench-app-next (vanilla Next.js) | PASS | 139s | 0 |
| bench-app-gemstack (Vike + @gemstack/ai-sdk) | PASS | 145s | 0 |

All 14 contract checks pass on both. Diffs comparable (~130-150 lines each).

## Reading
The rubric runs end to end and the gate is objective and reproducible, so Phase 0's goal is met. A plain CRUD-extension task does not exercise the orchestration layer, so it does not differentiate here (both fast, zero interventions). The intervention metric will only show signal on AI-integration / multi-step / refactor tasks. Phase 1 (#79) should weight the task set toward those.

See `benchmarks/runs/2026-06-28-task-001-tags/results.md` for the full log and reproduce steps.

### DIFF
--- a/benchmarks/runs/2026-06-28-task-001-tags/results.md
+++ b/benchmarks/runs/2026-06-28-task-001-tags/results.md
@@ -1,0 +1,48 @@
+# Phase 0 run — task-001-tags
+
+Date: 2026-06-28
+Task: [`spec/task-001-tags.md`](../../spec/task-001-tags.md) — add tagging to notes (data model + HTTP contract + UI).
+Gate: [`tasks/task-001-tags/accept.mjs`](../../tasks/task-001-tags/accept.mjs), objective contract check, exit 0 = pass.
+
+This is the first end-to-end Phase 0 run. Its purpose (per [#78](https://github.com/gemstack-land/gemstack/issues/78)) is to prove the method and the rubric are runnable, not to draw a final conclusion.
+
+## Method
+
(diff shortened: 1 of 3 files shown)


suleimansh:
Not merging. The benchmark is relocating to its real home, suleimansh/vike-data, where the extension family it now measures (vike-auth, vike-data, vike-notifications, vike-stripe) actually lives. This Phase 0 run proved the method (harness + rubric work; a plain CRUD task doesn't differentiate) — that learning carries over to the vike-data benchmark. Closing here rather than landing an AI-only run log on a repo the benchmark is leaving.

Issue #545, opened 2026-07-15
[The Framework] Remove meta-select + the architect


suleimansh:
Carrying over @brillout's call on #500, which is closed: **(c), remove the whole meta-select + architect thing**.

Both are agent turns that carry system prompts of their own and run *before* the #326 prompt gets a say. That's the point: they are prompts you didn't design and weren't aware of.

Scope, so the size is visible:

**meta-select** is self-contained. It's the turn that auto-picks a preset from your intent (`autoSelectPreset` in cli.ts, `meta-select.ts`, its bench). Remove it and `--preset` / `the-framework.yml` become the only way a preset gets chosen; a run without one takes the plain flow. Nothing else depends on it.

**The architect** is more tangled. It's the turn that decides the stack, and its structured output feeds things that aren't prompts:
- the decisions ledger + stack rationale the dashboard renders (`run-view.ts`, `RunOverview.tsx`)
- ai-autopilot's bootstrap steps

So cutting the turn also cuts what those panels show, unless the same information comes from somewhere else.

Two calls I'd like from you before I cut:

1. The architect currently answers "what stack?" for a from-scratch build in an empty dir. With it gone, does the #326 prompt own that (the agent just picks, and says so as it builds), or does a build stop being opinionated about the stack at all?
2. The decisions ledger + stack rationale panels are fed by the architect's structured output. Do they go with it, or get re-derived from the build turn?

I'd do it as two PRs, meta-select first since it's clean.

Related: #520 (see the entire prompt), #542.

brillout:
> does a build stop being opinionated about the stack at all?

This. Complex topic, let's postpone IMO. (I think it would require a lot of deep work to make something good.)

> they go with it

This. (I like it, but not a priority.)

Issue #476, opened 2026-07-14
Dashboard: show the exact prompt sent to the agent each turn


suleimansh:
Right now you can't see the prompts The Framework sends to Claude. The event feed renders each turn's start as just `> prompt sent` and drops the text, so evaluating prompt quality means digging through Claude Code's session .jsonl files by hand. "I feel blind."

The full prompt is already carried on the driver `start` event and persisted to events.jsonl, so this is a display-only fix.

## Fix
- Terminal/log formatter: show a one-line preview instead of dropping the prompt.
- Dashboard event feed: render the full prompt for each `start` event in a collapsible block (click to expand), for both live runs and replays.

Issue #326, opened 2026-07-10
System prompt


brillout:
## System prompt

```md
# System prompt

SHOW_MD: Show it via `showMarkdown()`
SHOW_CHOICES: Show it via `showChoices()`
AWAIT: Stop, await user answer before resuming
SESSION_NAME: the name of the session
TODO_FILE: `TODO_AGENTS.md`
ADD_ANALYSIS_ENTRY: Add entry to the ANALYSIS_RESULT.md list

## Analyze the user prompt

Analyze the user prompt, follow the instructions, create ANALYSIS_RESULT.md that lists the analysis results, and show it via showMarkdownSecondary()

### Ambiguous prompt

If it isn't clear what you should do (e.g. unclear scope, unclear user prompt), make a list of interpretations sorted by plausibility, <SHOW_CHOICES>, <AWAIT>

<ADD_ANALYSIS_ENTRY> whether YES/NO the prompt is ambiguous

### Scope

- If the scope of what you'll work on is *large*, create a `PLAN_<SESSION_NAME>.agent.md` of what you'll work on, <SHOW_MD>, <AWAIT>
- If the scope is potentially *very large* (e.g. spans over many hours/days of work), consider adding entries to <TODO_FILE> (backlog of follow-up tasks) and show new entries <SHOW_MD>

<ADD_ANALYSIS_ENTRY> whether the scope is small, large, or very large


## Before starting changes

Do the following before applying your first change.

### Session name

1. Create a <SESSION_NAME> as a string [a-z0-9-]+ that succinctly represents the intention of the user prompt
2. Create a new branch `the-framework/<SESSION_NAME>` and `$ git checkout` it — do all the work in that branch
3. Call setSessionName(<SESSION_NAME>)


## Before applying changes

Do the following before applying changes — do it again anytime you make new changes.

### Alternatives

Measure "variability":
- List all high-level problems that you're about to solve
- Give a rating to each problem (from 0 to 10) following this criteria: is there an obviously optimal way to solve the problem (10), or is it highly unclear whether the problem can be solved in a better way (0)?
- Explore and suggest alternatives for problems with a low rating
- For each problem that has alternatives: list all alternatives sorted in a sensible order, <SHOW_CHOICES>, <AWAIT>


## After applying changes

After you're done, decide: is <SESSION_NAME> finished, with no work left to do?
- Yes: call setReadyForMerge() — required, the work is never merged without it
- No: don't call it; say what's left instead



# User prompt

${{tf.prompt}}
[...]

PR #546, opened 2026-07-15
feat(framework): let the user see the system prompt

### LINKED ISSUE #520: Allow user to see system prompt
I think it's important that the user is able to understand what The Framework does — I think many (if not the vast majority of) early adopters will want to know what The Framework does.

Idea: below the prompt textarea, show a toggle:
>  ▶ See actual prompt sent (see system prompt)

When clicking the toggle, the user sees:
> The Framework wraps your prompt with a so-colled "system prompt" (it's just a prompt wrapping your prompt) in order to enable long-running autonomous agents. 
> [ ] Disable system prompt (caveat: some functionalities might not work)
> [the-actual-prompt-that-will-be-sent]

So `[ ] Disable system prompt` is basically our previous `[ ] Vanilla` idea. (It's just a rename, as I don't think everyone understands what "Vanilla" means.)

Technically, this means the prompt wrapping logic must be loaded on the client-side, I guess that's okay. (It's maybe even a good thing to decouple the prompt wrapping logic from the rest of the code => easily unit testable.)

suleimansh:
Closes #520. @brillout, built to your spec, with one honest limit called out below.

Under the editor: `▶ See actual prompt sent (see system prompt)`. Expanding shows your explainer text, `[ ] Disable system prompt (caveat: some functionalities might not work)`, and the prompt itself. `Vanilla` is gone as a label. The stored key stays `vanilla` (it's load-bearing across the registry, daemon and CLI), so only the words change.

It renders through `composeRunSystem` — the same function the run itself composes with, not a copy — so the toggles are shown doing exactly what they really do. You were right that it decouples nicely: the composition was already pure, and only reading `SYSTEM.md` needed moving out (now `system-prompt-file.ts`).

## The limit worth knowing about

A **prompt** run is fully shown: I verified the preview renders 3,874 chars and a real run reports "system prompt sent (3874 chars)".

A **build** run isn't. It also appends the repo's memory files (whole contents), plus the personas and skills for the detected stack, all read off disk at run time — on a real build that's most of the ~8.8k. And with auto-preset on, part of the prompt depends on a preset that meta-select picks *during* the run, so no pre-run preview can know it. I didn't want a panel labelled "actual prompt sent" quietly showing something else, so the panel says what it doesn't cover.

The exact text is now shown too, just not before the run: the `system-prompt` event has always carried the literal string handed to the agent, and the dashboard was reducing it to `(8852 chars)` and dropping the text. It already reached the browser and we threw it away. It now expands in the event log, same collapsible as the per-turn prompts from #476.

So: the toggles' effect is previewable before you run, and the entire final prompt is readable once you do. If you want the whole thing pre-run too, the clean answer is a server-side preview RPC that does the disk reads — happy to add it, but it can't be exact while meta-select picks a preset mid-run, which is #545.

## Guard

Exporting the composition from `client.ts` naively pulls `node:fs` into the browser bundle, which is the one thing that barrel exists to prevent. A test walks the **compiled** barrel's import graph and fails on any reachable `node:*`. I checked it fails when the leak is reintroduced, rather than trusting it.

589 tests pass.


brillout:
>  It also appends the repo's memory files (whole contents), plus the personas and skills for the detected stack, all read off disk at run time — on a real build that's most of the ~8.8k.

What is this? Can we remove it all?

> And with auto-preset on, part of the prompt depends on a preset that meta-select picks during the run, so no pre-run preview can know it.

Same: what is it? Can we remove it all?

suleimansh:
Answered in #547, with the actual text of a real run rather than my description of it.

Short version: it's three things — personas (3,764 chars), skills (609), project memory (609). Your #326 block is 3,870 of that 8,852, so the personas are as big as your whole prompt.

They're stack opinions appended to every build. A run whose prompt was "a single-file Node script that prints the 10th Fibonacci number" was told to build UI on Vike and to set up Prisma with postgresql, because an empty dir falls back to the Vike preset — its own log says `Detected Vike (confidence 0)`.

And a correction: I told you above that the memory files were most of the 8.8k. They're 7%. I had that wrong.

Yes to removing personas + skills — pure prompt, nothing else reads them. Project memory is half not-a-prompt (DECISIONS.md is written from the decisions ledger at run end), so that part is two decisions, not one. #547 has the detail and the calls I need from you.

Issue #190, opened 2026-07-03
Make The Framework modular (Vike-style extensions)


suleimansh:
Make The Framework modular, similar to Vike: installed `framework-*` packages self-register instead of the CLI hardcoding the list.

## Decisions (settled)

- **Agnostic core.** Nothing is Vike-gated. Capabilities are framework-agnostic.
- **Framework = a skill.** A framework isn't a special adapter package. It's just a skill = a pointer to its `llms.txt` (e.g. https://vike.dev/llms.txt for Vike, the equivalent for Next). The old "adapter" axis is gone.
- **Flat naming.** `framework-sentry`, not `framework-tracking-sentry`.
- **Open Loop / presets / marketplace / modes** moved to #204. This issue stays scoped to the extension SPI.

So the model is two things: **capabilities** (auth, data, ...) that self-register + **skills** (doc pointers for framework/domain knowledge). Skills are the shared unit with #204.

## Where it's hardwired today

`packages/framework/src/run.ts` pins the composition:

```ts
const { preset, detection } = builtinPresetRegistry().select(opts.signals ?? {})
const personas = detection.framework === 'Vike'
  ? presetPersonas(preset, vikeExtensionPersonas)   // fixed list, Vike-gated
  : presetPersonas(preset)
```

The registry and the extension list are pinned in code, so no third party can publish `framework-next` or `framework-drizzle` and have it auto-compose. The substrate is already there (Preset detection + Persona capabilities + registries in `@gemstack/ai-autopilot`), it just isn't exposed.

## What to build

A public `FrameworkExtension` unit (capability) that bundles:

- detection signals (deps/files) that decide when it activates
- persona(s) it contributes to the agent frame
- optional serve / checklist / deploy contributions
- optional skill(s) it pulls in (e.g. a data extension can register the ORM's `llms.txt`)

The CLI discovers installed `framework-*` packages, registers them, and composes whatever matched, instead of hardcoding `vikeExtensionPersonas`. Framework knowledge comes in as a skill, not a gate.

## Scope

- [x] Define the `FrameworkExtension` SPI (`defineFrameworkExtension`)
- [x] Define the skill unit (pointer to an `llms.txt`) and how extensions/presets register skills
- [x] Discovery: find + register installed `framework-*` packages (plus explicit opt-in via config)
- [x] Compose matched extensions in `run.ts`, drop the hardcoded `vikeExtensionPersonas` / Vike gate
[...]

PR #550, opened 2026-07-15
feat(framework): remove the persona / skill / memory framing (#547)

### LINKED ISSUE #547: [The Framework] Remove the persona / skill / memory framing
@brillout asked on #546: "What is this? Can we remove it all?" Answering with the real text rather than a description.

A live build run, prompt = *"a single-file Node script hello.js that prints the 10th Fibonacci number, and a README.md"*. 8,852 chars of system prompt:

| part | chars |
|---|---|
| your #326 block | 3,870 |
| **personas** | **3,764** |
| skills | 609 |
| project memory | 609 |

The personas are as big as your whole prompt.

**Personas (3,764)** — stack opinions appended to every build. That Fibonacci script was told to `build UI on Vike (Vite + SSR)` with `+Page` routing conventions; to `own the data layer. Default to Prisma` including `npx prisma init --datasource-provider postgresql`; and to `express UI as intent, not as concrete markup`. Its own run log reads:

```
Detected Vike (confidence 0); framing with 3 persona(s), 2 skill(s)
```

Confidence **0**: the empty dir matched nothing, the preset is the fallback, so it framed anyway.

**Skills (609)** — doc pointers. `# Skill: Vike` and `# Skill: Engineering Practices`, each a title, a description and an llms.txt URL, plus "consult that document rather than relying on memory".

**Project memory (609, #260)** — tells the agent to read and maintain CODE-OVERVIEW.md / KNOWLEDGE-BASE.md / BRAINSTORMING.md at the repo root, and inlines their current contents. Small here only because the dir was new; it inlines whole files, so in a real repo it is the part that grows without bound.

Correction to what I wrote on #546: I said the memory files were most of the 8.8k. They are 7%. The personas are the bulk. Sorry.

What removal costs:
- **Personas + skills**: pure prompt, nothing else reads them. Clean to cut.
- **Project memory**: partly not a prompt. DECISIONS.md is written from the decisions ledger at run end, so "stop telling the agent about the files" and "stop writing the files" are two separate calls.

Your call on all three. #545 is the same question for meta-select + architect. If both go, a build run's system prompt is your #326 block and nothing else — which also makes it fully previewable before the run, closing the gap I left in #520.

suleimansh:
Closes #547. Follows #548 + #549 (#545), both merged.

A run's system prompt is now the built-in #326 prompt plus your own `SYSTEM.md`. Nothing else. Nothing is read off disk and appended when the run starts.

## The number

A build used to append the personas and skills for its detected stack plus the full contents of the repo's memory files. Measured on a real `--fake` build:

| | before | after |
|---|---|---|
| build run system prompt | 8,852 chars | **3,874 chars** |

3,874 is exactly what a `prompt`-kind run already composed. **The two paths now produce byte-identical system channels**, which is the proof the framing is actually gone rather than merely disconnected.

None of the removed text was designed. It accumulated, and it was bigger than the prompt it was wrapping.

## Two things fall out of this

**1. The prompt preview is now exact. This closes a gap I left in #520.** The "See actual prompt sent" disclosure had to carry a caveat that a build appends more at run time and that a pre-run preview therefore could not be exact. That is no longer true, so the caveat is deleted. What you read before the run is what the agent gets, for every run kind.

**2. A build finally stops being opinionated about the stack.** #549 removed the architect turn, but `personas/library.ts` still hard-instructed "Default to Prisma" and the Vike personas were stack-opinionated. #549 and this PR together are what deliver that. Neither did it alone.

## The call I made that you may want to reverse

**The extension SPI (#190) goes with them.** `FrameworkExtension` is `{name, capability, personas, skills, signals}` - strip personas and skills and it carries literally nothing, so it cannot survive as a half-thing. #190 is closed, so this is not cutting across live work, but it is a deletion beyond "remove the prompt text" and it is yours to veto. If you want the registry kept as a seam for the first-class skills feature later, say so and I will restore it without the prompt text.

Deleted with it: `discoverExtensions`, `--compose-extensions`, and the two example packages that existed only to prove the SPI (`framework-discovery-demo`, `framework-hello`).

## What survives

- **Preset detection.** `readProjectSignals` moved to `project.ts`; a run still narrates `Detected Vike (confidence 2)`. Detection now only narrates - none of it reaches the prompt, and a test pins that.
- **Domain presets.** They lose `skills` and keep their loops, prompts, modes, and review policy, which is the substance.
- **`@gemstack/ai-skills` is untouched.** Different thing (an on-disk `SKILL.md` loader for ai-sdk agents), and load-bearing: its `parseSkillManifest` parses every prompt body and preset markdown in ai-autopilot. Confirmed zero diff.

## Locked down

Two tests pin the property, because a description is not a guarantee:
- **unit:** `composeRunSystem()` exactly equals `[#326 block, AWAIT_PROTOCOL, SIGNAL_PROTOCOL].join('\n\n')`, plus one proving no supported option can append a trailing section.
- **integration:** captures what the driver is really started with in a full `runFramework` run and asserts exact equality. That is the one that would catch `run.ts` reintroducing framing; the unit test alone cannot.

## Verify

- framework 524 tests (523 pass / 0 fail / 1 skipped, a pre-existing Docker-unreachable case); ai-autopilot 280 pass; bootstrap-quickstart 2 pass; autopilot-quickstart 2 pass
- workspace typecheck 21/21 and build clean
- `framework --fake --no-dashboard`: `system prompt sent (3874 chars)`, `Detected Vike (confidence 2)`, runs through to production-grade, exit 0


suleimansh:
Correcting an overstatement in the description now that this is merged, so the record is right.

I framed the result as "a build prompt is the #326 block". Measured, it is 47% of it:

| block | chars | who wrote it |
|---|---|---|
| #326 | 1,832 | you |
| `AWAIT_PROTOCOL` | 1,415 | me |
| `SIGNAL_PROTOCOL` | 623 | me |
| **total** | **3,874** | **53% mine** |

Everything this PR claims is still true: 8,852 -> 3,874, personas / skills / memory gone, build and prompt runs byte-identical, the preview exact. What is not true is that what remains is all yours. The emit protocols are the majority of it, and they are mine.

Picked up on #326 with the full text and the ask: https://github.com/gemstack-land/gemstack/issues/326#issuecomment-4983310471

Issue #556, opened 2026-07-15
Post-merge prompt: add TODO entries instead of running the presets inline


suleimansh:
The #326 OP second ```md block (`## Post-merge prompt`) adds TODO entries for the `maintainability` / `readability` / `security_audit` presets. What ships (`POST_MERGE_PASSES` + `runPostMergeSuite`, cli.ts) instead runs the three presets inline, as child `framework prompt` runs, right after `setReadyForMerge()`. The OP version routes through the backlog queue, so it composes with the TODO loop (#538). Decided on #326: TODO entries.

Follow-up to #555, which syncs the first block only.

Three blockers, all verified against the real renderer:

1. `tf.session_name` is not on `TfContext`. It is set by setSessionName() before the agent makes changes, so it has to be carried on run state and read at post-merge time. Not chicken-and-egg: it never appears in the system prompt block, only here.
2. `tf.settings.technical_control` is not on `TfContext`. The setting itself already exists (`Preferences.technical` in registry.ts, the `--technical` flag, and the Global options toggle from #314/#371), so this is a wire, not new infrastructure.
3. **The block does not render even with 1 and 2 present.** It nests `${{ tf.session_name }}` inside an outer `${{ ... }}`, and wraps `readability` in backticks inside a backtick template literal. `FRAGMENT` is non-greedy, so the outer fragment terminates at the inner `}}` and the leftover is not valid JS:

```
[framework] template fragment ${{ !tf.settings.technical_control ? '' : (`
- "Apply preset `readability` on the changes introduced by ${{ tf.session_name }} failed to evaluate: Unexpected identifier 'readability'
```

So this block cannot be synced verbatim: nested fragments and backticks inside backticks cannot work against this renderer. It needs a flattened equivalent, which is a change to prompt text and so needs a review round before it lands.

## Proposed flattened block (renders, verified)

Same logic, one fragment, no nesting:

```md
TODO_FILE: `TODO_<SESSION_NAME>.agent.md`

## Maintenance

If the changes introduced by ${{ tf.session_name }} aren't trivial and have refactor potential, add the following to <TODO_FILE>
- "Apply preset `maintainability` on the changes introduced by ${{ tf.session_name }}"
${{ tf.settings.technical_control ? '- "Apply preset `readability` on the changes introduced by ' + tf.session_name + '"\n' : '' }}
If the changes introduced by ${{ tf.session_name }} can potentially lead to security issues, add the following to <TODO_FILE>
[...]

Issue #439, opened 2026-07-12
[The Framework] dashboard: context selector + trust-on-add


suleimansh:
Part of #314. Two linked pieces, neither started: (1) a multi-select dropdown of directories that adds a `Context: [dirs]` line to the system prompt (AI can access all repos but shouldn't read all of them); (2) on adding a repo, a "do you trust this repository?" confirmation that warns about prompt-injection risk.

PR #559, opened 2026-07-15
feat(framework): collect business knowledge in the repo (#537)

### LINKED ISSUE #537: Product management: collect business knowledge
Maintain documents like: 
- .the-framework/README.md (whole repo overview)
- .the-framework/DECISIONS.md
- .the-framework/KNOWLEDGE-BASE.md
- ~~.the-framework/BRAINSTORMING.md~~ not sure I see the added value here, skip for now

Not sure yet whether MVP or post-MVP. If it's a quick-win, then why not MVP.

## Implementation

- Add these `.md` to `## Context` (same context as the directories the user can select)
- In post-merge system prompt, tell the agent to consider updating these.

suleimansh:
Closes #537

Both bullets from the issue's Implementation section.

**1. The docs go on the `Context:` line.** `.the-framework/README.md`, `DECISIONS.md` and `KNOWLEDGE-BASE.md` join the `#439` context line on every run, alongside the directories the user selects. One `KNOWLEDGE_DOCS` const; a test pins the post-merge prompt against it so the two halves cannot name different files.

**2. The post-merge prompt asks the agent to update them.** New `## Business knowledge` section:

```md
## Business knowledge

Consider whether the changes introduced by <SESSION_NAME> taught you something that belongs in these documents, and update them if so (create one if it doesn't exist yet):
- `.the-framework/README.md` (whole repo overview)
- `.the-framework/DECISIONS.md` (decisions taken, and why)
- `.the-framework/KNOWLEDGE-BASE.md` (business knowledge about the project)

Only write what a future agent would need and cannot get from the code itself.
```

`BRAINSTORMING.md` is left out, per the issue.

### Two calls made here

- **The docs ride with the built-in prompt, not with the user's dirs.** Under `--vanilla` ("Disable system prompt") only the user's own `--context` dirs survive. The docs are ours, and #547 rule 3 is no framework-authored prompt when it's off.
- **`--eco-auto-maintenance` no longer skips the whole post-merge run**, it drops just `## Maintenance`. Skipping wholesale was right while the post-merge prompt *was* exactly the maintenance section (#556). It isn't any more, and the flag does not name business knowledge, so it must not take it along.

### Ran it for real

Real binary, real git repo, ~$1. A repo whose `DECISIONS.md` said *"flat pricing, no surge"* while the session's commit shipped surge pricing:

- The agent read `DECISIONS.md` off the context line unprompted, and caught the contradiction.
- It wrote a `KNOWLEDGE-BASE.md` that records what the code cannot: that the `1.4` multiplier has no recorded derivation, that `bookedHour` has no timezone attached, and that `sameDay` is caller-supplied so a client could opt out of surge. It explicitly left the surge rationale out, because that one *is* in a code comment.
- Both maintenance gates came back negative on a 7-line change, so no TODO file. The conditions work.

`composeRunSystem` 4475 -> 4573 chars.

Green: framework tests pass except `runCli --daemon backgrounds the dashboard`, which fails identically on clean `main` here (a daemon is already live on :4200) — pre-existing and environmental, not from this change.

### DIFF
--- a/.changeset/business-knowledge-537.md
+++ b/.changeset/business-knowledge-537.md
@@ -1,0 +1,5 @@
+---
+"@gemstack/framework": minor
+---
+
+Collect business knowledge in the repo (#537). Every run now puts `.the-framework/README.md`, `.the-framework/DECISIONS.md` and `.the-framework/KNOWLEDGE-BASE.md` on the `Context:` line, so the agent reads whatever the project already knows about itself, and the post-merge prompt gained a `## Business knowledge` section asking it to fold back what the session taught that the code cannot show. The docs go with the built-in system prompt: `--vanilla` still injects nothing but the user's own dirs. `--eco-auto-maintenance` now drops the post-merge prompt's `## Maintenance` section instead of skipping the whole run, which would have taken business knowledge with it.
--- a/packages/framework/prompts/post_merge_prompt.md
+++ b/packages/framework/prompts/post_merge_prompt.md
@@ -7,3 +7,13 @@ If the changes introduced by ${{ tf.session_name }} aren't trivial and have refa
 ${{ tf.settings.technical_control ? '- "Apply preset `readability` on the changes introduced by ' + tf.session_name + '"\n' : '' }}
 If the changes introduced by ${{ tf.session_name }} can potentially lead to security issues, add the following to <TODO_FILE>
 - "Apply preset `security_audit` on the changes introduced by ${{ tf.session_name }}"
+
+
+## Business knowledge
+
+Consider whether the changes introduced by ${{ tf.session_name }} taught you something that belongs in these documents, and update them if so (create one if it doesn't exist yet):
+- `.the-framework/README.md` (whole repo overview)
+- `.the-framework/DECISIONS.md` (decisions taken, and why)
+- `.the-framework/KNOWLEDGE-BASE.md` (business knowledge about the project)
+
+Only write what a future agent would need and cannot get from the code itself.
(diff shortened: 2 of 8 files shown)


suleimansh:
Known limitation, tracked separately in #560: the post-merge run branches before writing, so the knowledge docs land on a branch nobody merges. Older than this PR (the #556 TODO entries do the same) and not widened into it.

brillout:
- Let's remove `README.md` for now (very likely to be redundant with already existing `README.md`)
- Let's move the files to the repo root
- Add the comments also in the context
  ```md
  - `DECISIONS.md` (decisions taken, and why)
  - `KNOWLEDGE-BASE.md` (knowledge and insights related to the project)
  ```
- post-merge is a misnommer, it should be called to something else e.g. on-before-mergeable
- I don't like the prompt, change it to:
  ```md
  ## Business knowledge

  If you didn't already, consider updating the following based on the changes and discussions of ${{ tf.session_name }} (you can create the files if they're missing):
  - `DECISIONS.md` (decisions taken, and why)
  - `KNOWLEDGE-BASE.md` (knowledge and insights related to the project)

  Only write what a future agent would need and cannot get from the code itself.
  ```

@suleimansh I don't mind AI slope e.g. (DRY issues) for the code, but for all the prompting let's always take an extra look and give it extra thoughts. Thoughful prompts, always. Zero vibe-coded half-hassed prompts, ever. Feel free to ping me for everything related to prompting. When AI will apply the changes above, let's review carefuly whether its changes related to prompting matches what I said above (zero tolerence for AI slope in prompting).

Issue #563, opened 2026-07-16
Backlog turns drop the agent's signals (views, session name, ready-for-merge)


suleimansh:
The backlog loop (#323) prompts the agent through the run's own `DriverSession`, so a backlog turn gets the same system channel as any other turn, `SIGNAL_PROTOCOL` included. `composeRunSystem` appends the protocols unconditionally and its own doc calls them "the emit contract ... the setSessionName()/setReadyForMerge() lifecycle", so the agent is told it can signal on any turn.

`promptItem` only calls `parseAwaitGate`. Everything else the turn carries is dropped:

- `showMarkdown()` views never reach the rail, so backlog work is invisible in the dashboard.
- `setSessionName()` is ignored, so a rename during backlog work is lost.
- `setReadyForMerge()` never emits `ready-for-merge`, and `maybeFirePostMerge` gates on exactly that, so **`--post-merge` cannot fire from backlog work**.

That last one matters now that post-merge writes the follow-up entries into the queue (#558) and the business-knowledge docs (#537): the queue is the backbone, and a turn that finishes there cannot signal it is done.

Proven with the fake driver against the real loop. The parsers agree the turn text carries all three signals; `runTodoLoop` emits `["log","log","log"]` and nothing else.

Cause is duplication: `emitTurnSignals` is byte-identical in `run.ts:577` and `prompt-run.ts:171`, and the third turn site never got a copy. Fix is to extract it once and call it in all three.

PR #602, opened 2026-07-16
refactor(framework): one shared continuation prompt for all gate paths

### LINKED ISSUE #570: The continuation prompt is written three times, differing by one clause
## What we found

When the agent stops to ask and the user answers, we re-prompt it to continue. That continuation prompt exists in three places, identical except for one clause:

- `prompt-run.ts:150` "You paused to ask: \"X\". The user chose: Y. **Continue** with that decision, and do not ask again unless a genuinely new choice comes up."
- `todo-loop.ts:315` "... **Continue the backlog entry** with that decision, ..."
- `steps.ts:166` "... **Continue building \"<intent>\"** with that decision, ..."

Three copies of one sentence means a reword lands in one path and not the others. That is the #563 failure mode applied to wording instead of code.

## What we can do better

Collapse the three into one shared constant. The varying clause does not need to vary: the agent already knows what it is working on from the session, so "Continue" / "Continue the backlog entry" / "Continue building X" all carry the same meaning for it.

If the clause ever does need to vary, the same move works as one template with a single slot instead of a constant.

#569 shares the loop mechanics around this and deliberately leaves all three literals untouched at their call sites, so this can be decided on its own.

suleimansh:
Draft: previews the #570 fix.

## What

The prompt that resumes the agent after it stops to ask lived in three places, identical but for one clause:

- `prompt-run.ts` "**Continue** with that decision, ..."
- `todo-loop.ts` "**Continue the backlog entry** with that decision, ..."
- `steps.ts` "**Continue building \"X\"** with that decision, ..."

Three copies means a reword lands in one path and skips the other two, the #563 drift risk applied to wording.

## Fix

One `continuationPrompt(question, answer)` in `turn-gate.ts` (the leaf module both `run.ts` and `steps.ts` already import, so no cycle). The clause did not need to vary: the agent already knows what it is working on from the session, so `Continue` / `Continue the backlog entry` / `Continue building X` all carry the same meaning to it. `runAwaitRounds` no longer takes a `continuation` callback and builds the prompt itself.

Also trimmed the trailing "do not ask again unless a genuinely new choice comes up" clause per review: a capable agent does not re-ask a settled question on its own, so that instruction was babysitting. Left off until a run shows it is needed; its absence is pinned in the unit test. The shared prompt is now just the decision hand-off.

## Scope / notes

- Five source files. `runAwaitRounds` / `AwaitRoundsOptions` are package-internal (not in `index.ts`), so no changeset.
- Loop mechanics untouched, so this decides on its own.
- Tests: updated the three `runAwaitRounds` call sites, added a `continuationPrompt` unit test asserting one wording, no caller-specific clause, and no "do not ask again" tail. Typecheck clean; turn-gate + run suites green.

Closes #570

### DIFF
--- a/packages/framework/src/prompt-run.ts
+++ b/packages/framework/src/prompt-run.ts
@@ -135,8 +135,6 @@ export async function runPrompt(opts: RunPromptOptions): Promise<RunPromptResult
     const rounds = await runAwaitRounds({
       session,
       prompt: firstPrompt,
-      continuation: (gate, answer) =>
-        `You paused to ask: "${gate.title}". The user chose: ${answer}. Continue with that decision, and do not ask again unless a genuinely new choice comes up.`,
       emitTurnSignals,
       requestChoice: opts.requestChoice,
       emit,
--- a/packages/framework/src/run.test.ts
+++ b/packages/framework/src/run.test.ts
@@ -10,7 +10,7 @@ import { FAKE_DEPLOY, FAKE_INTENT, FAKE_SIGNALS, fakeDriver } from './fake-scrip
 import { FakeDriver, type Driver, type DriverSession } from './driver/index.js'
 import { composeRunSystem } from './system-prompt.js'
 import type { ChoiceRequest, FrameworkEvent } from './events.js'
-import { MAX_AWAIT_ROUNDS, PLAN_DECLINED_MESSAGE } from './turn-gate.js'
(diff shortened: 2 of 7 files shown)


suleimansh:
Leaving this for @brillout to decide

brillout:
> Continue with that decision, and do not ask again unless a genuinely new choice comes up.

I would avoid this kind of babysitting, unless it's clearly needed . (I think agents are smart enough to not ask again.)

brillout:
Let me know if you want me to also look at something else.

Issue #610, opened 2026-07-16
Drive Claude Code on the web instead of the CLI (MVP shortcut: free worktrees + PR, 0% local CPU)


suleimansh:
> Captured from team discussion. Flagged as a potentially large MVP shortcut. The mechanism (how we drive it) and session/login are open; needs a spike.

## Idea

Add a driver that uses Claude Code on the web instead of the Claude Code CLI. We drive the web UI programmatically and let it run the task, rather than running the CLI locally.

## Why it's a big shortcut

- Claude Code on the web already does the whole Git worktree dance and opens the GitHub PR. We reuse that flow instead of building it (relates to #453).
- It runs off the user's machine: roughly 0% local CPU, so AI tasks parallelize essentially without limit.
- It is where at least some of us already work day to day.

## How to drive it (open)

- **Chrome extension**: the first idea. Simple-ish, but an install plus configuration.
- **Headless browser automation**: drive the web UI in a server-side headless browser (no install, seamless). Could reuse the headless-streaming primitive from the preview capability (#609), though the purpose is different (driving the agent vs previewing the agent's browser).
- Decision pending; a spike should compare the two on the hard part below.

## The one hard part: session / login

- [ ] Confirm we can drive an authenticated Claude Code web session programmatically (headless or via extension). This is the main unknown and likely decides extension vs headless.

## Related

- #495 (BYOS): the CLI path for "use the user's own agent". This is the web analog.
- #453 (git worktrees): the web flow does this for us, reducing what we build.
- #609: a separate idea, but shares the possible headless-browser mechanism.
- #605 (daemon): where a server-side headless driver would run.

suleimansh:
## Spike: drive Claude Code on the web

**Recommendation: neither extension nor headless. Use the routines fire API.** Both browser options script an authed claude.ai page to reinvent an API Anthropic already ships, and both are barred by the Usage Policy. The API needs no browser and no new auth beyond a per-routine token.

### The hard part (session / login): answered

Yes, an authenticated web session is drivable programmatically, without cookies or anti-bot. Two official paths, both on the same claude.ai subscription the CLI already holds:

- **Routines fire API** `POST /v1/claude_code/routines/{trig_id}/fire`, bearer token `sk-ant-oat01-...` (per-routine, made once in the web UI), beta header `experimental-cc-routine-2026-04-01`. Returns `{claude_code_session_id, claude_code_session_url}`. Plain HTTP, no TTY. This is the programmatic path.
[...]

Issue #605, opened 2026-07-16
Epic: Daemon + gateway split — headless 24/7 runtime, composable frontend


suleimansh:
> Tracked-but-later (**Phase 1**). Do not build before the MVP ships. Direction captured from team discussion so it is not lost; not scheduled work yet.

## Vision

Split the runtime into composable pieces so The Framework can run anywhere from a single local machine to a hosted service:

- **Daemon**: a separately deployable service that runs 24/7, hosts the workspace, and gives agents shell + filesystem access and the local coding-agent CLIs (Claude Code, Codex) so it uses existing subscriptions rather than API-key pricing.
- **Gateway**: a public forwarding layer (Telefunc server) that joins agents and users into rooms.
- **Frontend**: the dashboard is just "a" frontend that connects to the daemon. Slack/Discord are other frontends.

## Decided direction

- Composable from the start: daemon + gateway + frontend as separate seams.
- Git-as-data: persist everything through Git repo(s), so no database and no real server are needed for local use. It can run 100% local as a desktop app, which underpins the "100% open source / free / local" positioning.
- Current state, for reference: all background jobs run in the same process as the frontend server today (so `Ctrl + C` after `framework` terminates everything). The headless split is the future direction, not a regression to fix now.

## Open questions

- The 24/7 responsibility: exactly one instance must manage the AI chat bot. Options discussed: a paid hosted daemon, a user-run VPS, or a simple lead-designation where one dev-machine instance auto-designates itself as the manager.
- Where the headless-browser preview runs (most likely on the daemon).

## Related

- #298 (background jobs) and its usage-limit management: the daemon is where these run continuously.
- #297 (bootstrap mode), #453 (git worktrees), #454 (syncing UI <-> data): all assume a runtime that can host long-lived, multi-repo work.

brillout:
@suleimansh I think it's mostly done?

suleimansh:
Mostly — daemon, relay, remote devices exist. Hosted 24/7 leg is #1690 now.

🤖 *curated*

PR #623, opened 2026-07-16
spike(#610): WebDriver PoC on the routines fire API

suleimansh:
Spike for #610. Draft until a live-token `/fire` run happens; the mechanism and seam fit are proven.

## What this is

A `WebDriver` that runs a task on Claude Code on the web behind the existing `Driver` seam, via the supported **routines fire API** (`POST /v1/claude_code/routines/{id}/fire`), not the browser UI.

- `prompt(text)` fires the routine and returns the created session handle (id + url) as the turn.
- `collect()` awaits the eventual result (the branch the cloud session pushes / its PR).
- `readCode` is omitted (the workspace is a remote VM). `dispose` is a no-op (cloud session persists server-side).
- Transport is an injectable `fetch`, so it is unit-tested with no network and no real token.

## Why the API, not the browser or the CLI flag

- Driving the claude.ai UI (Chrome extension / headless) is barred by the Usage Policy: no automated access outside an API key, and third-party agents using subscription credentials are banned. The routines API is explicitly permitted.
- Verified `claude --cloud` is TTY-gated (CLI v2.1.211): run non-interactively (piped stdout) it exits 1 and would silently run locally. That is exactly the mode `ClaudeCodeDriver` uses, so `--cloud` cannot slot behind the non-interactive driver. `--cloud` stays the human-at-a-terminal path; the API is the programmatic one.

## Async contract (decided: option A)

The web flow is fire-and-forget, so `prompt` resolves with the session handle, not a final turn. Rather than block the driver, the seam gains an optional `collect()` + `DriverOutcome` (only async/off-machine drivers implement it). Branch discovery is the caller's policy via an injectable `resolveOutcome` (the daemon, #605, polls GitHub), so the driver stays transport-only.

## Tests

10 web-driver unit tests, full driver suite (60) green; `pnpm typecheck` clean. Covers fire + handle, framing fold, events, error-envelope on 429, unrecognized 200, aborted prompt, no `readCode`, and `collect` (no-resolver, injected-resolver, and pre-prompt guard).

Remaining before ready: one live `/fire` against a real routine token, then the changeset. No changeset yet, intentional.


suleimansh:
Closing without merging. The mechanism and seam fit are proven, so the value here is the finding, not the code.

**What the spike settled**
- Claude Code on the web is reachable programmatically only via the routines fire API (`POST /v1/claude_code/routines/{id}/fire`). Driving the claude.ai UI (extension / headless) is barred by the Usage Policy, and `claude --cloud` is TTY-gated (v2.1.211): run non-interactively it exits 1 and silently falls back to a local run, so it cannot slot behind the non-interactive `ClaudeCodeDriver`.
- The web flow is fire-and-forget, which does not fit the synchronous `Driver.prompt -> turn` contract. It needs the optional `collect()` + `DriverOutcome` seam (only async / off-machine drivers implement it), with branch discovery left to the caller (`resolveOutcome`, the daemon polling GitHub per #605), so the driver stays transport-only.

**Why not merge**
It never got a live `/fire` run against a real routine token, so shipping the async seam here would land it unproven. The code (10 web-driver unit tests, full driver suite green, typecheck clean) stays on `spike/610-web-driver` if #610 is picked up later; re-derive the `collect()` / `DriverOutcome` seam from that branch rather than assuming it stays around.

Parent #610 stays open.

Issue #799, opened 2026-07-19
End of session: surface the branch and diff, offer push and PR


suleimansh:
Part of #453.

A session that finishes cleanly leaves its work on a local branch and removes its worktree. Nothing pushes, nothing opens a PR, and the dashboard shows no diff, no branch and no next step. To find out what a session actually did you leave the dashboard and use git by hand.

That is the weakest link in the loop. Everything up to "the agent is done" is well covered; the handoff back to the human is missing.

## Shape

On a finished session:

- show what it produced: the branch, the commits, the changed files, ideally a diff view (`onGitStatus` and `onProjectFileStatus` already exist per run)
- offer the next step rather than describing it: push the branch, open a PR. `gh` is already a dependency of the interventions queue (#632), which then picks the PR up as a "needs you" item, closing the loop back into the dashboard
- say when there is nothing to hand off (a session that changed nothing), instead of showing an empty branch

## Related, not in scope here

The prompt side of the same gap is on #326: the contract tells the agent to commit pre-existing changes at the start but never its own work at the end, which is why teardown has to commit for it (#786).

Issue #737, opened 2026-07-18
Archive run state on worktree teardown + reconcile orphans


suleimansh:
Part of #453. Pairs with the concurrency sub-issue.

Each run's `.the-framework/` (event log, run.json, archived history) lives inside its cwd. With worktrees, that cwd is a throwaway worktree dir, so a run's history would be destroyed when the worktree is removed. This sub-issue makes run state survive teardown.

## Decision needed (Rom)

**Retention on failure.** On a clean finish, remove the worktree. On a failed/stopped run, remove it too, or retain it for inspection (with a prune policy so they do not accumulate)? Proposal: always archive history back to the main repo, remove the worktree on success, retain on failure behind a cap (e.g. keep the last N failed worktrees, prune older).

## Scope

- On child exit (`daemon.ts:505-506`, currently just `activeRuns.delete(key)`): archive the run's `runs/<id>.jsonl` + `runs/<id>.json` from the worktree's `.the-framework/` back into the **main repo's** `.the-framework/runs/`, then `removeWorktree`. Idempotent (the child is detached; the daemon only holds its pid).
- Extend boot-time `reconcileOrphanedRuns` (`run-store.ts:416`, called per project in `daemon.ts:330`) to scan `.the-framework/worktrees/*` for orphaned runs left by a daemon crash: archive their history, then `git worktree prune` + remove.
- Live run state (`run.json`, `events.jsonl`) stays inside each worktree while the run is live; only the archived, immutable history moves to the main repo. This keeps the worktree self-contained during the run and gives the dashboard-aggregation sub-issue a clear place to read live meta from.

## Out of scope

Dashboard reads (aggregating live meta across worktrees) are a sibling sub-issue; this one just guarantees history is not lost and worktrees do not leak.

suleimansh:
Two refinements for the teardown/reconcile path, borrowed from garrytan/gstack `lib/worktree.ts`:

1. **Age-guarded stale sweep.** The boot-time orphan reconcile should skip worktrees younger than a threshold (gstack uses 1h off the dir's `mtimeMs`) so a concurrent daemon, or a run still mid-startup, isn't torn down under it. Only worktrees older than the guard AND with no live owner get archived + pruned.
[...]

Issue #165, opened 2026-07-03
AI framework: presentation layer — headless server + CLI/web chat UIs


suleimansh:
Child of #110. The orchestration backend (`ai-autopilot`: Supervisor + loop + personas + runner + decisions + overview + bootstrap) is built. What's missing is the **presentation layer**: expose that backend over an API and drive it from two thin UIs.

## Scope

- **Headless server** — wrap the Supervisor in an HTTP server (a `serve` command). It spawns/streams runs and forwards user prompts to `ai-sdk` + the orchestration. Shaped after `opencode serve` (headless OpenAPI server that many clients drive).
- **CLI client** — the terminal UI, already the reference surface (`terminalSink` over the Supervisor `onEvent` stream).
- **Web client** — a localhost chat with a stop button (Claude-Code-for-web style). Renders the **same `EventStream`** the terminal consumes; the input calls the scope/bootstrap entry.

The event transport already exists (`EventStream`, replayable multi-consumer), so the web UI is mostly a renderer, not new orchestration.

## Open decision (with Rom, pending)

Build our own thin shell, or sit the orchestration on top of an **open, provider-agnostic** agent shell like [opencode](https://github.com/anomalyco/opencode) and stay the "brain"? Sitting on it gets CLI/web/IDE clients for free. Different risk profile than wrapping Claude Code (opencode is OSS + provider-agnostic). Resolve before building.

Reference: `opencode serve` = one headless backend, TUI/web/desktop/IDE all clients.

---

## Direction (design discussion, 2026-07-03)

Converged with Rom on the shape:

- **Web-first.** A localhost website is the primary UI; a CLI can come later as a subset. The web unlocks things the CLI can't (e.g. bootstrap mode showing the chosen stack + PROS/CONS in a sidebar).
- **Orchestration is an *outer* loop.** We treat the AI CLI (Claude Code, Codex, ...) as a **black box** and leverage its built-in loop. Our layer drives it via prompts, reads the code it produced, and decides the next prompt. We do **not** reach inside its loop or micromanage its tools. This keeps the "use a Claude subscription" path intact (we don't touch their internals) and is more future-proof.
- **The seam is the code**, not tool-level control. Verify by **outcome** (does it build, does it serve, did review pass, does the PR exist), and re-prompt on failure. Outcome-checked, not tool-checked. This is exactly what `ai-autopilot`'s loop already does (gates on verdicts/serve-check, not on tool calls).
[...]

PR #823, opened 2026-07-19
feat(framework-dashboard): show the session's file changes in the run output (#817)

### LINKED ISSUE #817: Show the session's file changes in the run output
Part of #453. Needs #816. Overlaps the "show the diff" half of #799.

Watching a run, you see `· Edit` and `· Write` go by and never learn which files moved. To find out you leave the dashboard and run `git diff` in the worktree by hand. Claude Code's own output shows the change inline; ours shows the tool name.

## Not from the tool arguments

The obvious route is to widen the driver event and carry `file_path` / `old_string` / `new_string`. That fights an explicit guardrail: `driver/types.ts` says "we surface the name only, not the arguments" (#165), because we drive by prompting and verify by outcome, not by watching which tool the agent reached for. It also only works for Claude Code; the codex adapter emits an item type and nothing else.

## From git instead

Derive it from the worktree, which is the outcome and is agent agnostic:

- a Changes section in the run view, from `onProjectFileStatus(projectId, runId)`, the same read the tree uses
- per file, added and removed line counts, expandable to the diff via `onFileDiff` (#816)
- it keeps working after the run ends, as long as the worktree is retained, which is the diff half of #799

One read, no driver change, and Codex sessions get it for free.

suleimansh:
Closes #817. Stacked on #822, review that first.

Watching a run you saw `· Edit` go by and never learned which file moved. Finding out meant leaving the dashboard for `git diff`.

A Changes section above the event log lists what the session touched, each row expanding to the diff.

## Not from the tool arguments

The obvious route is to widen the driver event and carry `file_path` / `old_string` / `new_string`. That fights an explicit guardrail: `driver/types.ts` says "we surface the name only, not the arguments" (#165), because we drive by prompting and verify by outcome. It would also only work for Claude Code; the codex adapter emits an item type and nothing else.

So this reads the worktree instead. One `git status` plus one `git diff --numstat` per poll, not a diff per file, and a row's diff is fetched only when you expand it. Codex sessions get it for free.

Only rendered for a session whose worktree is still there (live, or finished and retained #737). Once a worktree is removed `resolveRunPath` falls back to the project root, and the panel would otherwise present the user's own uncommitted files as the session's work.

This is the diff half of #799.

## Verified

Against a daemon driving a stub agent through a real dashboard-started session in its own worktree: `CHANGES 3 files +13 −1`, rows `app.ts MODIFIED +10`, `src/greet.ts NEW +3`, `src/keep.ts DELETED −1`, expanding to the inline diff. 5 server tests + 5 dashboard tests.


suleimansh:
Rebased onto the current stack after #820 (#799) merged. That PR's `RunHandoffPanel` wanted the same slot in `RunReplay`, so the two now split by session state rather than sitting on top of each other:

- **live session** -> `RunChanges` (this PR): uncommitted working-tree state, which is the only thing that exists while the run is still going.
- **finished session** -> `RunHandoffPanel` (#799): branch-addressed, so it survives the worktree removal that leaves this panel with nothing to read.

So `RunChanges` dropped out of `RunReplay` entirely and is live-only now. The export-list conflicts across the four shared files were additive on both sides and kept whole.

Rechecked after the rebase: 785 framework tests, 163 dashboard tests, 0 failures, both packages typecheck.

PR #827, opened 2026-07-19
fix(framework): tell the agent it has a browser (#824)

### LINKED ISSUE #824: Browser runs: tell the agent it has a browser
`--browser` wires the chrome-devtools MCP server into the run, and the system prompt never mentions it. So the agent reaches for `WebFetch`, which it knows, and the browser sits on `about:blank` for the whole run.

Seen on a real run: "Open https://news.ycombinator.com, then click into the top story's comments" produced two `WebFetch` calls and no navigation. The bridge came up and published its port correctly; there was simply never anything on the page to show. Re-running with "Use the chrome-devtools browser tools, not WebFetch" in the prompt made it navigate immediately.

That makes the Browser toggle a promise the agent cannot see. A user turns it on, watches the preview stay blank, and concludes the preview is broken, when nothing about the preview is.

**Fix**

Add a browser section to the system prompt, included only when the run has a browser. It should say the browser exists, and that anything the agent needs to *see*, *interact with*, or *log into* goes through the chrome-devtools tools rather than `WebFetch` — `WebFetch` stays right for reading a page's text.

This pairs with the `await-browser` protocol already in the prompt: that section tells the agent how to hand the browser over, while never establishing that it has one.

suleimansh:
Closes #824

`--browser` wires the chrome-devtools MCP server into the run. Nothing in the system channel ever said so, so the agent reached for `WebFetch`, which it knows, and the browser sat on `about:blank` for the whole run. The preview then had nothing to show, which reads as a broken preview rather than an unused browser.

**Before and after, same prompt, no hint in it**

| | tools used |
|---|---|
| before | `WebFetch`, `WebFetch` |
| after | `mcp__chrome-devtools__new_page`, `mcp__chrome-devtools__evaluate_script` |

Before, getting the browser used at all took spelling it out in the prompt: "Use the chrome-devtools browser tools, not WebFetch."

**Where it goes**

A new `prompts/protocols/browser.md`, added by `composeRunSystem` when the run has a browser. It sits with the emit protocols rather than inside the built-in prompt, because like them it describes what this run can do — so `--vanilla` keeps it, and transparent mode (an empty channel) does not.

It is placed ahead of AWAIT/SIGNAL so the signal protocol is still the last thing in the channel, which #547 pins.

**Gated on the tools really being there**

`browserAttached` is narrower than the flag: the tools ride Claude Code's MCP config, so `--browser` on another agent wires nothing (`unguardedNotices` already says so) and the fake driver has none. Claiming a browser the run does not have would be worse than saying nothing.

Suite green: 788 pass.

### DIFF
--- a/packages/framework/prompts/protocols/browser.md
+++ b/packages/framework/prompts/protocols/browser.md
@@ -1,0 +1,7 @@
+## You have a real browser
+
+This run has a real Chrome attached, through the `chrome-devtools` tools (`new_page`, `navigate_page`, `click`, `fill`, `take_snapshot`, `evaluate_script`). It is the same browser a human can watch and take over, so use it for anything you need to *see* or *act on*: pages that render their content with JavaScript, a flow you have to click through, a form to fill, an app you are checking actually works.
+
+`WebFetch` is still the right tool for reading an article or a doc page as text. Reach for the browser when fetching the HTML would not answer the question.
+
+Prefer one page and navigate it, rather than opening a new page per URL: the human watching sees the tab you are on.
--- a/.changeset/tell-agent-about-browser.md
+++ b/.changeset/tell-agent-about-browser.md
@@ -1,0 +1,5 @@
+---
+'@gemstack/framework': minor
+---
+
+Tell the agent it has a browser. `--browser` wired the chrome-devtools tools into the run but the system channel never mentioned them, so the agent reached for `WebFetch` and the browser sat on `about:blank` for the whole run, taking the preview with it. Runs with a browser attached now get a short section saying it exists and that anything it needs to see or act on goes through those tools. Only when the tools are really there: the flag wires nothing on another agent or the fake driver.
--- a/packages/framework/src/cli.ts
+++ b/packages/framework/src/cli.ts
@@ -1233,6 +1233,12 @@ export async function runCli(argv: string[], io: CliIO = defaultIO): Promise<num
     ? fakeDriver()
     : createDriver({ agent: opts.agent, claudeOpts: withBrowser(claudeOpts, opts.browser, sharedBrowser?.browserUrl) })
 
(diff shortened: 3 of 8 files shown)


packages/framework/prompts/protocols/browser.md
```diff
@@ -0,0 +1,7 @@
+## You have a real browser
+
+This run has a real Chrome attached, through the `chrome-devtools` tools (`new_page`, `navigate_page`, `click`, `fill`, `take_snapshot`, `evaluate_script`). It is the same browser a human can watch and take over, so use it for anything you need to *see* or *act on*: pages that render their content with JavaScript, a flow you have to click through, a form to fill, an app you are checking actually works.
+
+`WebFetch` is still the right tool for reading an article or a doc page as text. Reach for the browser when fetching the HTML would not answer the question.
+
+Prefer one page and navigate it, rather than opening a new page per URL: the human watching sees the tab you are on.
```
brillout, line 7: I don't understand this line, but it's fine if it isn't ambiguous to the agent.

brillout: (The other lines are good.)

brillout: Just seeing your PMs now, I understand the line now.

brillout, line 7:
```suggestion
Prefer navigating within a single page rather than opening new pages, so the user can more easily watch you navigate.
  ```

brillout: (Many users will read the system prompt, so good to make it clear for both humans and agents.)

suleimansh: Fixed on #838

suleimansh: Fixed on #838

Issue #801, opened 2026-07-19
Option labels that promise more than the code does


suleimansh:
Three option labels promise behavior the code does not deliver. Each is small on its own; together they make the options menu untrustworthy, which is worse than any one of them.

1. **Autopilot: "also relaxes the maintenance stance"** (tooltip in `Composer.tsx`). The mode only selects `conditions` variants of a domain preset's files. There is no `--preset` surface in the dashboard, and the built-in #326 prompt does not branch on `tf.params.autopilot`. Against the shipped prompt, Autopilot does nothing beyond the choice-gate countdown. Either wire the prompt to the mode, or say plainly that Autopilot auto-answers gates.

2. **Eco > Auto maintenance does nothing on its own.** That section moved out of the main prompt (#556), so the toggle now only bites when Post-merge cleanup fires. Enabled alone, it changes nothing. Either gate it on Post-merge in the UI, or fold it into that option.

3. **Browser is inert under Codex.** The CLI warns via `unguardedNotices()`, the dashboard does not. The checkbox should be disabled with the reason when the selected agent is Codex, the way Eco is disabled under Vanilla.

Also worth deciding: **Post-merge cleanup skips silently** when the agent never called `setSessionName()` or never signalled ready-for-merge. It logs a line the dashboard does not show, so from the UI the option simply did not happen. It should say so.

Issue #773, opened 2026-07-19
PM: auto work on quick-wins


brillout:
The idea is to automatically work on tickets that are quick-wins when TODO_AGENTS.md is empty and when there's plenty of available tokens to spend (to max out the user's AI subscription quota).

Implementation:
- Simple prompt: "look at all tickets/**.plan.md and add all quick-wins to TODO_AGENTS.md"
- Cron job: when enough available tokens => run the prompt above

See also:
- https://github.com/gemstack-land/gemstack/issues/683
- https://github.com/gemstack-land/gemstack/issues/685

Issue #684, opened 2026-07-18
Ticket format


brillout:
Static file `ticket-format.md`:

~~~md
## tickets/<DATE>_<SLUG>.md

DATE: yyyy-mm-dd
SLUG: succinct kebab-case slug of the ticket title
Content:
```md
priority: low/medium/high/urgent [optional]
topics: [list-of-topics] [optional]

# Ticket title

## TLDR

...

## Why it matters

...

[optional: more info (any heading and format you want)]
```


## tickets/<DATE>_<SLUG>.spike.md

For an existing ticket (e.g. `tickets/2042-01-01_some-ticket.md`), a spike can be created (`tickets/2042-01-01_some-ticket.spike.md`).

Content:
```md
# [Spike] Ticket title

## TLDR

...

## Analysis

...

[optional: more info (any heading and format you want)]
```

Typical spike content:
- High-level overview, for example:
  - What it takes to implement it
  - List all the ways to implement it
  - Potential "laziness" shortcuts
  - Full-fledged implementation VS minimal implementation
  - Open questions
  - ...
- Estimated effort (for each ways to implement it):
  - Human intervention effort: trivial/low/medium/high/very-high
  - Token consumption: time estimate (minutes, hours, or days)


## tickets/<DATE>_<SLUG>.plan.md

For an existing ticket (e.g. `tickets/2042-01-01_some-ticket.md`), a detailed plan can be created (`tickets/2042-01-01_some-ticket.plan.md`).

Content:
```md
# [Plan] Ticket title

## TLDR

...

## Plan

...

## Hard problems [optional]

...

## Variability [optional]

...

[optional: more info (any heading and format you want)]
```

Typical plan content:
- Concrete and detailed plan on how to implement the ticket, for example:
  - Exhaustive list of all aspects (including edge cases)
  - Thorough analysis
  - Overview of code changes
  - ...
- Plan broken down in multiple (sub)plans in case of variability
- Hard problems
  - Exhaustive list of all aspects with low confidence regarding how to solve them (the hard problems), with explanation why
- Variability
  - List all aspects that need to be implemented
  - Give a rating to each aspect (from 0 to 10) following this criteria: is there an obviously optimal way to implement it (10), or is it highly unclear whether it can be implemented in a better way (0)?
  - Explore and suggest alternatives for aspects with a low rating
  - For each aspect that has alternatives: list all alternatives sorted in a sensible order, <SHOW_CHOICES>
```
~~~

> [!NOTE]
[...]

Issue #683, opened 2026-07-18
Update Context prompt fragment


brillout:
```md
- TODO_AGENTS.md # AI task queue, format: path/to/todo-format.md
- GOAL.md # The goal of the project (includes long-term direction, scope, non-scope, ...)
- BUSINESS_LOGIC.md # Codebase business logic
- tickets/*.md # List of things to potentially work on, format: path/to/ticket-format.md
- conversations/<DATE>_<TOPICS_SLUG>.md # Conversations between humans, recorded by agents (e.g. via the agent Discord bot)
- knowledge-base/DECISIONS.md # Decisions taken, and why
- knowledge-base/FACTS.md # Non-common or non-obvious facts relevant to the project
- knowledge-base/INSIGHTS.md # Insights relevant to the project
- knowledge-base/MARKET_RESEARCH.md # Ongoing market research
- knowledge-base/*.md # More files holding knowledge related to the projcet
```

brillout:
@suleimansh Updated context

brillout:
@suleimansh OP updated

brillout:
@suleimansh updated

PR #839, opened 2026-07-19
feat(framework): spend leftover quota on spiking & planning tickets (#685)

### LINKED ISSUE #685: PM: auto spike & plan tickets
The idea is to automatically spike & plan tickets when TODO_AGENTS.md is empty and when there's plenty of available tokens to spend (to max out the user's AI subscription quota).

I updated the following to include the format for `tickets/**.spike.md` and `tickets/**.plan.md`:
- https://github.com/gemstack-land/gemstack/issues/684

See also:
- https://github.com/gemstack-land/gemstack/issues/683
- https://github.com/gemstack-land/gemstack/issues/773

suleimansh:
Closes #685

## What

When the agent queue has run dry, nothing is running, and there is plenty of budget left, the daemon starts a PM run by itself: **spike & plan the tickets that have neither yet**. Leftover subscription quota goes on the roadmap instead of expiring.

First of the two MVP v3 autonomy items on #538. #773 (auto quick-wins) builds on the same loop and follows.

## Shape

- `auto-pm.ts` holds the whole policy as a pure function, so the daemon only supplies readings and the decisions are testable off disk.
- The loop is the `startInterventionWatcher` pattern already in the daemon: start it, tick it, stop it on shutdown.
- The prompt is one line, per the #674 call on the ideation preset. The #683 context fragment already points the agent at `tickets/**.md` and the #684 format spec, so re-teaching the file shape here would just go stale.
- Also available as a **Spike & plan** preset button, so the same thing can be run by hand.

## The quota gate fails closed

This is deliberately the opposite of the per-run guard. #519 settled that an unreadable quota must never *stop* the user's own work, so `startConsumptionGuard` fails open. Work nobody asked for is the other way round: if we cannot see the meter we cannot promise we are spending a surplus, and quietly burning someone's subscription is a much worse failure than skipping a tick. So no reading means no run, and with every limit disabled there is no denominator to call anything "spare".

## Defaults

- Off unless `autoPm` is on. It spends the allowance unasked, so it is opt-in.
- Starts only while **every enabled limit still has half its budget free**.
- Ticks every 10 min; a project gets a 30 min cooldown after a start, so a tick landing before the spawn registers cannot double up.
- Dies with the daemon. Per #519 a Ctrl+C that stops everything is the feature, not a gap.

The toggle lives in the Usage panel, beside the limits it spends against, rather than in the run options menu: it is daemon behavior, not a flag on the run you are about to start, which is exactly the mismatch #801 is cleaning up.

## Open question

The threshold (half free) and the 10 min tick are my defaults, not yours. Say the word and I will change them.

## Verified

`pnpm build` green (10/10). Framework: 811 pass / 0 fail, including 14 new. Dashboard: 173 pass, typecheck clean.


suleimansh:
@brillout, we have many 'Defaults', maybe they will grow more..
Its ok, but UI/UX - maybe its better to have a nice simple 'User preferences' page ?

brillout:
How about we simplify and re-use the same global usage quota limits instead of introducing new ones?

Is there a use case for having two different limits (one for auto-PM and one for other autopilot features)? I don't think so? Maybe it's overkill?

So I'd suggest we replace the "when there's plenty of token left to spent" part with "when the usage quota limits aren't met".

In the future, we can maybe have two limits:
- Soft => stop autopilot (the current limit settings)
- Hard => stop all work (but, tbh, I ain't sure what the use case for this would be)

Issue #852, opened 2026-07-20
Auto PM re-does the same work forever: its output lands on a branch it never reads


suleimansh:
Proven end to end against a live daemon. Auto PM works, and then repeats itself indefinitely.

## What happened

A scratch repo with two `.plan.md` quick wins and an empty `TODO_AGENTS.md`. Daemon started, `autoPm` on, isolated config.

At the 10 minute tick it did exactly the right thing:

```
[framework] auto PM: harvesting quick-wins from the plans in .../tf-scratch
```

Worktree allocated, run spawned, no gate parked, `status: done`, `readyForMerge: true`. The agent read both plans, judged them quick wins, and queued them well — it even caught that one plan misquoted the source (`"Helo, world"` vs the actual `` `Helo, ${name}` ``).

The queue it wrote, on branch `the-framework/queue-quick-wins`:

```md
## Quick wins
- [ ] **Fix the greeting copy** - change `Helo` to `Hello` in `greet.js`. ...
- [ ] **Add a farewell helper** - export `farewell(name)` from `greet.js` ...
```

## The problem

That is on a **branch**. The project checkout still reads:

```
# Backlog

Nothing queued right now.
```

Auto PM decides idleness with `findTodoBacklog(project.path)` against the project checkout, which never sees the branch. So:

- the queue is still "empty", so after the 30 minute cooldown it starts **another** run
- the rotation moves on, and in two ticks it is back to quick-wins, re-deriving the same entries onto a new branch
- the backlog loop never drains them either, for the same reason

Left alone it produces an unbounded pile of near-identical branches and spends real quota on each. Runs in worktrees is correct (#736); the mistake is auto PM treating "the checkout's queue is empty" as "there is nothing outstanding".

This also falsifies a claim I made in #773: that a quick-wins run leaves the queue non-empty so auto PM stands down on its own. It does not, because the run's output never reaches the checkout it reads.

## Options

1. **Stand down while an auto-PM result is unmerged.** Treat a finished, unmerged auto-PM branch as outstanding work. Minimal, keeps worktrees, matches the settled "the proposal IS the PR" model from #624: it proposed, a human accepts or rejects, and until then there is nothing to add.
2. Have auto-PM runs commit the queue straight to the checkout's default branch. Closes the loop with no human, but writes to main unattended.
3. Make `backlogEmpty` consider the framework's own branches, not just the checkout.

[...]

PR #856, opened 2026-07-20
fix(framework): auto PM drains the agent queue, not just fills it (#855)

### LINKED ISSUE #855: Auto PM fills the queue but nothing drains it: autonomy stops after one turn
Auto PM (#685/#773) fills the agent queue but nothing unattended ever drains it, so the autonomy loop stops after exactly one turn.

## The deadlock

1. Queue is empty, quota is spare, so auto PM starts a quick-wins run.
2. #854 promotes that run's `TODO_AGENTS.md` from its worktree branch into the checkout.
3. Now `backlogEmpty` is false, and `autoPmDecision` refuses: `the agent queue still has open entries` (`auto-pm.ts:132`).
4. Nothing drains it. So step 3 is the permanent state.

Two independent facts make it permanent:

**a. There is no daemon-level backlog drainer.** `runTodoLoop` has exactly one non-test caller, `run.ts:410`, i.e. a phase at the end of an already-running build. In `daemon.ts` the only non-test caller of `runtime.onStart` is auto PM itself (`daemon.ts:465`). So the only thing that starts a run unattended is the one loop that refuses whenever there is work to do.

**b. Auto PM's own runs cannot drain, by construction.** It starts runs with kind `'prompt'` (`daemon.ts:465` → `['prompt', prompt]` at `daemon.ts:690`). The prompt path returns early through `runPrompt` (`cli.ts:1327-1350`) and never reaches `runFramework`, so `runTodoLoop` never runs. Auto PM can only ever fill.

The comment at `auto-pm.ts:130` says "the backlog loop has work to drain" - true inside a session, not true of a daemon with nobody at the keyboard.

## Verified

Against the real checkout, real functions, best-case inputs (1%-used week, zero active runs):

```
backlog file : TODO_AGENTS.md
open entries : 2
decision     : {"start":false,"reason":"the agent queue still has open entries"}
```

Live daemon on an isolated registry confirms the silence. Which is a second, smaller problem: refusals are not logged (`if (!decision.start) continue`, `auto-pm.ts:281`), so a wedged auto PM is indistinguishable from a healthy idle one.

## Fix options

1. **Auto PM drains before it fills.** When the queue is non-empty, start a build-kind run instead of standing down. Smallest change, keeps one loop, and the rotation already exists. Needs the run to be build-kind so `runTodoLoop` fires, and `--unattended` so the per-item gate takes the recommended option rather than parking (#846).
2. **A separate daemon drain loop**, symmetric to auto PM. More moving parts, two loops that must not start runs at the same time.
3. **Leave draining to a human.** Then #685/#773 are a queue-filling convenience, not autonomy, and #538's MVP v3 goal is not met.

Recommend 1.

Also worth logging the refusal reason at each tick either way.

suleimansh:
Closes #855

Auto PM filled the agent queue and nothing unattended ever emptied it, so the autonomy loop ran exactly one turn and then went quiet for good.

## Why it deadlocked

`autoPmDecision` refused whenever the queue had open entries, on the reasoning in the comment: "the backlog loop has work to drain". But that loop (`runTodoLoop`) has one non-test caller, `run.ts:410` - a phase at the end of an already-running build. In the daemon, the only non-test caller of `runtime.onStart` is auto PM itself. So the one thing that starts runs unattended is the one that refuses whenever there is work to do.

And auto PM's runs could not have drained it anyway: they start with kind `'prompt'`, which returns early through `runPrompt` (`cli.ts:1327`) and never reaches `runFramework`. It can only fill.

## What changed

A tick that finds open entries starts a run for the first one ([Drain queue], one entry per run) instead of standing down; once the queue is dry the [Quick wins] / [Spike & plan] rotation resumes where it left off. The existing #854 promotion already carries the check-off back from the run's branch into the checkout, which is what lets the next tick see progress.

Draining deliberately does not advance the PM rotation: that cycle is about what to *make* when there is nothing to do, and a queue worked off over several ticks must not skip a job once per entry.

One entry per run, not the whole backlog, so the cooldown and the quota gate apply per entry rather than letting a long queue go off in one unattended burst.

## Two things worth a second look

**An unreadable queue is now its own refusal.** It used to fall through as "non-empty", which was safe when non-empty meant stop. Now both readable answers start something, so "could not tell" needs its own branch or an unreadable queue would start a drain run. An existing #685 test caught this, which is the only reason it is in the diff.

**Refusals are now logged each tick.** They were silent, so a wedged sweep looked exactly like a healthy idle one - that is why this went unnoticed.

## Verified

Live, isolated registry + scratch repo, daemon on 4399. Before the fix: the tick fired at the 10-minute mark and nothing happened - no run, no branch, both queue entries still open, no log line. That is the bug, on merged main.

Build 11/11, typecheck 21/21, framework tests 838 pass / 0 fail (5 new, including one that runs three ticks and asserts drain -> fill -> drain).

## Note

The drain run's code changes stay on its worktree branch; only `TODO_AGENTS.md` is promoted. That is intended - a human still reviews the work - but it is the same "unattended write to the checkout" question raised on #852 and worth ratifying together.

### DIFF
--- a/.changeset/auto-pm-drains-queue.md
+++ b/.changeset/auto-pm-drains-queue.md
@@ -1,0 +1,15 @@
+---
+'@gemstack/framework': patch
+---
+
+Auto PM now drains the agent queue as well as filling it (#855)
+
+Auto PM only ever put work into `TODO_AGENTS.md`, and nothing unattended took it out
+again: the backlog loop is a phase inside a run a human started, and auto PM's own runs
+go down the prompt path, which never reaches it. So the queue filled once and every
+later tick refused because it was no longer empty, and the daemon went quiet for good.
+
+A tick that finds open entries now starts a run for the first one instead of standing
+down, and goes back to harvesting quick-wins and planning tickets once the queue is dry.
+A queue that cannot be read at all is still a refusal. The refusal reason is logged each
+tick, so a wedged sweep no longer looks the same as a healthy idle one.
--- a/packages/framework/prompts/presets/drain_queue.md
+++ b/packages/framework/prompts/presets/drain_queue.md
@@ -1,0 +1,1 @@
+Open TODO_AGENTS.md and work on the FIRST open entry only, then check it off. Do not start any other entry.
--- a/packages/framework/src/auto-pm.test.ts
+++ b/packages/framework/src/auto-pm.test.ts
@@ -6,6 +6,7 @@ import {
   startAutoPm,
   DEFAULT_MIN_FREE_PERCENT,
   AUTO_PM_JOBS,
+  AUTO_PM_DRAIN_JOB,
   type AutoPmDeps,
   type AutoPmJob,
   type AutoPmProject,
@@ -36,7 +37,7 @@ function status(weekPercent: number | undefined, accountWeek: number | undefined
 const IDLE = { enabled: true, backlogEmpty: true, activeRuns: 0, quota: status(1) } as const
 
 test('autoPmDecision starts when the queue is dry and the budget is barely touched (#685)', () => {
-  assert.deepEqual(autoPmDecision(IDLE), { start: true })
+  assert.deepEqual(autoPmDecision(IDLE), { start: true, mode: 'pm' })
 })
 
 test('autoPmDecision does nothing while the preference is off (#685)', () => {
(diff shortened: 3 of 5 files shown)


suleimansh:
Live-verified the fix on an isolated daemon (scratch repo, own XDG_CONFIG_HOME, port 4399).

Tick 1:
```
auto PM: draining the first open queue entry in .../tf-scratch
```
The run fixed `greet.js` (`Helo` -> `Hello`, verified by executing it), checked off the first entry only, and left the second open.

Tick 2:
```
auto PM: landed the queue from 1 run(s) in .../tf-scratch
```
Checkout went from 2 open entries to 1, commit `[The Framework] queue updates from 2026-07-20T01-49-54-032Z`.

That is the loop turning by itself, which it has never done before: on main the equivalent tick fired and did nothing at all.

One thing this surfaced that is worth knowing: the agent renames its own branch mid-run (`set-session-name`), so the run's branch is not the one it started on. `promoteQueue` reads the branch off the run record in the checkout, which tracks the rename, so it promotes correctly. Anything that assumes the original run-id branch name would silently promote a stale file.

Issue #848, opened 2026-07-20
Auto PM spends the last of the quota after a daemon restart


suleimansh:
The consumption meter is delta-based: it measures how much usage has gone *up* since it started watching, not what the account has actually spent. A restarted daemon starts watching from scratch, so its first reading has nothing to compare against and the window reports **0 consumed** rather than "unknown".

Auto PM (#685) reads that as a full budget and starts spending.

Reproduced against the live driver on main (`0f16b3e`):

```
account week actually used : 95%
daily  -> usedPercent 0 | complete false | consumed 0
5h     -> usedPercent 0 | complete false
reached: null
AUTO PM DECISION: {"start":true}
```

The account has 5% of its week left and auto PM says go.

## Why the existing guard misses it

`quotaHeadroom` fails closed on `usedPercent === undefined`, which is the right instinct, but an empty meter never reports `undefined` — it reports `0`. Unknown and zero are indistinguishable to that check and mean opposite things. `LimitStatus.complete` already carries the honest signal (`false` = this figure does not cover the window) and is not consulted.

Worst part is the timing: the meter is emptied by a restart, and restarts most often follow a long heavy session — exactly when the account has least left.

## Fix

Gate on the account's own weekly window as well. That figure is absolute and survives a restart, unlike the delta meter, and it is already in `QuotaView.windows` (`kind: 'week'`) that the daemon reads every tick.

- refuse when there is no absolute weekly reading (fail closed, for real this time)
- refuse when the week is already more than `minFreePercent` spent
- keep the existing rolling-window checks on top

Open question for the threshold: reusing the existing 50% means auto PM goes quiet for the back half of a heavy week. Safe, possibly too conservative. Tunable, and worth a second opinion.

Issue #519, opened 2026-07-15
Consumption limits


brillout:
> usage guard: start now?

Yes, that would be great.

If it's a quick-win, I'd say three limits:
- Session consumption limit
- 5h consumption limit
- 24h consumption limit

And, ideally, these limits are set as `%`:

- Session consumption limit (default value: 40% of daily limit)
- 5h consumption limit (default value: 60% of daily limit)
- Daily (24h) consumption limit (default value: 20% of weekly limit)

Note that the limits are measured either against the daily or weekly limit.

Also, these limits can be enabled/disabled. So two UI elements: a checkbox (disable/enable), and a progress bar (%).

I deliberately use the term "consumption limit" instead of "guard", because I think it's easier to understand? (I like the term guard in itslef, but it might be a bit ambigious depending on the context.)

suleimansh:
Good news: we can have your bars, and they cost nothing.

Claude Code has a `/usage` command. We can run it and read the answer: session and weekly, percent used and reset time, exactly what your screenshot shows. It costs no tokens, and we never touch the user's login, because Claude Code asks Anthropic itself. Same source Traycer uses.

One correction to the spec: **there's no daily limit.** Claude measures a 5 hour session and a week, nothing per day. Your screenshot shows that too, session and weekly, no day. So your three limits become two:

- Session limit, default 40%
- Weekly limit, default 20%

Each meaning "pause background work once it has eaten this much of the window", with the checkbox and the bar you drew. Does that cover what your third limit was for?

Two small things I'll design around: the numbers come back as text, so a Claude Code update could reword them and we'd have to keep up. And Anthropic pushes back if you ask too often, so we ask on a slow timer and cache the answer.

brillout:
Nice!

With daily limit I meant the (artificial) daily consumption limit set by the user.

brillout:
> pause background work

For the session limit, maybe also send a "Pause" instruction (with a new TODO.md entry `Resume ${{ tf.session_name }}`).

But let me think a bit more about it.

brillout:
> also send a "Pause" instruction (with a new TODO.md entry `Resume ${{ tf.session_name }}`).

Actually, I like it because it will auto-resume anyways. So I'd suggest we pause the current session(s) for all limits.

As always, we can make it post-MVP if it's complex to implement.

suleimansh:
[...]

PR #871, opened 2026-07-20
fix(framework): gate auto PM on the configured usage limits (#870)

### LINKED ISSUE #870: Auto PM should gate on the configured usage limits, not its own headroom threshold
From @brillout on #839:

> How about we simplify and re-use the same global usage quota limits instead of introducing new ones?
>
> Is there a use case for having two different limits (one for auto-PM and one for other autopilot features)? I don't think so? Maybe it's overkill?
>
> So I'd suggest we replace the "when there's plenty of token left to spent" part with "when the usage quota limits aren't met".

Agreed, and it maps onto something that already exists: `ConsumptionStatus.reached` is documented as "the limit to pause for, or `null`", which is exactly "the usage quota limits are met".

## What changes

- `DEFAULT_MIN_FREE_PERCENT` (50) and the `minFreePercent` override go away. Auto PM stops having a budget notion of its own.
- `quotaHeadroom` refuses when `status.reached` names a window, and otherwise starts.
- It keeps failing closed on a quota it cannot read at all. That is a separate property from the threshold (#519: an unreadable quota must never *stop* the user's own work, but work nobody asked for is the other way round), and is not what this issue changes.

## The consequence worth deciding on

The 50% rule was doing a second job: since #848 the gate also checked the account's own absolute weekly figure, because the rolling meter is delta-based and a restarted daemon honestly reports 0% consumed no matter what the account has spent.

With the threshold gone, that protection goes with it. A daemon restarted at 95% of the account's week sees an empty meter, no limit `reached`, and starts work nobody asked for. It only bites when the user has no rolling limits configured, which is the default.

That lands close to the two-tier idea in the same comment:

> Soft => stop autopilot (the current limit settings)
> Hard => stop all work

A soft limit that stops unattended work while leaving the user's own runs alone is exactly what auto PM wants. Implementing the simplification as asked; flagging this so it is a decision rather than a silent regression.

suleimansh:
Closes #870. Implements @brillout's call on #839:

> So I'd suggest we replace the "when there's plenty of token left to spent" part with "when the usage quota limits aren't met".

Agreed, and there was no use case for two limits. It maps onto something that already existed: `ConsumptionStatus.reached` is documented as "the limit to pause for, or `null`", widest-first, which is exactly "the usage quota limits are met".

## What changed

- `DEFAULT_MIN_FREE_PERCENT` (50) and the `minFreePercent` override are gone. Auto PM has no budget notion of its own any more.
- `quotaHeadroom(quota)` starts while nothing is `reached`, and stands down naming the window when something is.
- It still fails closed on a quota it cannot read at all. That is a separate property from the threshold (#519: an unreadable quota must never stop *your* work, but it does stop work nobody asked for), so it stayed.

Switching every rolling limit off now lets auto PM run, which is the point: that is you saying "do not pace me". It used to refuse anyway, on its own invisible rule.

## The part worth your decision

Removing the threshold also removes what #848 added. That fix checked the account's absolute weekly figure, but only ever as a percentage against the 50%-free rule, so it could not survive the rule.

The blind spot it covered is real: the meter is delta-based, so a restarted daemon has nothing to diff and honestly reports **0 consumed** however much the account has spent. With the default limits configured that is fine. With the limits off, nothing stands between a fresh daemon and work nobody asked for.

I did not quietly keep a threshold of my own to paper over it, since that is the thing you asked me to remove. There is a test pinning the new behaviour so it is recorded rather than rediscovered.

This lands close to your own next paragraph:

> Soft => stop autopilot (the current limit settings)
> Hard => stop all work

A soft limit that stops unattended work while leaving your own runs alone is exactly what auto PM wants. Say the word and I will do that as its own ticket.

## Verified

Framework auto-pm suite 25 pass (rewritten around the new gate: starts while nothing is reached, stands down naming the window, starts with the limits off, still refuses an unreadable quota, and the restart blind spot). consumption, consumption-guard and client suites green. Root build 11/11.

### DIFF
--- a/.changeset/fix-870-auto-pm-configured-limits.md
+++ b/.changeset/fix-870-auto-pm-configured-limits.md
@@ -1,0 +1,9 @@
+---
+"@gemstack/framework": patch
+---
+
+Auto PM now paces itself by your own usage limits (#870). It used to keep a second budget rule of its own, refusing unless half of every window was still free, so there were two sets of limits to reason about and the stricter one was invisible. It now runs while your configured limits are not met and stands down once one is, which is the same line autopilot already stops at.
+
+It still refuses to start anything when the quota cannot be read at all: an unreadable budget must never stop your own work, but it does stop work nobody asked for.
+
+The `DEFAULT_MIN_FREE_PERCENT` export and the `minFreePercent` override are gone with the rule.
--- a/packages/framework/src/auto-pm.test.ts
+++ b/packages/framework/src/auto-pm.test.ts
@@ -4,7 +4,6 @@ import {
   autoPmDecision,
   quotaHeadroom,
   startAutoPm,
-  DEFAULT_MIN_FREE_PERCENT,
   AUTO_PM_JOBS,
   AUTO_PM_DRAIN_JOB,
   type AutoPmDeps,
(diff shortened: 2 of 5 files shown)


brillout:
> The blind spot it covered is real: the meter is delta-based, so a restarted daemon has nothing to diff and honestly reports 

Don't we retrieve the real quota usage from CC?

Issue #917, opened 2026-07-20
[The Framework] Record the originating surface on a conversation turn (via: discord)


suleimansh:
## TLDR

Carry the transport through `control.jsonl` so a committed conversation records `via: discord` instead of the local surface.

## Why

#908 records each chat turn with a `via` field naming the surface it came through, and #680 routes Discord messages into a run via the control channel. But `ControlEntry`'s `message` kind carries only `{ kind: 'message', text }`, so by the time the run records the turn the origin is gone and it is written as the local surface (`dashboard`/`cli`).

The conversation is therefore correct in content and wrong in attribution: a chat that happened in Discord reads as if it happened in the dashboard. That matters as more surfaces arrive (#606 explicitly plans Slack and others).

## Rough shape

- Add an optional `via?: string` to the `message` control entry (`control.ts:21-27`), keeping `isControlEntry` tolerant of its absence so older entries still parse.
- `sendMessage(projectId, text, runId?)` gains an optional origin; the Discord bot passes `'discord'`, the dashboard keeps its default.
- Thread it to the `recordMessage` call in `run.ts` so it lands in the conversation file.
- The agent's reply inherits the same `via` as the message it answers.

## Related

- #908 — the conversation record.
- #680 — the first non-dashboard surface.

Issue #908, opened 2026-07-20
[The Framework] Git as database: commit conversations to .the-framework/conversations/


suleimansh:
## TLDR

Commit conversations into the Git repo as `.the-framework/conversations/<runId>.md`, keyed to the run id that `LOGS.md` already records. This is the "conversations" half of #857, and the storage layer #680 needs before a Discord bot can have anywhere to put chat history.

## Why now

#857 asks for as much as possible inside the Git repo, naming conversations (#680) and the session list. The session-list half shipped: `LOGS.md` is unforgeable (#897) and is now the complete, joinable list of sessions carrying run id / session name / branch (#898).

The conversation half is still missing, and it is the blocker under #680 — its one stated open question is *where to save the chat history*, with "ideally inside the Git repo" as the preferred answer. Answering it transport-agnostically means the Discord bot inherits persistence rather than inventing its own, and it is worth doing on its own merits: the dashboard live chat (#714) already produces real conversations today and loses every one of them, because everything under `.the-framework/` except `LOGS.md` is transient by design (#313).

## What a conversation is

Exactly the human turns and the agent's replies — the thing a person would want to reread. Not the verbose transcript; #857 is explicit that the tool-call-level log stays with the model provider and is not ours to commit.

Both already exist as discrete events in `runChatPhase` (`run.ts:733-751`): each user message is emitted as `You: <message>`, and each `session.prompt()` returns the reply text. So this reads off the existing event stream and adds no capture machinery.

## Design decisions

**One file per conversation, append-only.** `.the-framework/conversations/<runId>.md`. A single shared file would be a merge-conflict generator: run worktrees are live concurrently and `commitPendingWork()` (`store/worktree.ts:135`) already auto-commits inside each one. Per-file append-only keeps concurrent runs from ever touching the same path.

**Keyed by run id.** That is the join key `LOGS.md` gained in #898 (`- run:`). `LOGS.md` stays the session list; `conversations/<runId>.md` is that session's chat. Both committed, joinable, no database.

**Body escaping, not body encoding.** `LOGS.md` collapses free text to one line via `encodeField` (`logs.ts:68`) because it is line-parsed — correct there, wrong for a transcript, where a multi-paragraph agent reply must stay readable in a `git diff`. Instead keep the text raw and escape only lines that could forge structure (a leading `#` or `\`), stripped on parse. Same #897 threat model, minimal readability cost, since replies rarely begin a line with `#`.

[...]

Issue #680, opened 2026-07-18
Discord chatbot


brillout:
I think this can be a quick-win?
- If user hits `Cltr+C` => chatbot goes offline
- I think the (only?) non-obvious part is where to save the chat history
  - I'd say ideally inside the Git repo
  - Otherwise, if using the Git repo is problematic, then maybe simply ~/.config/the-framework/cache/discord for now

It's a nice combo with the new agentic PM features we have.

FYI @nitedani

Issue #606, opened 2026-07-16
Epic: Collaboration layer — agents as teammates in the comms layer


suleimansh:
> Tracked-but-later (**Phase 1**). Do not build before the MVP ships. This captures a team-discussion direction so it is not lost; it is not scheduled work yet.

## Vision

Agents get real identities and presence and live in the company's communication layer (chat, threads, later voice/video) as peers, not as a passive sidebar assistant. You add an agent to a channel, `@mention` it, or DM it. A manager agent reads a channel, DMs specialist agents, they coordinate, reply in-channel, and maintain the knowledge base, tasks, and issues. This is the "then extend in this direction" step that sits on top of the autonomous core, not instead of it.

## Decided direction

- Keep the integration seam simple and generic so Slack / Discord / others are all easy to add. Start with a single bot user that acts as a channel for multiple agent personas (no per-persona accounts needed).
- Realtime via Telefunc (Room API).
- Agentic-PM entry point: a team chats to consensus, then `/plan-from-chat` (or "@ai create a plan from this chat") turns the discussion into a plan/queue. Nudge the agent to be succinct.

## Open questions

- Multi-user chat mapping: how per-user TODO/state maps when one bot user serves many humans.
- Voice/video participation (hosted-only, needs webrtc infra).
- Compete vs collaborate with the main competitor (affects how much of this we build ourselves).

## Related

- Depends on #605 (daemon + gateway split) for where the agents actually run.
- #454 (syncing UI <-> data): the realtime / source-of-truth substrate this rides on.
- #462 (ticketing / PM): the tasks and issues agents maintain from chat.

brillout:
I suggest we postpone this for now.

PR #933, opened 2026-07-20
feat(framework): send a session's answers back to Discord (#932)

### LINKED ISSUE #932: [The Framework] The agent's reply never reaches Discord
## TLDR

When a session answers, post the reply into the Discord channel it was asked from. Today the bot acknowledges a message and nothing else comes back, so you cannot hold a conversation without opening the dashboard.

## What is actually there today

Checked every post-to-Discord call site in `src/discord/`. Only two things ever reach a channel:

1. **Canned acks from `bot.ts`** -- `Sent to the running session.`, `Starting a session in **X**.`, `Stopping the session.` These are `action.reply` strings from `routing.ts`; none contains agent output.
2. **The #627 notification watchers**, over the separate `DISCORD_WEBHOOK`: run started / finished / needs-you. Transitions, not content.

`!status` reports what is running plus a parked gate (`snapshotLiveRun`), which is state, not the answer.

So the integration is **write-only**: you can talk to the agent from Discord and never see what it said. #680 delivered inbound; this is the outbound half.

## Why the conversation file is the right source

The events log has no "agent reply" kind -- `log` carries rendered console lines and `driver` carries raw driver events, neither of which is the settled answer.

`.the-framework/conversations/<runId>.md` (#908) records exactly `{ at, role, via, text }` per turn, and the text is deliberately *the settled reply the user actually read*. Since #917 the agent's reply also inherits the origin of the message it answers, so **an agent turn with `via: discord` is precisely "an answer owed to Discord"**. No new bookkeeping about what counts as a reply.

## Rough shape

- Bind `runId -> channelId` when the bot starts or messages a run (it has `message.channelId` in hand). In memory is fine for a first cut; a daemon restart just stops mirroring, it does not misdeliver.
- A watcher alongside the other daemon services (injectable seams, unref'd timer, swallows errors) polls the bound runs' conversations for agent turns whose `via` is the bot's surface.
- **Baseline on first sight**, the same rule `activity-watcher.ts` uses: adopt the current transcript length rather than replaying it, so enabling the bot never floods a channel with a backlog.
- Post with the existing `postMessage` + `clampContent` (Discord hard-rejects over 2000 chars), replying onto the asking message so a busy channel stays threaded.
- Drop the binding when the run ends.

## Out of scope

Streaming partial output, and mirroring a dashboard-started session into Discord (that needs a channel to mirror *to*, which is a #606 routing question).

## Related

- #680 -- delivered the inbound half.
- #917 -- makes an agent turn say which surface it owes its answer to.
- #606 -- the collaboration-layer direction this serves.
- `tickets/2026-07-18_discord-two-way-gate-replies.md` is already implemented by #680 (`decideAction` returns a `choice` kind and the bot wires `sendChoice`) and can be deleted.

suleimansh:
Closes #932

## What

A session's answers now reach the Discord channel that asked for them. #680 delivered the inbound half; the bot only ever acked (`Sent to the running session.`) and the reply itself lived in the dashboard. The integration was write-only: you could talk to the agent and never hear it.

## Why the conversation, not the event log

`events.jsonl` has no "agent reply" kind. `log` carries rendered console lines and `driver` carries raw driver events, neither of which is the settled answer. `.the-framework/conversations/<runId>.md` (#908) records exactly the text the user would have read, which is what belongs in a chat channel.

## Design

| | |
|---|---|
| **Which channel** | a binding, not a guess. The bot records `runId -> channelId` when it starts or messages a run, the only moment the channel is known |
| **Unaddressed runs** | not mirrored at all. Routing a dashboard-started session into some channel is #606's job |
| **Backlog** | binding adopts the transcript, so attaching to an hour-old session does not dump an hour of answers into the channel |
| **Bot switched off** | the cursor still advances without posting, so turning it on starts from now (same contract as the notification watchers) |
| **Delivery** | existing `postMessage` + `clampContent` (Discord hard-rejects over 2000 chars) |

**The baseline is taken at bind, not at first poll.** That is the one subtle bit: deferring it to the first tick would swallow a reply from a fast agent that answered before the tick. The bot awaits `bind()` *before* handing the run the message, which makes the answer reliably new. There is a test for exactly this race.

## Correction to the issue

The issue proposed filtering on `via: discord` from #917. That turned out to be unnecessary: the binding already decides where to post, so `via` adds nothing here, and dropping it means this does **not** depend on unmerged #931 and works on `main` today.

## Verification

Full suite: **1007 tests, 0 failures**; `tsc --noEmit` clean. 13 unit tests covering the backlog rule, the pre-first-poll race, the off-switch cursor, dedup across polls, per-run channel isolation, and failure paths.

Real e2e driving the built `dist`: a real `framework` run writes a real transcript, and the built mirror reads that file and posts from it (Discord itself stubbed).

```
PASS  the run reached its chat phase
PASS  binding adopts the transcript so far, posting nothing
PASS  the agent answer was posted to the bound channel
PASS  it went to the channel that asked
PASS  the posted text is the agent reply, not the user message
PASS  re-polling does not repost the same answer
```

### Not yet verified

**Against live Discord.** The gateway path is exercised only by fakes here. Token + MESSAGE CONTENT intent have been confirmed working out-of-band; a live end-to-end run against a real server is the remaining check before this leaves draft.

### DIFF
--- a/.changeset/discord-agent-replies.md
+++ b/.changeset/discord-agent-replies.md
@@ -1,0 +1,5 @@
+---
+'@gemstack/framework': minor
+---
+
+A session's answers now come back to the Discord channel that asked for them. Until now the bot acknowledged a message and nothing else followed, so you could talk to an agent from Discord but had to open the dashboard to read what it said. The bot binds a run to the channel it was addressed from, and a watcher posts each new agent turn there as it lands, reusing the committed conversation as the source since that holds the settled reply rather than raw console output. Binding adopts whatever the session has already said, so attaching to a long-running session never replays its backlog into a channel, and while the bot is switched off the cursor still advances, so turning it on starts from now rather than flushing everything said meanwhile. A run nobody addressed from Discord is not mirrored at all.
--- a/packages/framework/src/daemon.ts
+++ b/packages/framework/src/daemon.ts
@@ -46,7 +46,10 @@ import { loadFrameworkConfig } from './config.js'
 import { installProject, enumerateGitRepos } from './install.js'
 import { JsonlTailer } from './jsonl-tail.js'
 import { startDiscordBot, DISCORD_VIA } from './discord/bot.js'
+import { startDiscordReplyMirror } from './discord/reply-mirror.js'
 import { snapshotLiveRun } from './discord/live-run.js'
+import { readConversation } from './conversations.js'
+import { postMessage } from './discord/rest.js'
 import { sendChoice, sendMessage, sendStop } from './dashboard-rpc/control.telefunc.js'
 
 /**
@@ -668,6 +671,26 @@ export async function runDaemon(cwd: string, opts: RunDaemonOptions = {}): Promi
   // picker, so a message with no run to join starts one here.
   const botHomeId = projectId(resolve(cwd))
   const botProjectPath = async (id: string) => (await listSummaries()).find(p => p.id === id)?.path ?? cwd
+  // Send a session's answers back to the channel that asked (#932). The conversation (#908) is the
+  // source: it holds the settled reply the user would have read, which is what belongs in chat.
+  // Bound per run by the bot, since the channel is only known when a message arrives.
+  const replyMirror = botToken
+    ? startDiscordReplyMirror({
+        readConversation: async runId => {
+          // A run's transcript lives in the checkout the run used, which for a daemon-spawned run
+          // is its own worktree rather than the project root.
+          for (const project of await listSummaries()) {
+            const meta = (await readLiveMetas(project.path).catch(() => [])).find(run => run.id === runId)
+            if (meta) return readConversation(meta.cwd, runId).catch(() => [])
+          }
+          return []
+        },
+        post: (channelId, text) => postMessage(botToken, channelId, text),
+        enabled: async () => (await readPrefs()).discordBot === true,
+        onLog: message => console.log(`[framework] ${message}`),
+      })
+    : undefined
+
   const discordBot = botToken
     ? startDiscordBot({
         token: botToken,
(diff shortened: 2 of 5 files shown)


suleimansh:
## Verified against live Discord

Ran against a real bot in a real server (isolated config + scratch project, so nothing touched the real registry).

**Inbound** (#680's half, re-confirmed on this branch): `!help` returned the command list threaded onto the asking message, `!status` correctly reported `Nothing running in live-repo`. That exercises gateway -> MESSAGE CONTENT intent -> routing -> REST reply.

**Outbound** (this PR): the built mirror posting to the real channel through the real token.

```
backlog suppressed:      PASS (0 posted)
agent reply posted:      PASS (1)
user turn not echoed:    PASS
no duplicate on re-poll: PASS (1)
```

Driven by feeding the mirror a transcript directly rather than starting an agent, so it costs no quota while still exercising the real REST path end to end. The run-produces-turns half is covered by the earlier e2e against a real `framework` run and its real conversation file.

Removes the "not yet verified against live Discord" caveat in the description.

### Still worth knowing

A **full** path (message in Discord -> real agent run -> its reply mirrored back) has not been run, because it spends real quota. Everything in that chain is covered by one of the two tests above, but they meet in the middle rather than overlapping.

suleimansh:
## Full chain verified live, with a real agent

The remaining gap (Discord message -> real agent run -> reply mirrored back) is now closed, confirmed by the user in the channel.

Better than the scripted test: the run was started from the **dashboard** and then spoken to from **Discord**, which is the mixed-surface case the unit tests could only fake. One real transcript, committed by the run's worktree:

```
## 2026-07-20T23:01:25.805Z · user  · dashboard
## 2026-07-20T23:01:45.899Z · agent · dashboard
## 2026-07-20T23:03:58.701Z · user  · discord
## 2026-07-20T23:04:04.893Z · agent · discord
```

- The Discord turn is attributed to Discord, and **the agent's reply inherited that origin** (#917's rule, previously only fake-tested).
- The dashboard turns kept saying `dashboard`: per-turn attribution, not a per-run flag.
- That `23:04:04` agent turn is verbatim what arrived in the channel -- this PR's path.
- Committed at `cf78ca5` on `the-framework/run-2026-07-20T23-01-25-504Z` with no git command from anyone.

Nothing about this PR is unverified now: unit (13 tests), e2e against a real run's transcript, live outbound to a real channel, and the full chain with a real agent.

PR #949, opened 2026-07-21
fix(framework): a malformed request can no longer kill the daemon, relay, or preview server

### LINKED ISSUE #938: A malformed percent-encoded request can take the daemon down (static.ts + preview.ts)

Two static-file paths call decodeURIComponent on the raw request path with no try/catch:

- packages/framework/src/dashboard/static.ts:25, dispatched via `void serveClientBundle(...)` in server.ts (lines 134/138) with no rejection guard
- packages/framework/src/preview.ts serveStaticFile (line ~335), dispatched the same way on its own port

`GET /%zz` throws URIError inside an async function whose rejection nothing handles; Node's default policy takes the process down, so one malformed request kills the daemon (or a preview server). telefunc-serve.ts wraps its path handling in try/catch for exactly this class of failure ("never let a failure become an unhandled rejection that kills the daemon"); the static branches got no such care.

Fix: try/catch the decode (fall through to index.html / respond 400), or wrap both void-dispatched handlers. Also worth normalizing a trailing-slash cwd in serveStaticFile's prefix check while there.

suleimansh:
Closes #938

Verified at head of main before fixing: a live daemon (node dist/bin.js --daemon-serve) died to a single GET /%zz, and separately to an absolute-form request target (GET http://[ HTTP/1.1), which Node passes through verbatim and new URL(req.url, ...) throws on synchronously. Both were reproduced with curl / a raw socket against a real daemon.

Three crash sites of the same class, all fixed:

- packages/framework/src/dashboard/static.ts: decodeURIComponent on the raw path in the void-dispatched bundle handler. A malformed escape now falls back to the SPA shell like any unknown path.
- packages/framework/src/preview.ts serveStaticFile: same unguarded decode on its own port; now answers 400. While there, the path-prefix check normalizes a trailing-slash cwd, which previously made it refuse every path (issue OP called this out).
- packages/framework/src/relay.ts + dashboard/server.ts: the synchronous new URL(req.url, ...) parse, plus the relay run-id decodeURIComponent (/r/%zz/). One shared requestPathname() helper parses defensively; unparseable targets answer 400. The relay one matters most: it is the public surface.

Regression tests (each fails without the fix; mutation-checked by reverting the fix files and re-running exactly these):

- dashboard/server.test.ts: GET /%zz serves the SPA shell and the server keeps serving; a raw-socket absolute-form target gets 400 and the server survives.
- preview.test.ts: GET /%zz gets 400 and the server keeps serving; a trailing-slash cwd serves instead of refusing every path.
- relay.test.ts: GET /r/%zz/ gets 400 and healthz still answers.

Live-verified against a real daemon on a scratch repo: GET /%zz answers 200, the absolute-form target answers 400, and the daemon serves normally afterwards.

### DIFF
--- a/.changeset/static-decode-daemon-safe.md
+++ b/.changeset/static-decode-daemon-safe.md
@@ -1,0 +1,11 @@
+---
+'@gemstack/framework': patch
+---
+
+A malformed percent-encoded request no longer kills the daemon or a preview server (#938)
+
+Both static-file paths (the dashboard bundle server and the preview fallback server) called
+`decodeURIComponent` on the raw request path inside a void-dispatched async handler, so a single
+`GET /%zz` became an unhandled rejection that took the whole process down. The dashboard now
+treats a malformed escape as an unknown path (the SPA shell), the preview server answers 400, and
+a trailing-slash cwd no longer fails the preview server's path-prefix check.
--- a/.changeset/static-decode-daemon-safe.md
+++ b/.changeset/static-decode-daemon-safe.md
@@ -2,10 +2,13 @@
 '@gemstack/framework': patch
 ---
 
-A malformed percent-encoded request no longer kills the daemon or a preview server (#938)
+A malformed request can no longer kill the daemon, the relay, or a preview server (#938)
 
-Both static-file paths (the dashboard bundle server and the preview fallback server) called
-`decodeURIComponent` on the raw request path inside a void-dispatched async handler, so a single
-`GET /%zz` became an unhandled rejection that took the whole process down. The dashboard now
-treats a malformed escape as an unknown path (the SPA shell), the preview server answers 400, and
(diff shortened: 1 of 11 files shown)


suleimansh:
Round-2 adversarial pass live-reproduced a miss in this branch: GET /browser/p/%zz/stream passed the new pathname guard (the URL parses; %zz explodes only at decode time inside parseBrowserRoute) and killed the daemon through the uncaught proxy dispatch. Fixed in the follow-up commit: parseBrowserRoute parses and decodes in a try/catch (unparseable falls through to the bundle), and the proxy dispatch got a .catch as belt and braces. New unit + HTTP tests cover the path; re-verified against a live daemon with all three attack shapes, which now answer 200/200/400 with the daemon serving afterwards.

PR #954, opened 2026-07-21
fix(framework): release reply-mirror bindings once their run is gone

### LINKED ISSUE #941: Reply-mirror bindings are never released

discord/reply-mirror.ts exports unbind/isBound, but the only production caller is daemon-services.ts (`onRunBound: ... replyMirror.bind(...)`) - nothing ever unbinds. Every chat-touched run stays in the map for the daemon's lifetime, and each 3s poll then scans every project's live metas per bound run (the readConversation lookup in daemon-services). No misposts (archived runs resolve to []), but the per-poll IO grows monotonically.

Fix: unbind when the run archives, or have the poll drop a binding whose run no longer resolves.

suleimansh:
Closes #941

Verified at head of main: reply-mirror exports unbind/isBound but the only production caller is daemon-services (bind on onRunBound); nothing ever unbinds. Every chat-touched run stays in the map for the daemon lifetime, and each 3s poll then scans every project live metas per bound run. No misposts (archived runs resolve to []), but per-poll IO grows monotonically, exactly as filed.

Fix, the poll-drops-a-dead-binding option from the OP:

- daemon-services readConversation now answers undefined when the run has no live meta in any project (archived, or its project removed); a throwing read still resolves [].
- The mirror counts consecutive unresolvable polls per binding and releases it after UNBIND_AFTER_MISSES (5 polls, 15s at the 3s cadence), logging the release. The grace window exists for the freshly bound run whose live meta is not on disk yet; a transient miss resets on the next successful read.

Regression tests (mutation-checked: flattening undefined back to [] in the poll, the pre-fix behavior, fails exactly the release test):

- discord/reply-mirror.test.ts: a run that stops resolving is released after exactly UNBIND_AFTER_MISSES consecutive misses, with a log line.
- discord/reply-mirror.test.ts: a single transient miss keeps the binding, posting resumes, and the miss counter resets.

### DIFF
--- a/.changeset/reply-mirror-release-bindings.md
+++ b/.changeset/reply-mirror-release-bindings.md
@@ -1,0 +1,12 @@
+---
+'@gemstack/framework': patch
+---
+
+Reply-mirror bindings are released once their run is gone (#941)
+
+Nothing ever unbound a chat-touched run, so every binding stayed in the map for the daemon's
+lifetime and each 3s poll scanned every project's live metas per bound run — IO that only ever
+grew. The daemon's conversation reader now answers `undefined` for a run with no live meta
+anywhere (archived, or its project removed), and after a few consecutive misses the mirror drops
+the binding and logs it. A transient miss (or a throwing read) still costs one poll, not the
+binding, and a fresh bind gets the same grace against the meta-not-yet-on-disk race.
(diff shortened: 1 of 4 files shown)


suleimansh:
Round-2 adversarial pass caught one real (low-probability) issue, fixed in the follow-up commit: the daemon flattened a registry/FS read outage into "run gone", so 15s of unavailability could unbind a live chat. undefined is now reserved for "listings worked and the run is not there"; outages resolve [] and keep the binding. Grace widened to 10 polls for slow-booting children.

PR #956, opened 2026-07-21
fix(framework): let BrowserPanel recover from a failed stream instead of latching

### LINKED ISSUE #946: BrowserPanel latches "not reachable" until remount

framework-dashboard/components/BrowserPanel.tsx: one img onError (e.g. opening the tab before the run's stream endpoint is up) permanently swaps in the "not reachable" message; only a remount recovers. The right-rail tab switch happens to remount it, but that is the parent's accident, not this component's contract. Fix: a Retry affordance, or reset the failed flag keyed off runId/stream identity.

suleimansh:
Closes #946

Verified at head of main: one img onError in BrowserPanel (e.g. opening the Browser tab before the run stream endpoint is up) permanently swaps in the "not reachable" message. Only a remount recovers, and only the right-rail tab switch happens to provide one; that is the parent accident, not this component contract.

Fix (framework-dashboard/components/BrowserPanel.tsx), both affordances from the OP:

- The failed flag is keyed to the exact stream identity it happened on (run base + attempt), so the parent switching runs without a remount starts clean.
- A Retry button bumps the attempt, which clears the latch and cache-busts the stream URL (?r=N; the daemon proxy routes on pathname only) so the browser re-issues the request instead of replaying the failed one.

Regression tests are state-level (jsdom has no layout engine, so no rendering claims), in BrowserPanel.test.tsx, mutation-checked (reverting the component fails all three):

- error swaps in the message + Retry; Retry restores the img on a distinct URL.
- a failed retry latches only that attempt and can be retried again.
- a failure on run A does not latch for run B on a prop change without remount.

No changeset: @gemstack/framework-dashboard is private.

### DIFF
--- a/.changeset/browserpanel-retry.md
+++ b/.changeset/browserpanel-retry.md
@@ -1,0 +1,10 @@
+---
+'@gemstack/framework': patch
+---
+
+The Browser pane recovers from a failed stream instead of latching "not reachable" (#946)
+
+One img error (e.g. opening the tab before the run's stream endpoint was up) permanently swapped
+in the failure message until a remount. The failure is now keyed to the exact stream it happened
+on, so switching runs starts clean, and a Retry button re-requests the stream. The fix lives in
+the dashboard client, which ships bundled inside this package.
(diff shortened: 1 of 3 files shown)


suleimansh:
Round-2 adversarial pass: added the @gemstack/framework patch changeset (the dashboard client ships bundled inside the framework package, so without it the fix never publishes) and fixed the flagged nit: returning to a previously failed run now retries instead of replaying the stale failure. New test covers the return path.

PR #1093, opened 2026-07-23
[AI] Remove docs/ site and its CI/workspace references

brillout:
It clashes with the GH action for the-framework.ai, can we remove it? (Assuming it was all AI generated anyways.)

Not sure if had meaningful content? Or maybe we move it to another repo?


suleimansh:
Will check this now

suleimansh:
Agreed, let's remove it. I backed up the full `docs/` tree first (25 md files; it wasn't all filler, the `ai-sdk/*` and `mcp-connectors/*` guides were real), so we can move it to its own repo later without losing anything.

Issue #1129, opened 2026-07-24
Topics: recommend gate for bind + read for list (#1121 mechanism)


suleimansh:
**Recommendation: gate for the bind/create, read for the list.** Both approaches were spiked end to end and are green; evidence below. For #1121, part of #1115.

## The call

- **`create_project` / bind a project -> await-gate.** The agent ends a turn with an `await-create-project` / `await-bind-project` block; the framework shows Approve/pick, resolves it, resumes.
- **`list_projects` -> a read**, not a gate: inject the registered projects into the topic run's context (add a read-only tool later only if the list gets large/dynamic).

The issue framed both as "tools", but a pure read and a permission-carrying respawn are different shapes. Splitting them is cleaner than forcing one mechanism on both.

## Why a gate for the bind

1. **This is a permission feature** ("the confirmation is the permission gate"). A gate IS that gate: Approve/Decline, asked once per new project, autopilot auto-accepts, all for free. Agents run in `bypassPermissions`, so an MCP tool call does not prompt the user; to get the required approval you would rebuild the gate round-trip anyway.
2. **Binding is a respawn.** The run moves to the project's worktree (#1122), so it is a turn-ending transition. A gate models that. A tool returns a value mid-turn into a run that is about to be torn down.

## Why a read (not a gate) for the list

A pure read is naturally tool/context-shaped; ending a turn to read a list is awkward.

## Evidence from the two spikes (both draft, both CI green)

| | tools (#1130) | gates (#1131) |
|---|---|---|
| footprint | 6 files / +193 | 10 files / +249 |
| tests | 1260 | 1266 |
| base-prompt edit | none | none (runtime append layer) |
| permission gate | must bolt on | native |
| respawn fit (#1122) | fire-and-effect | native |

Findings that settled earlier unknowns:
- **Neither needs the drift-guarded system prompt.** The gate protocol lives in the runtime append layer (`system-prompt.ts`), appended only for topic runs; `system_prompt.md` is untouched.
- **UI/UX is identical either way.** Both feed the same `bind` event + `RunMeta.boundProjectId`; the approval routes through the existing choice panel. The mechanism choice is internal, not user-facing.
- The bind-recording code is byte-identical across the two spike PRs, so the diff between them is purely the trigger.

## Next once confirmed

Promote the #1131 gates spike to the real #1121 (add list-as-context, harden, fuller tests, changeset), then build #1122 re-home (allocate the worktree + continue-run into the chosen project). Retire the #1130 tools spike.

[...]

Issue #1122, opened 2026-07-24
Topics: re-home a run's cwd/worktree on bind


suleimansh:
Part of #1115.

## What

On bind, allocate a worktree for the chosen project and continue the **same** conversation there. This is the one genuinely new engineering piece in the epic.

## Why it is the keystone

A run's project membership is **structural**, not a stored field: `RunMeta` has no project id or persisted cwd (`store/run-store.ts:69-150`) — a run simply is whichever `.the-framework/` directory it lives in. The detached child is spawned with `--cwd` baked into argv and there is **no re-homing anywhere today** (`daemon-runtime.ts:399-447`).

## Approach

Reuse the existing continue-run machinery (#762: `continueWorkspace`, `--continue-run`, `--resume-session`) to respawn the conversation in the newly allocated worktree, preserving the agent session so the chat continues unbroken.

## Delta to validate during build

Continue-run today re-homes into the **same** project's worktree; a topic must continue into a **newly chosen** project. That is the load-bearing detail to prove out.

Issue #1121, opened 2026-07-24
Topics: bind-project and create-project await-gates


suleimansh:
Part of #1115.

## What

Two new `await-gate` kinds so a project-less run can bind to a project: pick an existing one, or register a new one.

## Design call

Model these as **gates, not MCP tools**. Gates are pure text-protocol (the framework parses a fenced block from the agent's stdout and drives the choice round-trip itself), so they work on **both** the Claude and Codex drivers and autopilot handles them for free. Real MCP tools would be Claude-only (the Codex driver has no MCP wiring) and are the wrong shape here anyway, since a bind triggers a process respawn (see the re-home slice) so there is nothing to return into mid-turn.

## Scope

- `await-bind-project`: choose from registered projects. Reuses the existing choice round-trip. Binding to an already-registered project needs **no** permission gate (registry membership is the trust).
- `await-create-project`: register a new path. Its confirmation **is** the permission grant, reusing `await-confirmation` (#358). Asked once per new project.
- Anchors: `await-gate.ts`, `turn-gate.ts` `parseAwaitGate`, and `await-browser` / `browser.ts` as the precedent for a gate that grants a capability.

## Open

The parent frames these as "tools". If the tool-call ergonomics matter, that is the one thing to confirm before build; functionally the gate delivers the same "calling it is the bind, its confirmation is the permission gate".

Issue #1120, opened 2026-07-24
Topics: start a run with no project, in a neutral cwd


suleimansh:
Part of #1115.

## What

Let a run start with **no project**: a "topic" that lives in a neutral scratch cwd with no repo access, instead of defaulting to the daemon's home project.

## Scope

- `sendStart` accepts an absent/null project id meaning "topic".
- Today `onStart` forces a project: an absent id defaults to `homeId` (`daemon-runtime.ts:398`), then `resolveProject` requires a cwd (`daemon-runtime.ts:399`). A topic needs a genuine null-project branch and a neutral scratch cwd (under the config home).
- No worktree allocated, no repo files reachable, until a bind happens (see the bind-gates slice).

## Why first

Smallest real change and it unblocks the rest of the epic. The order-of-thinking and zero-project onboarding cases in the parent both start here.

PR #1130, opened 2026-07-24
spike(the-framework): MCP-tools bind for topics (#1121)

suleimansh:
Spike for #1121, part of the #1129 decision. **Draft — not for merge.** Evaluates the MCP-tools variant of the topic-to-project bind against the await-gate variant (spiked separately).

## What it does

A project-less "topic" run (#1120) now gets two MCP tools, so its agent binds to a project by calling a tool instead of parking on a gate:

- `list_projects` — returns the registered projects (reads the registry).
- `create_project` — takes a path, registers it (`addProject`), and binds this run to it.

The bind is signalled back to the run over the existing control channel: `create_project` appends a new `{ kind: 'bind', projectId }` control entry to the run's `control.jsonl`, the run's watcher folds it to a `bind` event, and that lands on `RunMeta.boundProjectId`. Same file-is-the-seam design the dashboard already steers with, so there is no new run<->subprocess IPC.

The worktree re-home / respawn that a real bind implies is **out of scope — that is #1122**; a comment marks where it goes.

### Wiring

- `src/projects-mcp.ts` (new) — the server + tools (authored on `@gemstack/mcp` directly), `projectsMcpServers` spawn spec, and `withProjectsMcp` (mirrors `withBrowser`).
- `src/cli.ts` — hidden `mcp-projects` subcommand (the spawnable server, reads the run cwd from `FRAMEWORK_RUN_CWD`); folds the server into topic runs' `claudeOpts.mcpServers`; handles the `bind` control entry.
- `src/control.ts` / `src/events.ts` / `src/store/run-store.ts` / `src/terminal.ts` — the `bind` control entry, the `bind` event, `RunMeta.boundProjectId`, and the terminal line.

No prompt change was needed — MCP advertises the tools itself.

### Proof

- 5 new tests (in-memory registry + real control-channel round-trip + meta fold + spawn spec): green. Suite **1260 pass / 0 fail** (+1 skipped), root typecheck clean.
- Drove the real subprocess over stdio (`node dist/bin.js mcp-projects`): `tools/list` returns `list_projects, create_project`; `create_project` returns the record with `bound: true`; the run's `control.jsonl` gets `{"kind":"bind","projectId":"smoke-142xwjl"}`.

## How it felt (verdict)

**Clean:**
- The spawn side is a near-exact copy of the `--browser` precedent (`withBrowser` -> `withProjectsMcp`, one `mcpServers` entry). The MCP plumbing (`--mcp-config` temp file, merge-not-replace) already existed, so wiring a second server was trivial.
- No prompt edit. The tools are self-describing, which is the real ergonomic win over a gate: the agent discovers "I can make a project" without us scripting the moment.
- The bind signal reuses the control channel verbatim — one new entry kind, one new event, one meta field. It slotted in exactly where `handoff-armed` already lives.

**Awkward:**
- **Reuse hit a naming wall.** The obvious reuse (`@gemstack/mcp-connectors` `defineConnector`) kebab-cases and namespaces tool names, and its id regex rejects `_`, so it cannot produce `create_project` verbatim (best it does is `projects_create-project`). I dropped to `@gemstack/mcp` `McpTool` directly to honor the names, which costs a small factory-closure-per-tool. Worth flagging for the decision: if we accept `create-project`, connectors is even less code.
- **The cross-process callback is genuinely indirect.** The tool runs in a *separate spawned process* from the run; it cannot return anything to the run loop. The only way back is to write the control file and have the run tail it. It works and is consistent with the rest of the framework, but "the tool call updated the run" is two processes and a file, not a function return.
- **The respawn nuance bites, and it is the crux.** Binding will (in #1122) re-home the run into the project's worktree, which means killing and respawning the run process. So the agent's `create_project` call returns into a turn that is about to be torn down — there is nothing to continue into mid-turn. The tool's return value is effectively discarded; the *side effect* (the bind) is what matters. This is a slightly unnatural fit for "the agent calls a tool and uses the result": here the agent calls a tool and then its whole process is replaced. A gate that the orchestrator resolves by respawning models that reality more directly. The tools approach still works (the bind is recorded, #1122 picks it up on respawn), but the "tool returns a value the agent reasons about" mental model does not hold at the bind moment.

**Recommendation:** The MCP-tools path is clean to build and gives the nicest discovery story (no prompt, self-describing, list + create in one surface). Its weak spot is exactly the bind semantics: the respawn means the tool return is throwaway, so we get the ergonomics of "call a tool" without the payoff of "use its result." If the #1129 decision weights *agent-facing discoverability*, pick tools. If it weights *modeling the respawn honestly*, the gate is the more faithful shape. My lean: **tools for discovery, but only if we are fine treating create_project as fire-and-effect** (bind recorded, run respawned by #1122) rather than a value-returning call. Keep the tool return minimal (it will be discarded) and let #1122 own the respawn.

## Codex note (not wired here)

On Codex the same server would ride the driver's `extraArgs` hatch (`src/driver/codex.ts`): `codex exec` takes `-c` config overrides, so a topic run would append
`-c mcp_servers.projects.command=<node>` `-c mcp_servers.projects.args=["<bin>","mcp-projects"]` (plus the `FRAMEWORK_RUN_CWD` / `XDG_CONFIG_HOME` env). The server binary is agent-agnostic — only the "how the CLI is told about it" differs (Claude's `--mcp-config` file vs Codex's `-c mcp_servers.*`). Not built here to keep the spike to one driver.

### DIFF
--- a/packages/the-framework/package.json
+++ b/packages/the-framework/package.json
@@ -63,8 +63,10 @@
   },
   "dependencies": {
     "@gemstack/ai-autopilot": "workspace:^",
+    "@gemstack/mcp": "workspace:^",
     "telefunc": "^0.2.22",
-    "yaml": "^2.5.0"
+    "yaml": "^2.5.0",
+    "zod": "^4.0.0"
   },
   "devDependencies": {
     "@types/jsdom": "^28.0.3",
--- a/packages/the-framework/src/cli.ts
+++ b/packages/the-framework/src/cli.ts
@@ -18,6 +18,7 @@ import { createRunDriver } from './run-driver.js'
 import { githubSlugFor } from './dashboard/github.js'
 import type { ActionsDriverOptions } from './driver/index.js'
 import { launchSharedBrowser, withBrowser, type SharedBrowser } from './browser.js'
+import { runProjectsMcp, withProjectsMcp, RUN_CWD_ENV } from './projects-mcp.js'
 import { connectCdp, startBrowserStream, type BrowserStream } from './browser-stream.js'
 import { hostExecutor } from './host-exec.js'
 import { startDashboard, singleProjectProvider, resolveDashboardBundle, type Dashboard } from './dashboard/index.js'
@@ -355,6 +356,8 @@ export interface CliOptions {
   host?: string | undefined
   dashboard: boolean
   relayServe: boolean
+  /** `mcp-projects` (#1121, spike): run the projects MCP stdio server. Internal — a topic run spawns it. */
+  mcpProjects?: boolean
   share?: string | undefined
   sessionLink?: string | undefined
   permissionMode?: PermissionMode | undefined
@@ -661,6 +664,9 @@ export function parseArgs(argv: string[]): CliOptions {
     words.shift()
   } else if (words[0] === 'relay') {
     opts.relayServe = true
+    words.shift()
+  } else if (words[0] === 'mcp-projects') {
+    opts.mcpProjects = true
     words.shift()
   } else if (words[0] === 'stop') {
     opts.stop = true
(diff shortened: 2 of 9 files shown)


suleimansh:
Superseded: per the #1129 recommendation we are going with the gate for the bind (permission model + respawn fit). This MCP-tools spike proved the approach and stays as reference. The real #1121 builds on the gate spike (#1131). Closing.

PR #1152, opened 2026-07-25
fix(dashboard): write only the preferences you changed

### LINKED ISSUE #1148: fix(dashboard): a stale tab reverts your theme (and every other preference) on its next write
Reported while testing the dashboard: change the theme, refresh, and the old theme is back.

The theme itself is not special. The dashboard writes **the whole global preferences object** on every change, from a cache each tab loads once and never refreshes, so any tab that has been open since before your change silently reverts it the next time you touch any setting in that tab. The theme is just the one that is impossible to miss, because you see it.

`lib/preferences.ts`:

```ts
export function updatePreferences(patch: Partial<Preferences>): void {
  ...
  cache = { ...(cache ?? {}), ...Object.fromEntries(globalPatch) }
  void savePreferences(cache).catch(() => {})   // the whole object, from a cache loaded once
}
```

`cache` is filled by one `onPreferences()` per page load (`ensureLoaded`) and there is no refresh after that, so a tab's copy goes stale the moment anything else writes. `savePreferences` on the daemon replaces the stored block wholesale, so the stale copy wins.

### Repro (measured, real Chrome against a daemon on an isolated registry)

1. Tab A: open the dashboard, set the theme to Dark.
2. Tab B: open the dashboard (loads theme = dark into its own cache), set the theme to Light. Stored value is now `light`, both tabs look light.
3. Tab A (still open from step 1, cache still says `dark`): toggle anything unrelated, e.g. Desktop notifications in the bell menu.
4. Stored value is `"theme":"dark"` again. Refresh Tab B: dark.

Step 3 changed one notification toggle and reverted a setting it never touched. Same for `agent`, `model`, `editor`, the notification set: whichever tab writes last replays its whole snapshot over everyone else's.

### Fix

Send the patch, not the snapshot: a new `patchPreferences` / `patchProjectPreferences` RPC that merges the changed keys over what is on disk and then sanitizes. Clearing keeps working for free, because the sanitizer already drops blank strings and empty lists (`editor: ''` is how the editor is cleared today).

Plus refresh the preferences cache when a tab regains focus, the way `refreshFileConfigs` already does for the repo tier, so a tab that has been in the background does not keep *showing* a value someone else changed.

### Not the cause, but adjacent (out of scope here)

- A failed save is invisible: `savePreferences` returns `{ ok: false, error }` and every call site does `.catch(() => {})` and ignores the result, so on a host with no preferences store (the relay) every setting silently reverts on refresh with no hint.
- The theme is applied in an effect after the preferences round-trip, so first paint is always the OS theme. Not the reported bug, but the reason a wrong theme reads as "it flipped back".

suleimansh:
Closes #1148.

Reported while testing the dashboard: change the theme, refresh, old theme is back.

The theme was not special. Every settings write sent **the whole cached preferences object**, and a tab loads that cache once and never refreshes it. So a tab open since before someone else's change replayed its own snapshot on its next write, whatever key that write was actually about. The theme is just the one you cannot miss.

### What changed

- `patchPreferences` / `patchProjectPreferences` on the registry, the RPC layer and the store seam: merge the given keys over what is stored, sanitize the merge, return the result. Both store methods are optional, so a host that does not implement them (the relay) answers not-enabled exactly like the save pair.
- `lib/preferences.ts` sends only the keys it changed, and **adopts what comes back** — so writing is also how a stale tab converges. A reply is discarded when a newer write has since gone out, so it can never put back a value the user just moved off.
- Both tiers are re-read on `focus` / `visibilitychange`, next to the existing `refreshFileConfigs`, so a background tab stops *showing* a value someone else changed. A tier with a write in flight is skipped.

No sentinel for clearing: the sanitizer already drops blank strings and empty lists, so `{ editor: '' }` still clears the editor, and `{ customPresets: [] }` still clears the presets. Both are pinned by tests.

### Verified

Real Chrome against a daemon on a throwaway registry, the same repro as the issue:

| | before | after |
|---|---|---|
| tab B sets Light, then stale tab A toggles desktop notifications | stored theme back to `dark`; B refreshes dark | stored theme stays `light`; B refreshes light |
| stale tab A regains focus | keeps showing dark | re-themes to light with no write |

(The second row was driven by dispatching the event: Playwright gives each page its own window, so neither ever goes hidden.)

Single-tab behaviour is unchanged — pick a theme, refresh, it sticks — on the project page, the overview and `/settings`.

framework 1286 tests, dashboard 429, typecheck + build clean. Patch changeset on `@gemstack/the-framework` (the dashboard is bundled into it).

### Left out, worth their own issues

- A failed save is invisible: every call site does `.catch(() => {})` and ignores the `{ ok: false }` result, so on a host with no preferences store every setting silently reverts on refresh with no hint.
- The theme is applied in an effect after the round-trip, so first paint is always the OS theme.


suleimansh:
CI is red on `binding a topic run re-homes it into the bound project` (#1122), which is **not this change** — the same test fails on `main` at 47049ab (run 30136805342) with no preferences code involved. It times out at a 6s poll budget on a loaded runner. Filed as #1153, fix in #1154; rebase this once that lands.

PR #1156, opened 2026-07-25
fix(dashboard): name what "New" makes, and keep the session list titled when empty

### LINKED ISSUE #1147: "New" => "New session"
New what?

Image

I suggest "New session"

suleimansh:
Closes #1147.

- `New` → `New session`, in all three forms of the button (start in this project, pick a project, add one first). The dropdown's `aria-label` already said "New session", so this only brings the visible label in line.
- `Recents` → `Recent sessions`, and the heading now renders even when the list is empty, with `No sessions yet.` under it — your second comment on the issue.

Driven in Chrome at 1280px and 900px: both labels fit the sidebar, no horizontal overflow, and the empty state reads `Recent sessions` / `No sessions yet.`

dashboard 429 tests (the two that asserted the old label updated), typecheck + build clean, patch changeset.

### DIFF
--- a/.changeset/new-session-label.md
+++ b/.changeset/new-session-label.md
@@ -1,0 +1,5 @@
+---
+'@gemstack/the-framework': patch
+---
+
+Name what the sidebar's "New" makes: it reads "New session". The session list below it is titled "Recent sessions" and keeps its heading when empty, so "No sessions yet." reads as the state of that list.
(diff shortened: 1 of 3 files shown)


suleimansh:
CI here failed on `daemon-workspace.test.js` timing out, not on anything in this PR — it is the poll budget I raised too far in #1154 colliding with the suite's 60s per-test timeout. Fix in #1160; this will go green on a rebase once that lands.

Issue #1163, opened 2026-07-25
`TODO_AGENTS.md` doesn't respect `todo_format.md`


brillout:
Maybe because [`TODO_AGENTS.md`](https://github.com/gemstack-land/gemstack/blob/dc7d79c3f4d0afef5416520fb499fdd69cdb9558/TODO_AGENTS.md#L1) was generated before we introduced `todo_format.md`?

Issue #1117, opened 2026-07-24
Record the ticket a run starts from, for a true 'implementing' status


suleimansh:
## Follows #1112

The "Hot tickets" overview added in #1112 has an **In progress** lane. Today it infers a ticket is being worked from its `.plan` / `.spike` sibling. There is **no link between a run and the ticket it is implementing**, so a ticket being coded right now does not show as "implementing".

## What is missing

`RunMeta` carries no ticket reference. To show a true `implementing · live` state, a run needs to record the ticket it started from.

## Wrinkle

The pipeline loses ticket identity along the way: a ticket is promoted into the plain-text queue (`TODO_AGENTS.md`), and the loop works text items. So the link needs one of:

- the queue item to carry the ticket's slug through to the run that picks it up, or
- a dedicated "start a run **for this ticket**" entry point that records the ticket.

## Outcome

With the link in place, the Hot tickets card can mark a ticket that has a live run as **implementing**, instead of approximating it via plan/spike.

## Priority

Optional / post-MVP — the planned-or-spiked proxy is acceptable for now.

Issue #529, opened 2026-07-15
[The Framework] Pause the run when a consumption limit is reached


suleimansh:
Part of #519. The act half: the limits can be set (#527), kept fed (#525) and judged (#523), but nothing acts on them.

Rom's call on #519: pause the current session(s) on all three limits, and leave a `Resume <session name>` entry behind, because the backlog is already what a later run drains, so the resume needs no machinery of its own.

- [ ] Consult the limits between turns, from a cached reading (a live read spawns the whole CLI, ~5s, so it can't sit on a hot path)
- [ ] Pause as a clean stop, like the budget cap, not a failure
- [ ] Leave the resume note on the backlog, named after the session
- [ ] Carry on when the quota can't be read: Rom's "Agreed, keep running" on #519
- [ ] Tests

**Not in scope:** wiring the poller in at the daemon/CLI, the server waking itself at the reset, and the bars. #519 covers those.

Issue #1165, opened 2026-07-25
test: the #1122 re-home test loses the bind under load (not a poll budget)


suleimansh:
`binding a topic run re-homes it into the bound project` (#1122, `daemon-workspace.test.ts:269`) fails intermittently on CI. It passes locally every time and passes on most CI runs, so it is a race, not a broken assertion.

**I diagnosed this wrong in #1153 and the fix in #1154/#1160 does not address it.** I read the first failure (6578ms against a 6s poll budget) as a slow runner and raised the budget. The evidence since says otherwise:

| budget | outcome |
|---|---|
| 6s (#1153) | failed at 6578ms |
| 30s (#1154) | whole file hit the 60s `--test-timeout` |
| 15s (#1160) | failed at 16311ms |

Every failure burns the **entire** budget and then reports `1 !== 2`. If this were slowness, a bigger budget would have caught it; instead the second spawn never arrives at all. The daemon is not re-homing on those runs — the bind is being lost.

The test seeds the bind by writing `run.json` and `events.jsonl` into the topic run's scratch dir after `onStart` returns, and the daemon tails that events file for `{kind:'bind'}` (`rehomeTopicRun` in `daemon-runtime.ts`). Two candidates for the lost event, both ordering-dependent and both invisible locally where the machine is fast:

1. The watcher's tail is armed against a file that does not exist yet and does not pick up the later creation on that path.
2. The daemon creates or cleans the scratch dir around the same moment the test writes into it, so the seeded files are clobbered.

Worth reproducing under load (`taskset`/`stress`, or a CI runner with the box busy) rather than reading the code cold — the timing is the bug.

Until it is fixed the test is a merge blocker on unrelated PRs (#1156 hit it twice). If the fix is not quick, quarantining it with a pointer here beats re-running CI until it passes.

suleimansh:
Two things for whoever picks this up.

**A diagnostic is cheap and worth doing first.** The reason is already on disk when it fails — a failed re-home writes `could not re-home this run: <reason>` into the scratch events log and retains the scratch — so the test should read both when the second spawn is missing, instead of asserting a bare `1 !== 2`:

```ts
if (starts.length < 2) {
  const events = await readFile(join(scratch, FRAMEWORK_DIR, EVENTS_FILE), 'utf8').catch(() => '<no scratch events log>')
  const scratchGone = await stat(scratch).then(() => false, () => true)
[...]

PR #1168, opened 2026-07-25
fix: give a queued ticket a place in the queue, and a way back to its ticket

### LINKED ISSUE #1164: `Queue` button doesn't seem to work?
I clicked on `Queue`:
Image

All it seems to do is this:

Image

So I don't think it works?

suleimansh:
Closes #1164
Part of #1117

Clicking `Queue` did write. It appended `- [ ] Notifications` to the end of `TODO_AGENTS.md`, which is why the diff in the issue looks like the button did almost nothing. Two things were actually wrong.

## The entry went to the back of the queue

The queue is worked front to back: the drain preset takes "the FIRST open entry", and `parseTodoEntries` reads list items in file order with headings ignored entirely. So appending put a ticket you had just deliberately picked *behind everything already in the file*. On our own backlog that is seventh in line, behind items like "Dogfooding: start using The Framework on our own repos".

Entries now go into the `## Priority N` section that `todo_format.md` specifies, created when the file has none, ranked by the ticket's own `priority:` key.

## The entry lost the ticket

It carried nothing but a title, so the ticket it came from was gone the moment it was queued. The entry is now a markdown link to the ticket file: it reads well in the file, survives `parseTodoEntries` verbatim, and gives the agent draining it the ticket to open.

That is the first half of #1117, whose "wrinkle" is exactly this: *the queue item to carry the ticket's slug through to the run that picks it up*. Recording it on `RunMeta` is still open and I have not claimed it here, because the **agent** picks which entry to work and the daemon never parses the queue, so there is nothing daemon-side that knows which ticket a run is on. Hence "Part of", not "Closes".

## The priority mapping

`todoPriorityForTicket` translates the ticket format's words into the backlog format's numbers: `urgent` 9, `high` 7, `medium` 5, `low` 2, unmarked 5. The two formats were specified separately and never mapped onto each other, so this is a new call and worth a second opinion. The scale's ends are deliberately unreachable from a click: `todo_format.md` reserves 10 for "critical production bugs" and 0 for "only if capacity", and neither is something a button press should be able to claim.

## Notes

- `insertTodoEntry` is pure. Where an entry lands is the whole point, and that is far easier to pin than through the filesystem.
- Writes with no priority are untouched and still plain appends: the #529 resume note and the agent's own follow-ups are a running list, and the order is theirs.
- With no priority sections at all, the new section goes *above* the file's own headings. Those sections are unranked (ours are `## MVP v1` / `## MVP v2`), so burying a deliberate pick under them is the bug being fixed. Files converge on the format as they are used, and #1163 is the other half of that.

## Verified

Framework 1295 pass / 0 fail (9 new), dashboard 435 pass (2 new, 1 rewritten), typecheck and build clean.

Live, against a daemon serving a copy of our real `TODO_AGENTS.md`, queueing `627-Notifications.md` exactly as in the issue:

```diff
+## Priority 7
+
+- [ ] [Ship the thing](tickets/2026-07-25_ship-the-thing.md)
+
+## Priority 5
+
+- [ ] [Notifications](tickets/627-Notifications.md)
+
 ## MVP v1
```

`onQueue` then returned them as entries 1 and 2 of 8, where a queued ticket used to come back eighth.


suleimansh:
CI is red on `binding a topic run re-homes it into the bound project (#1122)`, which is the pre-existing flake tracked in #1165, not this change.

Evidence:
- the diagnostic merged as #1166 prints its known signature on this run: `scratch removed: true` and `<no scratch events log>`, i.e. the scratch dir was torn down before the bind was seen
- #1167, opened minutes earlier and touching a completely disjoint set of files, failed on the same single test and then passed on a re-run
- nothing here goes near topic runs, worktrees, or daemon-runtime: this PR touches `todo-loop.ts`, `tickets.ts`, `control.telefunc.ts` and `TicketsPanel.tsx`

Locally: framework 1295 pass / 0 fail, dashboard 435 pass, typecheck and build clean. Re-run three times on CI, same one test each time.

Issue #1112, opened 2026-07-24
Add overview of "hot" tickets to dashboard


brillout:
Show "hot" tickets:
- Tickets being worked on (either being implemented, or being planned/spiked)
- Tickets queued
- Tickets likely to be "next" (e.g. high prio ones)

Labeled this high-prio because it's advertised on the landing page, but we can make it post-MVP (I can label it "Coming soon").

PR #1176, opened 2026-07-25
feat(the-framework): record the ticket a run is implementing (#1117)

### LINKED ISSUE #1117: Record the ticket a run starts from, for a true 'implementing' status

suleimansh:
Closes #1117.

## What was missing

The Overview's **In progress** lane (#1112) worked out that a ticket was being worked from the `.plan`/`.spike` sitting beside it, because nothing linked a run to a ticket. Two things fell out of that: a ticket being coded *right now* showed up wherever its sibling files happened to put it, and a ticket nobody had planned yet sat in **Queued** while an agent was writing it.

#1164 had already closed the other half: queueing a ticket writes the entry as a markdown link back to it, so the ticket's identity survives the trip through the plain-text queue. This picks that link back up.

## Where the link is made

At the one moment the framework actually knows what a run is about to implement: the auto-PM **drain** start. That preset works "the FIRST open entry only", and `parseTodoEntries` reads in file order, so the entry the daemon reads is the entry the run will take. Every other job puts work *on* the queue rather than taking it off, so there is nothing for them to name — which is why `AutoPmJob` gained a `drains` marker rather than the call site matching on a preset name.

From there it takes the path every other durable run fact takes:

```
queue entry  ──#1164 link──►  daemon (drain start)  ──--ticket──►  run
                                                                    │
                                        RunMeta.ticket  ◄──event────┘
                                                │
                              buildHotTickets ──┴──►  "implementing"
```

An event rather than a start argument, because only an event reaches the meta, and the meta is what a dashboard tab opened mid-run reads.

The lane prefers hard evidence and keeps the old inference behind it: a live run's ticket is `in-progress` because a run said so, and everything else buckets exactly as before, so a ticket someone is working by hand still shows up. The card tags it `implementing` in the primary colour rather than muted — `planned` and `spiked` are marks work left behind, and a lane holding all three should not read as though they were the same claim. (That tag also fixes a row that would otherwise show nothing: `in-progress` used to imply planned-or-spiked.)

## Verified end to end, not just in units

Drove the real binary against a temp repo whose queue holds one linked ticket. The event is the first line of the run log:

```
{"kind":"ticket","path":"tickets/2026-07-25_login.md"}
```

and it lands in the meta. Then `buildHotTickets` with its **real** disk readers, against that run:

```
{"file":"2026-07-25_login.md","title":"Add a login page","bucket":"in-progress","runId":"2026-07-25T16-31-00-176Z"}
```

With the same run finished, the same ticket falls back to `{"bucket":"next"}` — that ticket is `priority: high` with no plan or spike, so it is exactly the case the old inference got wrong.

## Safety

A ticket path comes off a file an agent writes and ends up rendered in the dashboard and opened by a reader, so `isTicketPath` gates both ends — the queue read and the `--ticket` flag. Only `tickets/<name>.md` travels; traversal, absolute paths, URLs, dotfiles and nested paths are all refused, pinned by a test.

## Scope

The wrinkle in the issue is real and unfixed by design: this is a best guess about which entry the run will take, since it reads the checkout a moment before the run reads its worktree. Being wrong costs a mislabelled lane, never a wrong run — nothing acts on it but the display. A run started from a hand-written prompt records no ticket, as before.

`pnpm typecheck` clean; `packages/the-framework` 1313 pass / 0 fail, `packages/framework-dashboard` 443 pass / 0 fail.

### DIFF
--- a/.changeset/run-ticket-implementing.md
+++ b/.changeset/run-ticket-implementing.md
@@ -1,0 +1,5 @@
+---
+"@gemstack/the-framework": minor
+---
+
+feat(the-framework): record the ticket a run is implementing (#1117): a draining run is started with the ticket its queue entry links back to (#1164), which it emits once and folds onto `RunMeta.ticket` — so the Overview's hot-tickets card can mark a ticket as `implementing` because a run says so, instead of inferring it from the plan or spike the work left behind.
--- a/packages/framework-dashboard/components/HotTickets.test.tsx
+++ b/packages/framework-dashboard/components/HotTickets.test.tsx
@@ -40,4 +40,24 @@ describe('HotTickets (#1112)', () => {
     fireEvent.click(screen.getByText('b'))
     expect(picked).toBe('beta')
   })
+
+  test('a ticket a run is implementing right now says so, over its plan/spike (#1117)', async () => {
+    onHotTickets.mockResolvedValue([
+      { ...ht('a.md', 'alpha', 'in-progress', { planned: true }), runId: 'run-7' },
+      ht('b.md', 'alpha', 'in-progress', { planned: true }),
+    ])
+    render(<HotTickets onSelectProject={() => {}} />)
+    await waitFor(() => expect(screen.getByText('a')).toBeTruthy())
+    // Live work outranks the mark older work left behind, so the same lane can say which is which.
+    expect(screen.getByText('implementing')).toBeTruthy()
+    expect(screen.getByText('planned')).toBeTruthy()
+  })
+
+  test('a ticket being implemented with no plan or spike still gets a tag (#1117)', async () => {
+    // The gap the run link opens up: in-progress used to imply planned-or-spiked, so a bare
+    // implementing ticket would have shown an unexplained row.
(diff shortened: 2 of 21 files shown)


Deniwn22 (commented):
Approach is solid. Minor nit: this would benefit from a brief comment explaining why we skip validation for internal callers.

PR #1187, opened 2026-07-25
fix(the-framework): make the ticket import ask for the same thing wherever it is pressed

suleimansh:
Found while fixing #1169, and it sits on the demo path.

## One label, two asks

"Import tickets from GitHub" is offered from two places, and they sent different instructions:

| where | what the agent was actually asked |
|---|---|
| onboarding checklist (Overview, `/settings`) | `Import tickets from GitHub` |
| tickets panel in the right rail | `Import this repo's open GitHub issues into tickets/, one file per issue, following the ticket format.` |

The first is the button's own label handed back to the agent. It names no destination and no format, and it is the one on the onboarding path, so it is the copy a new user (and the demo) hits first.

The other one-line presets are terse but they all name the work: `Look at all tickets/**.plan.md and add all quick-wins to TODO_AGENTS.md`, `Open TODO_AGENTS.md and work on the FIRST open entry only`. This one was the odd one out.

## The fix

`prompts/presets/import_tickets.md` now carries the instruction the panel had been using, and `TicketsPanel` reads `presets.importTickets.render()` instead of keeping a second copy. One source, so the two entry points cannot drift again.

Still deliberately short: the ticket format reaches the agent through the context fragment (#1174), so the prompt names the destination and the shape rather than restating the format.

## Checks

- `pnpm typecheck` clean, `pnpm build` clean.
- `packages/the-framework`: 1316 pass, 0 fail, 1 skipped.
- `packages/framework-dashboard`: 454 pass, 0 fail.
- Two assertions had pinned the old text as though the label and the prompt were meant to be the same string (`preset-catalog.test.ts`, `Composer.test.tsx`); both now assert against the preset, and the catalog test says why the two must differ.

### DIFF
--- a/.changeset/handoff-commits-settled-work.md
+++ b/.changeset/handoff-commits-settled-work.md
@@ -1,0 +1,5 @@
+---
+"@gemstack/the-framework": patch
+---
+
+Commit what a session left in its checkout when its work is handed off, so a settled agent that never committed no longer dead-ends at "No commits between main and the-framework/...". The finishing step also stops calling an empty branch a dead end while the work is sitting uncommitted next to it.
--- a/.changeset/import-preset-says-what-to-import.md
+++ b/.changeset/import-preset-says-what-to-import.md
@@ -1,0 +1,5 @@
+---
+"@gemstack/the-framework": patch
+---
+
+Make "Import tickets from GitHub" ask for the same thing wherever it is pressed. The preset's prompt was the button's own label repeated back, naming neither the destination nor the format, while the tickets panel spelled out the real instruction in a constant of its own. The instruction now lives on the preset and the panel reads it from there.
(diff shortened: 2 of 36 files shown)


suleimansh:
The `drift` check is red for a reason that predates this PR and is not about it.

`63095ae` ("minor refactor: rename ANLYSIS_RESULT.md => ANALYSIS_RESULT.md") corrected the spelling in `prompts/system_prompt.md` but not in the ```md block of #326, and `check-prompt-drift.mjs` holds those two byte-for-byte:

```
prompts/system_prompt.md has drifted from #326 (block 1).
  first difference at line 8:
  issue:  "ADD_ANALYSIS_ENTRY: Add entry to the ANLYSIS_RESULT.md list"
  repo:   "ADD_ANALYSIS_ENTRY: Add entry to the ANALYSIS_RESULT.md list"
```

This PR touches `prompts/presets/import_tickets.md`, which is what made the workflow run at all: it is not part of CI, it runs daily and on PRs that touch `prompts/**`. So the daily cron is failing on main too, and every prompt change is blocked behind it until #326's block is corrected. The guard's rule is that the issue is the source of truth, so the fix is the one character in the issue, not a revert here.

Issue #405, opened 2026-07-11
The dashboard: rebuild on Vike + React + shadcn + Telefunc-stream


suleimansh:
The localhost dashboard is one hand-written HTML/CSS/vanilla-JS string (page.ts, ~1k lines). It has outgrown that: manual XSS escaping on every value, no components, hand-rolled state. Decided to rebuild on Vike + React + shadcn + Telefunc(-stream).

Plan (gated):
- [x] Spike: new packages/framework-dashboard (Vike SPA + React + shadcn + Telefunc-stream). Thin slice = live event stream + Projects sidebar, reading the same files the daemon writes. Side-by-side; page.ts untouched. Gate: does it feel better?
- [x] Transport + serve: Telefunc replaces /api/*; Telefunc-stream (or SSE) for events; daemon serves the built bundle.
- [x] Full port + fold in per-project live streaming (#393); retire page.ts.

Decisions: keep self-contained / relay-shareable (static bundle) vs SSR. WebSockets: no - one-way stream is SSE / Telefunc-stream, commands are RPC.

suleimansh:
Spike merged (#413) - it renders and builds. SSE for the event stream, Telefunc for the Projects RPC (telefunc-stream has no clean public API in 0.2.22 yet, so SSE for events like the daemon does today).

Phase 2, two halves:
1. Port the rest of `/api/*` (runs, docs, logs) to Telefunc RPCs + their panels. Self-contained, doing this now.
2. Daemon serves the built bundle. One architecture question here - how should the daemon relate to the Vike server?
   - (a) the daemon mounts the Vike-built handler + Telefunc + SSE inside its own node http server; run orchestration stays where it is.
   - (b) the dashboard IS the server (Vike + Telefunc), and run spawn/steer move into Telefunc functions; the daemon folds into it.

My lean is (a) now (smaller change), (b) later if we want a single server. Preference? Retiring page.ts comes after parity.

suleimansh:
@brillout

nitedani:
Genuinely curious what you hit here — when you say Telefunc-stream "has the primitive (createPushReadableStream) but not a clean public API yet," what were you reaching for that wasn't there?

[...]

PR #1196, opened 2026-07-25
fix(dashboard): catch render errors instead of blanking the whole UI (#1194)

### LINKED ISSUE #1194: Bug: white blank
Bug: the UI sometimes goes blank (just a white screen), seems to be random.

brillout:
Fixes #1194.

## The bug

The dashboard sometimes went to a blank white screen, seemingly at random.

## Diagnosis

The dashboard had **no error boundary anywhere**. In React, an uncaught error thrown during render propagates to the root and unmounts the entire tree — a blank page. With `ssr:false` (per #405) there isn't even an SSR error page to fall back to, so the whole app just vanishes with nothing on screen and nothing to click.

It looked random because it was data-driven. The dashboard is a projection of a live event stream and a fistful of polls, and it renders that data across many panels (the feed, markdown views, the rails, the overview cards). A single value the code didn't expect — a stray line off `events.jsonl`, a field the daemon hasn't written yet, an unusual shape from a poll — throws during render on the one render that hits it, and takes everything with it. The data hooks are already defensive (`usePolled` keeps the last value on a rejected read, `format-date` handles unparseable timestamps), but there was no net for a render-time throw.

## The fix

Add a small `ErrorBoundary` and place it inside `LayoutDefault`, wrapping the page tree but sitting within the themed `bg-background text-foreground` shell. Now a caught render error shows a **recoverable card** — *Try again* / *Reload* — on the app's own background instead of the browser's blank white, and logs the error + component stack to the console so the next occurrence leaves a trace to debug (the daemon never sees the browser console; a blank page carried none).

- **Try again** clears the boundary and re-mounts the subtree — a transient cause (a stream hiccup, a half-written read the next poll completes) renders straight through.
- **Reload** is the sure way out for a durable cause.

This is the net under all of it: a render crash now costs you a view and a click, not the session.

## Changes

- `components/ErrorBoundary.tsx` — new class-based error boundary with a themed, recoverable fallback.
- `layouts/LayoutDefault.tsx` — wrap the page tree in `<ErrorBoundary>` inside the themed shell.
- `components/ErrorBoundary.test.tsx` — renders children normally; shows the alert (not a blank) on a throw; surfaces the message; logs the error; recovers on *Try again* once the cause is gone.
- Changeset (`@gemstack/the-framework: patch`).

## Verification

- `pnpm test` — 65 files, **479 tests pass** (4 new).
- `pnpm typecheck` — clean.
- `pnpm build` (turbo build + `bundle:dashboard`) — green.

## Note / possible follow-up

This is an app-wide safety net: a crash anywhere in the page replaces the whole page with the fallback. A natural follow-up would be a second, finer-grained boundary around just the main pane so the sidebar and rails survive a main-pane crash and navigation stays usable — deliberately left out here to keep the fix focused on the reported symptom (never a blank screen).

---
_Generated by [Claude Code](https://claude.ai/code/session_01W5PHAJ2sXPqwdbPZJsvYyx)_

### DIFF
--- a/.changeset/dashboard-error-boundary.md
+++ b/.changeset/dashboard-error-boundary.md
@@ -1,0 +1,5 @@
+---
+'@gemstack/the-framework': patch
+---
+
+The dashboard no longer goes to a blank white screen when a view hits a render error. It had no error boundary, so any uncaught error while drawing a view — a stray line off the live feed, a panel reading a field the daemon had not written yet — unmounted the whole React tree to the root and left nothing on screen and nothing to click. It read as random because it was data-driven: a particular event, a particular poll, a particular moment. A caught render error now shows a recoverable card on the themed shell — Try again, or Reload — and logs the stack to the console so the next occurrence leaves a trace, rather than taking the session with it.
--- a/packages/framework-dashboard/components/ErrorBoundary.test.tsx
+++ b/packages/framework-dashboard/components/ErrorBoundary.test.tsx
@@ -1,0 +1,78 @@
+import { afterEach, beforeEach, describe, expect, test, vi } from 'vitest'
+import { cleanup, fireEvent, render, screen } from '@testing-library/react'
+import { ErrorBoundary } from './ErrorBoundary.js'
+
+afterEach(cleanup)
+
+// React logs a caught render error to console.error (twice, plus the boundary's own log). That is
+// expected here and only clutters the run, so swallow it for these cases.
+beforeEach(() => {
+  vi.spyOn(console, 'error').mockImplementation(() => {})
+})
+afterEach(() => {
+  vi.restoreAllMocks()
+})
+
+/** Throws on demand, so a test can flip it from crashing to healthy between renders. */
+function Boom({ crash }: { crash: boolean }) {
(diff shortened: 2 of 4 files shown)


brillout:
@suleimansh Shall we merge while leaving https://github.com/gemstack-land/the-framework/issues/1194 open?

brillout:
Yes — I'd merge this and keep #1194 open.

This PR is a **mitigation, not a root-cause fix**. The error boundary stops the blank white screen (you get a recoverable card instead) and — the part that matters for #1194 — logs the failing error + component stack to the browser console. The specific render error that triggers the crash is still unidentified; until now it left no trace, so there was nothing to chase.

So the split is clean:
- **Merge now** for the immediate win: no more blank screen, and every future occurrence is finally traceable.
- **Keep #1194 open** to track the actual culprit, now with a console stack to point at it. Once we've caught and fixed the underlying throw, that's when #1194 closes.

CI is running on the latest commit (green locally: 479 tests, typecheck, and the full `build` + `bundle:dashboard`). Happy to merge on your go, or to keep digging for the root cause under #1194 — just say which.

---

Issue #1149, opened 2026-07-25
Improve tooltip


brillout:
This tooltip is nice:

<img width="238" height="121" alt="Image" src="https://github.com/user-attachments/assets/b56576e4-c008-49e6-8c37-bc882f0f99fb" />

But:
1. Show it immediately (no delay)
2. Always use it, sometimes the system tooltip is used (which always comes with a bad slow delay):

<img width="263" height="114" alt="Image" src="https://github.com/user-attachments/assets/89e0daff-6093-4fa8-831f-1d8d4b9ccb65" />

suleimansh:
I agree about tooltips, 
one note:

<img width="466" height="314" alt="Image" src="https://github.com/user-attachments/assets/73405f49-3e02-4695-9356-dfb4105793ea" />

I have grouped them in one menu dropdown [...] (check session page to see it), 
What do you think? keep them separated or one dropdown? (I prefer one dropdown for this actions - less element shifting when you navigate between sessions, as action buttons count can change by active session) 

Note: the inner dropdown content will be improved more..

brillout:
> I have grouped them in one menu dropdown

Love it, awesome idea.

brillout:
How about we automatically show the dropdown when any action is hovered (so we can remove the `...` button)? Only if quick-win.

brillout:
Same for this:

<img width="197" height="93" alt="Image" src="https://github.com/user-attachments/assets/f791b4eb-60fc-48a6-8438-c4291da432c9" />

How about We remove `Settings: `, show the gear on the right side of the `Autopilot`, `Open PR` => auto-expend the dropdown on hovering any settings.

Also `Autopilot`, `Open PR` look too much like buttons, maybe make them look like labels.

brillout:
Closing in favor of:
- https://github.com/gemstack-land/the-framework/issues/1368

PR #1201, opened 2026-07-25
[AI] Tooltips: no delay, and the custom one everywhere

brillout:
The shared TooltipTrigger now defaults to delay={0} — Base UI reads the delay off the trigger and an explicit one beats any provider, so a single default governs every tooltip and the two delay={300} providers go.

The ~40 native `title=` tooltips left on buttons, menu items, chips and indicators are the custom tooltip now, including the "Open in editor" button the issue's screenshot was taken on. Titles that only repeated already-visible text were dropped instead; `title` as an overflow fallback on truncated text, and the `title` props on Section/Dialog, stay as they were.

Tests that read a hint off a `title` attribute hover instead, via a small hoverTooltip() helper (packages/framework-dashboard/test-utils.ts).

Ticket: tickets/2026-07-25_improve-tooltip.md (#1149) — the OP's two items; the thread's design follow-ups keep it open.


brillout:
@suleimansh Same thing, didn't read any of the code changes

PR #1202, opened 2026-07-25
The framework/improve tooltip

brillout:


brillout:
https://github.com/gemstack-land/the-framework/pull/1201

Issue #1050, opened 2026-07-22
Run target: wire GitHub Actions into the "Run on" gear selector (driver axis)


suleimansh:
Part of #1049. The **driver axis** of the "Run on" selector. Zero security surface, so it ships first.

Adds "Current device" (default) and "GitHub Actions" as run targets in the Composer's options gear, wiring the already-merged `ActionsDriver` (#934) into the run path. "Current device" is today's behavior; "GitHub Actions" runs the turn on a fresh Actions runner.

## Why this is self-contained

`ActionsDriver` is already on `main` (`packages/framework/src/driver/actions.ts`, exported from `driver/index.ts`), but nothing constructs it: `createDriver` (`agent.ts`) only knows `claude`/`codex`. This issue is the wiring, no new driver code.

## Scope

- `StartRunOptions.target?: 'local' | 'actions'` in `packages/framework/src/dashboard/types.ts` (browser-safe leaf).
- `Preferences.target` in `registry.ts` with a **dedicated sanitizer branch** in `sanitizePreferences` (the `PREFERENCE_KEYS` loop is boolean-only and silently drops strings; mirror `theme`/`agent`). Make it a `PROJECT_PREFERENCE_KEY` so "run this repo on Actions" sticks per project like `agent`/`model`.
- `--run-on` flag: emit `--run-on actions` from `startOptionFlags` in `daemon-runtime.ts` (mirror the `--agent` branch; `local` emits nothing = byte-identical default); parse it in `cli.ts`.
- Driver construction: a `createRunDriver({ agent, target, actionsConfig })` wrapper at the `cli.ts` driver-construction site that returns an `ActionsDriver` for `target==='actions'`, else the normal agent driver. Don't bloat `createDriver` (ActionsDriver's owner/repo/token don't fit `CreateDriverOptions`).
- Actions config: owner/repo from the project's `origin` git remote; the user PAT (`repo`+`workflow`, must be a user token not a GitHub App) from env (`GH_TOKEN`) or `the-framework.yml`.
- UI: a new `DropdownMenuSub` ("Run on →") at the top of `OptionsMenu`, single-select with a `Check` mark (model on `AgentModelMenu`, not `OptionRow` which is boolean-only). Rows: Current device, GitHub Actions, Claude web (disabled placeholder). Hide/inert in-session (baked at spawn, same as the agent select).

## Acceptance

- Picking "GitHub Actions" in the gear starts the run via `ActionsDriver`; "Current device" is unchanged.
[...]

Issue #495, opened 2026-07-14
Epic: Bring your own subscription (BYOS) — run the user's own coding-agent CLI


suleimansh:
Bring your own subscription (BYOS): run the AI on the CLI the user already pays for, instead of reselling tokens or asking for API keys.

Idea from Traycer (open source). They don't touch keys. They run the user's already-installed coding-agent CLI (claude-code, codex, gemini), which already holds that subscription's auth. Requests go straight through it.

Fits our existing seams with almost no new concepts. Two plug points already exist:
- Runner (where code executes) — the local runner spawns the CLI.
- ai-sdk ProviderAdapter (which model does the work) — a new `cli/*` adapter.

BYOS = a new adapter that spawns the user's CLI through the local runner, plus a small discovery step. Other providers are just more entries: ChatGPT via Codex CLI, Gemini via Gemini CLI, Claude via Claude Code. A plain web subscription with no CLI is not reachable this way.

Plan:
- [x] Spike `cli/claude-code` adapter + PATH discovery (child)
- [x] Widen discovery to codex + gemini CLI
- [x] Provider config: where the pick + custom paths persist (framework config, TBD with Rom)

Open calls for Rom:
- go/no-go on the spike
- config location: framework config (env-paths/XDG) vs a separate dotfile

suleimansh:
Read Traycer's repo. What's actually borrowable is less than it looks, worth knowing before we commit.

Open: the wire protocol, the provider-id schemas, the usage panel rendering, the polling policy. Closed: the host binary that does the actual reads and runs the CLIs. Their provider resolution has the shape we want (bundled / PATH / custom, persisted to `provider-overrides.json`) but no implementation, and there are no headless invocation args for claude, codex, or gemini anywhere in the repo.

So we get the schema and the credential model, not the brain.

The credential model is the part worth copying exactly: "ambient" is the user's own `~/.claude`, read-only, never written to. "managed" is an isolated config dir Traycer owns. They never parse a token, they point the CLI at a config dir and let it authenticate itself. We should hold that same line.

#496 scope is unchanged then, we still have to find the invocation ourselves. The generation half I've already proven locally: `claude -p`, no API key set, clean JSON back with the result, a usage block, and a session id.

Still need two things from you before I build:

- go/no-go on the spike
- config location: framework config (env-paths/XDG, like the dashboard prefs) or a separate dotfile

suleimansh:
[...]

PR #1207, opened 2026-07-25
feat(the-framework): run a session on Claude Code on the web

### LINKED ISSUE #610: Drive Claude Code on the web instead of the CLI (MVP shortcut: free worktrees + PR, 0% local CPU)

suleimansh:
Makes "Run on: Claude web" a real target instead of the disabled placeholder it has been since #1050. Picking it hands the session to Claude Code on the web on your own account: the work runs on Anthropic's infrastructure at no local CPU cost, does its own git worktree, and opens its own pull request. That is #610's original pitch, and it now works.

Closes #610.

## What changed the verdict

The last look at this concluded `--cloud` was interactive-only and therefore not drivable behind the `Driver` seam. That was wrong in one specific way: the flag is gated on **stdout being a terminal**, not on a human being present. The refusal exists so a piped invocation does not silently run locally instead. Run the CLI under a pty and the gate is satisfied, and it then prints exactly what a driver needs:

```
Created cloud session: Add the --verbose flag
View: https://claude.ai/code/session_01ABC...?from=cli&m=0
Resume with: claude --teleport session_01ABC...
```

So there is no browser automation, no Chrome extension and no UI scraping here — the two mechanisms the issue floated, both of which the Usage Policy rules out. It is the CLI's own flag, with the same subscription auth every other target already uses (#495).

## What this target is, and is not

It is a **hand-off**, not a streamed peer like local / Actions / a connected device, and that is a property of the surface rather than a shortcut taken here: a cloud session exposes **no read-back API at all** — no status, no transcript, no output endpoint, only the session URL. So the turn resolves once the session exists, and the run view shows where the work went:

- a link straight to the session on claude.ai
- the `claude --teleport <id>` command that pulls it back to this machine, with a copy button

For the same reason there is no `readCode`: the workspace lives in a cloud VM this machine never sees. The PR the session opens is the read-back.

## Shape

- `driver/cloud.ts` — `CloudDriver` / `CloudSession`. Fits `Driver`/`DriverSession` as written, no new methods.
- Wired through the same sites as the Actions target (#1050): `--run-on web`, `Preferences.target`, `StartRunOptions.target`, `RunMeta.target`, the daemon's flag, the gear's "Run on" list and the settings row.
- `CloudRunNotice` mirrors `ActionsRunNotice`; the session link rides the driver `action` event (`cloud <url>`) the same way the Actions run link rides `run <url>`, so a tab opened mid-run replays it.

Two details worth flagging for review:

- **The prompt never touches a shell.** `script`'s two dialects differ (BSD takes argv, util-linux takes `-c`), so the hosted command is a **fixed literal** and the prompt and model arrive as environment variables. The model is additionally restricted to `[A-Za-z0-9._:-]`.
- **An untrusted workspace is reported, not answered.** Trusting a directory is a decision to make once in your own terminal, so the run says which directory needs it rather than pressing Enter on the dialog for you.

## Verified live, not just under test

Driven end to end against a real account, through the built CLI, twice:

```
◆ run on: Claude Code on the web (a cloud session on your own account)
    · cloud https://claude.ai/code/session_015JkksZJJzABLHL9bS14rzt
  ‹ turn complete
✓ finished
```

The first live run is what found the one bug the tests could not: BSD `script` reads its own stdin's terminal attributes to mirror onto the new pty, and under the daemon stdin is a socketpair, so it died with `tcgetattr/ioctl: Operation not supported on socket` before running anything. stdin is now a regular file, which `tcgetattr` fails on harmlessly. The tests inject the pty runner, so only a real run could have surfaced it — second commit.

Tests: framework 1364 pass / 0 fail / 1 skipped, dashboard 504 pass. 14 driver tests plus the run-target and menu cases.

## Known limits, stated rather than hidden

- **One cloud session per prompt.** The CLI can start a session and pull one back, but it cannot send a second message to one, so a follow-up message starts a fresh session rather than pretending to continue the first.
- **The session clones from GitHub, not from disk.** Local commits that were never pushed are not in it. A repo with no GitHub remote falls back to the CLI's bundle upload, and the run says so.
- **No live feed, no diff panel, no browser pane** for a web run, for the read-back reason above.


suleimansh:
@brillout finally we have this working 🎉

brillout:
:star_struck:

PR #1242, opened 2026-07-26
Only the maintenance sweep describes itself in the routine list

brillout:
## Summary
- The "Routine work" card on the Overview subtitled every routine with its `describe` line, which mostly repeated the label in different words. Now only the maintenance sweep keeps its line — "sweeping the codebase for maintenance work" — since its label alone does not say what it does.
- `AutoPmJob.describe` is optional and removed from the other four jobs entirely (drain, triage quick, triage consensual, spike & plan). The card renders it when present; the daemon's log line and outcome message fall back to the label (`describe ?? label ?? name`), which for those routines already says the same thing.
- The decision is data on the job rather than a name match in the card, per the file's own `drains` rationale: a renamed routine cannot quietly gain or lose its line.
- Includes a changeset (patch).

## Test plan
- [x] New model test in `auto-pm.test.ts`: only the maintenance sweep carries a `describe`, pinned to the exact wording.
- [x] New card test in `RoutineWork.test.tsx`: a describing routine gets a second line, every other row renders exactly one. Full file: 19 passed.
- [x] `the-framework` suite: 1397 pass; failures are pre-existing on the base (triage-prompt assertions, see comment below), sandbox-only daemon-spawn tests, and one known `daemon-workspace` ENOTEMPTY flake — all identical with this diff removed.
- [x] `tsc --noEmit` clean on `framework-dashboard`; `turbo build` clean on `the-framework`.

### DIFF
--- a/.changeset/maintenance-detail-only.md
+++ b/.changeset/maintenance-detail-only.md
@@ -1,0 +1,5 @@
+---
+"@gemstack/the-framework": patch
+---
+
+Only the maintenance sweep describes itself now, in the Overview's Routine work card and the daemon's log line alike, since "Maintenance" alone does not say what it does. The other routines' labels already read as what they do, so their rows lose the line that repeated the label in different words and their log lines say the label itself (`AutoPmJob.describe` is optional, carried only by the maintenance job).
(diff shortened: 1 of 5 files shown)


brillout:
CI is red on `build`, but it's failing on the base branch too: main's own CI is red on `3bcffa0` (this PR's exact base) with the same two failures — `preset-catalog.test.ts`'s triage-pair assertions still pin the old prompt wording ("Only pick tickets that are quick-wins and consensual", "abort and do nothing"), while the `wip` commit reworded `prompts/presets/triage_quick.md` / `triage_consensual.md`. This PR's own packages pass (`framework-dashboard` tests green, `the-framework` failures identical with this diff removed). Since those prompts look mid-edit, I'm leaving them alone — I'll re-merge and re-run once main recovers.

---

Issue #1225, opened 2026-07-26
Choices no working with CC web


brillout:
My prompt I fired using the dashboard (run via CC web):

```
https://github.com/gemstack-land/the-framework/issues/1144
```

I get this in the dashboard before even the agent replies on CC web:

<img width="1366" height="729" alt="Image" src="https://github.com/user-attachments/assets/3793a1c7-3506-4365-a058-0e97e43327ce" />

Later, I get this reply on CC web: 

<img width="1366" height="725" alt="Image" src="https://github.com/user-attachments/assets/3279b4c8-0886-4c66-95f5-61a19ed40d1b" />

But I still get the same view shown above in the dashboard, also after refreshing the page.

brillout:
Tried CC web 3 times, got this exact problem 3 times (with surprisingly the same choice)

suleimansh:
Re-test evidence (2026-08-17), current stack (post #1518/#1536/#1544/#1548):

**The original symptom is gone.** A web run no longer strands a choice in the dashboard. Plainly: the cloud session now answers its own gates instead of parking on them. I launched a run whose prompt explicitly demanded a choice gate ("ask me alpha or beta and wait") — the session refused to park, per the detached-session rule in its instructions: it stated *"This session runs detached, so the choice gate can't actually be answered by anyone — per the detached-session rule I take the option I would have marked as recommended"*, picked the recommended option, said which assumption it made, and finished clean (no repo change, so correctly no PR). Run `2026-08-17T12-31-39-194Z`, session `session_01EsFVgZXdkUrz77fanBe4n7`.

**What still fails on this machine: the bridge/mirror half.** The run view's new "Cloud session mirror" never got past "Connecting to the cloud session…". On the claude.ai page, the bridge panel explains why — it reports `bridge: Failed to fetch`, `transcript: Failed to fetch`, `answer: none delivered`. And the panel titles itself **v0.7.1**: the copy Chrome has loaded is stale (repo ships 0.8.0 since #1548; the version comes from `chrome.runtime.getManifest()`, so this is genuinely the loaded manifest, not a stale string). The stale copy even false-positive-detects the system prompt's own rendered template as "the question" (title `<what the human needs to do>`, options `Handled it | Could not handle it`) — exactly the decoy problem the current content.SPEC.md fixes ("the page's rendered copy of our own protocol template is a decoy").

[...]

PR #1243, opened 2026-07-26
feat(the-framework): let the routine keep several concurrent agents (#1204)

### LINKED ISSUE #1204: Setting to set number of concurrent agents
New setting for routine: number of concurrent agents (default 2).

The setting drives the number of concurrent agents that are spinned by the routine.

Needed for the demo video, so we can show off 10 CC web sessions working in parallel.

brillout:
Closes #1204.

A new **Concurrent agents** setting (`autoPmConcurrency`, default **2**, up to **10**) drives how many runs the routine sweep keeps live per project. A sweep now tops a project up **in one tick** instead of one run per interval — so with the setting at 10 and a standing queue, one click of "Trigger routine now" fans out ten sessions in parallel, which is the demo scenario the issue asks for.

## What changed

**Setting** — `Preferences.autoPmConcurrency`, sanitized like the spend slider (whole number, clamped 1..`MAX_AUTO_PM_CONCURRENCY`), with the default and bound in `preference-defaults.ts` so the browser control and the daemon sanitizer read the same numbers. The control sits on the Routine work card beside the Auto-run switch, and the card's fine print restates the cap ("Keeps up to N agents going at once…").

**Sweep** — `autoPmDecision` refuses only at `activeRuns >= concurrency` (the refusal names the cap, per #960's lesson), and the sweep builds a per-tick batch up to the project's headroom:

- **Drain mode**: one agent per open queue entry, each **pinned to its own entry** by naming it in the prompt. Unpinned concurrent drains would all fork the same checkout and implement the same first entry. An entry assigned to a run still in flight is never handed out again, and with everything taken the report says so.
- **PM mode**: distinct rotation jobs fill the headroom (each at most once per tick — a doubled one would only no-op on its own branch guard), and a due maintenance sweep joins the batch without costing the rotation its turn. The rotation advances only past starts that took; the batch stops at the first refusal so a refused job is retried, not skipped.

**Queue promotion** — landing a finished run's `TODO_AGENTS.md` used to copy the whole file, so with two agents the later promotion reverted the earlier one: it un-checked the entry a concurrent run had just retired and the sweep started that work over again. `promoteQueue` now reads the queue at the run's fork point (`git merge-base`) and, when the checkout has moved since, **merges the run's own changes on** — check-offs, removals, and additions at entry granularity, with added entries keeping their priority section. An unmoved checkout keeps the exact wholesale copy it always had. The queue's line grammar moved to a new leaf `queue-merge.ts` (with `insertTodoEntry`), shared by the backlog loop and the merge so the two can never read the same line differently.

**Wiring** — the daemon passes the setting and the queue's open entries to the sweep, and names a pinned drain's ticket from its own entry (`#1117`'s lane keeps working when several drains run at once).

## Notes

- Existing behavior at concurrency 1 is preserved exactly (the old tests run under an explicit `concurrency: 1` harness); the new default of 2 is what the issue specifies.
- For `target: web` runs the local process ends at the hand-off (#1225), so the daemon cannot count cloud sessions as live and the per-project cooldown is what paces re-starts; the promotion loop also never closes for them locally. That reconciliation gap predates this PR and is orthogonal — within a tick, pinning still guarantees the fanned-out sessions work disjoint entries.

## Tests

- `registry`: round-trip + clamping of the new preference.
- `auto-pm` (+13): decision cap/default/floor, fleet-in-one-tick with pinned prompts, no double assignment, live-run counting, rotation fan-out without doubling, refusal-stops-the-batch, maintenance sharing a tick, unreadable-setting fallback.
- `queue-merge` (new): concurrent check-offs compose, both sides' queued entries survive, priority placement, removals/rewords, bare items, prose untouched.
- `queue-promote` (+4): merge path, exact-copy path, no-merge-base fallback, nothing-to-land skip.
- `RoutineWork` (+3): default shown, writes + clamps the preference, fine print follows the value.

`turbo run typecheck` passes for all 22 packages; `the-framework` 1420/1431 tests pass (the 10 failing daemon integration tests fail identically on `main` in this sandbox — spawn-restricted environment); `framework-dashboard` 558/558.

https://claude.ai/code/session_01AoDsAvEMKzShkiRY6Wba9z

---
_Generated by [Claude Code](https://claude.ai/code/session_01AoDsAvEMKzShkiRY6Wba9z)_


brillout:
Closing as superseded: #1204 landed on `main` via #1252 (91a23ff), which covers the same surface — the `autoPmConcurrency` setting with the same bounds and control, per-entry-pinned concurrent drains, and order-independent promotions — with a deliberately narrower design (only draining fans out, and a pinned drain's promotion lands just its own entry instead of a general document merge).

The one piece here that #1252 chose not to need is the general three-way queue-file merge (`queue-merge.ts`), which composes *any* concurrent promotions — including rotation/hand-started runs that rewrite the whole file. If that ever becomes wanted, this branch (`claude/github-issue-1204-vyjxvx`) has it with tests.

---

PR #1309, opened 2026-07-27
Open TODO_AGENTS.md and work on this one open entry only, then check it 

brillout:
Open TODO_AGENTS.md and work on this one open entry only, then check it off. Do not start any other entry. The entry:

[Settle-moment UX: always offer `Open PR`, auto-commit, no-diff guard](tickets/2026-07-25_unclear-ux-what-now.md) — implement the in-thread maintainer decisions: default `[x] Open PR`; unattended runs commit their work automatically (safe: every run is on its own branch); when the branch has no diff, don't offer `Open PR` — say so and name the uncommitted work; launcher gear reduced to one `Open PR` row (the push setting stays available in `the-framework.yml` and the CLI flag). The `Auto commit`/`Auto push` settings split is explicitly postponed — do not add settings.

Opened from The Framework session `2026-07-27T16-52-27-859Z`.

### DIFF
--- a/.changeset/settle-moment-no-diff-guard.md
+++ b/.changeset/settle-moment-no-diff-guard.md
@@ -1,0 +1,8 @@
+---
+'@gemstack/the-framework': minor
+'@gemstack/framework-dashboard': patch
+---
+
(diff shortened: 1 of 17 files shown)


brillout:
@suleimansh Feel free to close or merge. (I didn't read the code.)

Issue #1328, opened 2026-07-27
Web runs: drive claude.ai sessions through the browser extension instead of claude --cloud


suleimansh:
Maintainer direction (chat, 2026-07-28): invest in the extension, not `--cloud`. Rationale: `--cloud` is a CLI flag that could be removed at any time, while Claude web itself is a stable product surface. And `--cloud` is currently broken outright on at least one account: every session it creates is a bundle upload that can never push (#1320, filed upstream as anthropics/claude-code#81776).

The decisive technical fact from the #1320 investigation: sessions created through the claude.ai UI **with the repo picker** are repo-bound and CAN push and open PRs - that is the working path. The extension already operates on claude.ai pages (v0.7.1: mirrors transcripts, delivers bridge answers by filling the composer and clicking send, #1237). Extending it to CREATE sessions rides the working path end to end.

## Sketch

- Daemon queues a session request on the bridge (sibling of the answer queue): repo, branch, prompt.
- Extension opens a pinned claude.ai/code tab, drives the new-session flow (repo + branch chips, paste prompt, submit), reports the session URL back through the bridge.
- CloudDriver gains an extension-backed mode: hand-off resolves when the bridge reports the session, exactly like the `--cloud` path resolves on the printed URL. Mirror/answers machinery already exists.
- `--cloud` stays as the no-extension fallback (it may also simply work on machines the upstream bug does not hit - maintainer's own test still pending).

## Open questions

1. **The #610 decision reversed:** driver/cloud.ts's header explicitly ruled out driving the claude.ai UI (browser/extension/scraping) on Usage Policy grounds when `--cloud` was chosen. Adopting this direction is a conscious reversal and the policy angle should be re-read deliberately: the extension is user-installed, acts inside the user's own logged-in browser, and automates exactly the clicks the user would make - but the decision should be on the record, not incidental.
2. **UI-selector fragility:** the v0.7 lessons apply (composer render races, tab revival, claude.ai DOM changes). The bridge's report-status plumbing gives observability the `--cloud` path never had.
3. **Concurrency:** 10 parallel sessions = 10 tabs the extension owns. Feasibility and resource behavior need a spike before #1327's 10-agent goal rides on this.

Refs: #1320 (why --cloud is a dead end today), #1327 (the 10-agent goal this serves), #1237/#1225 (existing extension bridge).

suleimansh:
[...]

PR #1330, opened 2026-07-27
feat(bridge): session start-queue for extension-created sessions (#1328)

suleimansh:
Refs #1328. **Incomplete on purpose - see "What is missing" before reviewing.**

## What is here

**Daemon half (complete, tested).** A start-queue the daemon fills and the extension drains.

- `bridge-starts.ts`: validation (`owner/name`, a git branch name, a bounded prompt), claim-on-read, and a claim that ages out after 90s so a browser quitting mid-creation retries instead of bricking the request.
- `GET /_bridge/start` claims the next request as it serves it. A GET that mutates, which is unusual and deliberate: handing the same request to two polling tabs would create two cloud sessions on the user's account.
- `POST /_bridge/started` takes back an id, a boolean, an optional session id and a note. Success without a session id is recorded as a *failure*, because a run pointing nowhere is not a usable outcome.

The bridge's "deliberately tiny input" rule is intact. A start request carries the free text that rule excludes, and it stays excluded because the direction is outward: the daemon produces it, the extension reads it, and what comes back is only an id, a boolean and a session id. Nothing here widens what someone holding the bridge token can say.

**Page half (`content.js`).** `createSession` drives the repo picker, the branch picker, the composer and send, then reads the session id from the URL - the one thing claude.ai is guaranteed to tell us and exactly what the daemon joins runs on. `probeNewSession` describes the page's controls without touching them.

Every selector in there is a guess about someone else's UI, so the code is built to report rather than insist: each failure names the control it could not find, and the probe exists so the first live run diagnoses instead of just failing.

## What is missing

**The service-worker loop in `background.js`** - poll `/_bridge/start`, open the new-session tab, hand the request to the content script, report back. My tooling refused to write it (a permission classifier blocked the edit), so it is deliberately absent rather than forgotten. Without it nothing drains the queue, so this PR changes no runtime behaviour: the routes exist, the queue exists, and nobody fills or empties it.

Also not started, both named in #1328: `CloudDriver`'s extension-backed mode (hand-off resolving on the bridge report), and the concurrency question. On the concurrency point the honest answer from the page half is that creation is inherently serial - the flow navigates one new-session page, so ten sessions are ten sequential creations, a throughput question rather than a correctness one.

## The policy question is still open

#1328 records that this direction needs the Usage Policy angle settled deliberately. My reading is that Consumer Terms section 3 does not permit it, and that the "user-installed extension in your own browser" arguments are not carve-outs; the two carve-outs are an API key or explicit permission. That is written up in full for the issue. This PR is the code that exists, not a verdict that the direction is settled.

## Checks

Framework suite 1535/0, typecheck clean. Extension harness (`node check.mjs`) 13/13, unchanged by the content.js additions.

### DIFF
--- a/.changeset/bridge-session-start-queue.md
+++ b/.changeset/bridge-session-start-queue.md
@@ -1,0 +1,5 @@
+---
+'@gemstack/the-framework': minor
+---
+
+The browser bridge gains a session start-queue: the daemon can queue a repo, branch and prompt for the extension to create a session from, and the extension reports back the session it became. Two new routes, `GET /_bridge/start` (claims the next request as it serves it, so two polling tabs cannot create two sessions for one request) and `POST /_bridge/started`. The queue lives daemon-side with a claim that ages out, so a browser that quits mid-creation retries rather than bricking the request. The bridge's input surface is unchanged: the daemon is the producer of a start request and the extension only ever posts back an id, a boolean and a session id.
--- a/packages/the-framework/src/dashboard/bridge-endpoints.test.ts
+++ b/packages/the-framework/src/dashboard/bridge-endpoints.test.ts
@@ -225,3 +225,97 @@ test('a daemon with no answer source degrades to null rather than failing (#1237
     await s.close()
   }
 })
+
+test('the start-queue is served to the extension and its report travels back (#1328)', async () => {
+  const reports: { id: string; ok: boolean; sessionId?: string; note?: string }[] = []
+  let pending: { id: string; repo: string; branch: string; prompt: string } | undefined = {
+    id: 'req-1',
+    repo: 'gemstack-land/the-framework',
+    branch: 'main',
+    prompt: 'Spike and plan the queue ticket',
+  }
+  const s = await serve({
+    token: TOKEN,
+    record: () => {},
+    // Claim-on-read, as the daemon wires it: a second poll must not be handed the same request.
+    start: () => {
+      const next = pending
+      pending = undefined
+      return next
+    },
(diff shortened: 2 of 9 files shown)


suleimansh:
Second commit adds the offline half of the testing story: `check.mjs` gains a fake new-session page and eight cases driving `createSession` and `probeNewSession` in jsdom. No browser, no claude.ai, no session spent - the shape Rom asked for on fake-agent e2e.

What they pin is the flow rather than the markup: repo and branch are chosen before the prompt is sent, the session id is read from the URL rather than from anything on the page, each failure names the control it could not find, and a send that never becomes a session URL is a failure rather than a silent success.

They also caught a real bug. `pickFromMenu`'s already-set check used contains-matching, so branch `main` was satisfied by a repo trigger reading `acme/domain-tools` and the branch was never picked at all. Exact match only for that check now; the case named for it is revert-proven (reverting the fix reports `branch already set to acme/domain-tools`).

Extension harness now 21/21. Unchanged from the first commit: the service-worker loop is still absent, so nothing drains the queue and this PR still changes no runtime behaviour.

CI note: the first run of this commit went red on `@gemstack/framework-dashboard#test`, which this commit does not touch (it only changes `spike/cc-web-extension/`, outside the workspace globs). 633/633 locally; one rerun cleared it. Same flake family as the earlier `#1122` signatures.

suleimansh:
Superseded by #1693 (merged): the daemon half re-landed on today's paths, the extension half built and dogfooded.

🤖 *curated*

Issue #185, opened 2026-07-03
Support running against an existing project (not just from-scratch)


suleimansh:
The CLI is tuned for new/from-scratch projects. It *runs* pointed at an existing project (Claude Code reads and edits whatever is in `--cwd`), but two things make it a poor fit today:

**1. Live detection is not wired to the real workspace.**
`runFramework` selects the preset from `opts.signals`, and the CLI only sets those for `--fake` (`cli.ts:323`). A live run passes no signals, so `select({})` returns confidence 0 and always falls back to the flagship Vike framing (`presets/library.ts:87`) regardless of what is actually in the dir. Point it at an existing Next.js app and it still frames the agent with Vike personas.

Fix: build `FrameworkSignals` from the real cwd (read `package.json` deps + a few marker files) and pass them into `runFramework`, so an existing project gets the right preset. This also helps a partially-scaffolded dir.

**2. Prompts are greenfield-phrased.**
`buildPrompt` / the improve flow say "Build this app end to end" and "the workspace may be empty, scaffold from scratch". For an existing app the framing should be "work within this codebase / add what is asked", not "build the whole thing". The intended home for working *on* an existing/large codebase is scale mode + `CODE-OVERVIEW.md` (#114) in ai-autopilot, which is not wired into the framework CLI.

Fix: detect existing vs empty (we already have `isWorkspaceEmpty` from #182) and pick greenfield vs add-to-existing framing; optionally bridge to scale mode / CODE-OVERVIEW for large repos.

Acceptance: running in an existing project detects its real stack and the agent extends the codebase instead of trying to rebuild it from scratch.

Part of #110. Surfaced while verifying the #181/#182 from-scratch fix.

PR #1340, opened 2026-07-28
fix: drop the how-to-behave rules from the existing-codebase prompt (#1224)

### LINKED ISSUE #1224: Remove ugly babysitting
@suleimansh Can we remove this from the system prompt? Let's avoid infantilizing agents, that's bad.

> This project already exists — do NOT re-scaffold or rebuild it, and do not
replace its structure or swap its stack. Read the existing code first, follow
its conventions, and make the smallest coherent set of changes that adds what
is asked; new files and dependencies are fine when the feature needs them.
When done, summarize what you changed in one short paragraph.

suleimansh:
Closes #1224

The existing-codebase prompt opened with four lines of rules:

> This project already exists — do NOT re-scaffold or rebuild it, and do not replace its structure or swap its stack. Read the existing code first, follow its conventions, and make the smallest coherent set of changes that adds what is asked; new files and dependencies are fine when the feature needs them.

That is how a capable agent already works, so the lines bought nothing.

## What stays

The one thing an agent cannot infer is that the workspace already holds a project, and that is the first line. So #185's whole purpose (this is not a greenfield build, do not scaffold from scratch) survives:

```
Work within the existing codebase in this workspace to deliver: <intent>
When done, summarize what you changed in one short paragraph.
```

I kept the summarize line even though the quoted block includes it. It states a deliverable rather than a manner, and `buildPrompt` and `scaffoldPrompt` both carry the same sentence, so removing it only here would have made the three prompts disagree for no reason. Say the word if you want it gone too and I will drop all three.

## Proof

1571 pass, 0 fail. The updated test asserts the rules are *gone* rather than reworded:

```
assert.doesNotMatch(prompt, /do NOT re-scaffold|swap its stack|smallest coherent|read the existing code first/i)
```

Proven by revert: restoring the four lines fails `extendPrompt names the work and the workspace, and nothing else (#1224)` and nothing else.

One neighbouring assertion was an `or` (`/existing codebase|do NOT re-scaffold/`) that would have kept passing on the dead branch, so it is now tightened to the branch that still means something.

### DIFF
--- a/packages/the-framework/src/steps.spec.md
+++ b/packages/the-framework/src/steps.spec.md
@@ -2,7 +2,7 @@ Driver-backed implementations of ai-autopilot `Bootstrap`'s injectable steps (op
 
 ## TLDR
 
-- Prompt composers: `buildPrompt` (greenfield), `extendPrompt` (#185, existing codebase: extend, don't re-scaffold), `scaffoldPrompt` (#182, hard "create from scratch" directive), `PRODUCTION_GRADE_PROMPT` + `improvePrompt`.
+- Prompt composers: `buildPrompt` (greenfield), `extendPrompt` (#185, names the existing codebase and the intent; the how-to-behave rules were dropped in #1224), `scaffoldPrompt` (#182, hard "create from scratch" directive), `PRODUCTION_GRADE_PROMPT` + `improvePrompt`.
 - `driverBuild()`: one build turn wrapped in synthetic Supervisor plan/dispatch events; picks extend vs greenfield by whether the workspace holds source at build time; re-prompts once with `scaffoldPrompt` when the workspace is still empty after the turn (#182) — unless the turn ended on an await block, which means the agent stopped on purpose to ask (#337/#339).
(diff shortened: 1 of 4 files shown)


packages/the-framework/src/steps.spec.md
```diff
@@ -2,7 +2,7 @@ Driver-backed implementations of ai-autopilot `Bootstrap`'s injectable steps (op
 
 ## TLDR
 
-- Prompt composers: `buildPrompt` (greenfield), `extendPrompt` (#185, existing codebase: extend, don't re-scaffold), `scaffoldPrompt` (#182, hard "create from scratch" directive), `PRODUCTION_GRADE_PROMPT` + `improvePrompt`.
+- Prompt composers: `buildPrompt` (greenfield), `extendPrompt` (#185, names the existing codebase and the intent; the how-to-behave rules were dropped in #1224), `scaffoldPrompt` (#182, hard "create from scratch" directive), `PRODUCTION_GRADE_PROMPT` + `improvePrompt`.
```
brillout, line 5: Hm, it's bad that the spec files mention tickets and history. Too verbose. I'll have a look at how to fix that.

brillout:
Nice :heart_eyes: much cleaner.

PR #1341, opened 2026-07-28
feat(dashboard): colour the log by what a row is, and open it with your prompt (#1199, #1170)

### LINKED ISSUE #1170: Where can I see the user prompt?
In the following I can see TF's sys prompt, but I don't see the sent user prompt.

Image

Only if quick-win.

### LINKED ISSUE #1199: Show agent error msg in red
Show it in red:

Image

Only if quick-win and only for AI. (Let me try to get AI to auto work on this.)

suleimansh:
Closes #1199
Closes #1170

Two reads of the same log, both flagged as quick wins, both in `EventList.tsx`.

## #1199 — failures in red

A failure now renders `text-danger`. Two shapes say it: the agent (or its transport) erroring mid-run (`driver` `error`), and the run settling badly (`end` with `ok: false`).

A **stopped** run is deliberately not red. The user asked for that, so colouring it like a fault would misreport their own action as a failure. Same for a clean finish.

## #1170 — your turn in blue, and first

Per your follow-up comment, both halves:

- The YOU badge renders `text-info`, so your turn reads apart from the agent's work.
- The run's first prompt is hoisted to the top of the log. It is emitted after `session` and `system-prompt`, so the one line you wrote opened three rows down, underneath a char-count summary of a prompt you did not write.

Only the **first** prompt moves. A later turn is part of the conversation and stays where it happened, otherwise a multi-turn session would collect all its prompts at the top and stop reading as a conversation.

**One consequence worth your call:** the arrival-time column is no longer monotonic for that hoisted row, since it genuinely arrived after the rows now below it. It looked like the right trade for "I want to see what I asked without hunting", but say so and I will drop the hoist and keep only the blue.

Both use existing theme tokens (`--danger`, `--info`), so light and dark both work with no new colour values.

## Proof

649 pass, 0 fail. 8 new tests, proven by two reverts:

| revert | tests that fail |
|---|---|
| `rowTone` returns nothing | 3 (agent error, failed run, your turn blue) |
| `promptFirst` returns input unchanged | 2 (hoist, and later-turn-stays-put) |

The negative cases are covered too: a stopped run and a finished run are asserted **not** red, and a log with no prompt at all is left in its original order.


suleimansh:
CI went red on the first run with `RoutineWork.test.tsx > the checkbox is the one global preference, and carries the tooltip`, which this PR does not touch (it imports no EventList). Passed 3/3 locally and passed on a re-run of the same commit, so it is a flake in that tooltip test, not this change. Flagging it rather than staying quiet: it is timing-sensitive and will bite someone else.

PR #1354, opened 2026-07-28
feat: Auto-merge setting — a run's PR can land itself once opened (#1216)

### LINKED ISSUE #1216: Auto-merge
New setting `[ ] Auto-merge`. By default: false.

Should be set to `true` for some prompts, e.g. the "quick-win work" and "consensual work" routines.

suleimansh:
Closes #1216

## What

A new `autoMerge` boolean, default **off**, mirroring `autoOpenPr` through every tier: `the-framework.yml`, `--auto-merge`/`--no-auto-merge` CLI flags, config layers, preferences (global + per-project), the launcher's option rows, and the daemon's spawn flags.

When a run with it on finishes and the handoff opens its PR, the run merges it: `gh pr merge --squash --auto` where the repo allows GitHub auto-merge (the PR lands when its checks pass), direct `--squash` merge where it does not — including the "PR is already green" refusal, where the direct merge is exactly what was meant. A rerun/restart that finds its predecessor's PR already open still merges it (the merge is the half that has not happened yet). A merge that cannot happen is reported on the handoff outcome, never thrown — it must not turn a successful handoff into a failed one.

## Why the drain job turns it on

The issue asks for auto-merge on the "quick-win work" / "consensual work" routines. Those triage jobs only *queue* tickets; the job that produces the implementation PRs is the queue drain. So `AUTO_PM_DRAIN_JOB` declares `autoMerge: true` as data (like `drains` and `pinnedBranch`): what it implements has already been triaged as consensual, quick-win work a human could have vetoed on the queue — its PR is the one kind whose review happened before the run. Everything else stays off unless the repo, a flag, or the launcher checkbox says otherwise.

## Notes

- Squash for both merge forms: a session branch is working history, not a story worth preserving. A repo that disallows squash merges gets a reported failure.
- The launcher row is disabled while `Open PR` is off (nothing to merge), with one documented corner: a run configured `autoMerge` without `autoOpenPr` can still merge a pre-existing open PR on the already-open path.
- gh's refusal wording is matched loosely on the documented GraphQL error strings; an unrecognized rephrase degrades to a reported failure, never a wrong merge.

Tests: the-framework 1592 (all green; one unrelated `daemon-workspace` temp-dir flake reproduced on main), framework-dashboard 652 green, full build green.


packages/the-framework/src/cli.ts
```diff
@@ -222,6 +222,11 @@ Options:
   --no-auto-open-pr      Do not open a draft PR when the session finishes. Opening one
                          is the default; it implies the push above (#1102). Without
                          either flag the-framework.yml autoOpenPr decides.
+  --auto-merge           Merge the session's PR once it is opened (#1216): GitHub
```
brillout, line 225: We might eventually want to remove all CLI flags (they're useless AFAICT), but let's keep AI do its thing for now. (Let's see how far we can get without micro managing AI.)

I also think most tests are useless and just slow AI down. But same thing: let's AI do its thing for now.

PR #1358, opened 2026-07-28
feat: a build that declares small scope skips the production-grade checklist (#1356)

### LINKED ISSUE #1356: Every run pays the production-grade checklist, however trivial the prompt: scope is hardcoded to 'full'
Observed three times on 2026-07-28/29, twice while manually testing #1216:

- Prompt: *"Create a file test-auto-merge-2.md containing the single line 'auto-merge test'. Nothing else."* — the agent finished in 3 tool calls, then the bootstrap's **"Checking the app is production-grade"** phase spent hundreds of tool calls on it: forced fresh rebuild, typecheck (22/22), the full test suite (21 turbo tasks), a CLI smoke test, and four parallel review subagents. The run was stopped by the maintainer before it ever reached the handoff — so the thing the run was fired to test (the PR) never happened.
- Prompt: *"Add new simple file called 'test01.md'…"* with **Disable system prompt** checked — build phase done in 1 turn, $0.023, then the same checklist phase started anyway. The phase is host-side, so no prompt/system-prompt setting can suppress it.
- An accidental `bin.js status` run — intent "status", read-only report — got the same full checklist pass.

## Why

`runFramework` wires the bootstrap's scope step as a constant (`packages/the-framework/src/run.ts`):

```ts
scope: () => ({ scope: opts.scope ?? 'full', intent: opts.intent }),
build: …,
...(handsOff ? {} : { checklist, improve: driverImprove(session, workspaceOpt) }),
```

Nothing ever passes `opts.scope`, so every run is `'full'` and the checklist+improve loop always runs — the only exception is a hands-off driver (#1225), which drops the phase entirely. Meanwhile the agent's own first-turn analysis literally writes `Scope: small — read-only, no code changes` into ANALYSIS_RESULT.md, and that verdict goes nowhere.

## Cost

The trivial task is seconds and cents; the checklist phase is minutes and the dominant token spend of the run. It also sits **between the task and the handoff**, so on a trivial run it delays (or, when the user gives up and hits Stop, prevents) the push/PR.

## Fix directions

1. Feed the scope step instead of hardcoding it: from the launcher/routine (`opts.scope`, plumbing already exists), or from the agent's ANALYSIS_RESULT scope verdict.
2. Small scope drops the checklist/improve steps exactly the way `handsOff` already does — the mechanism is one conditional away.
3. Keep 'full' as the default for real feature runs; this is about not running an app-wide production audit for a one-file change.

suleimansh:
Closes #1356

Observed three times on 2026-07-28/29: a one-line prompt finishes in one turn, then the bootstrap's "Checking the app is production-grade" phase spends minutes and the run's dominant token cost reviewing the whole app — and it sits between the task and the handoff, so twice the maintainer gave up and stopped the run before its PR ever opened.

## How

The agent's protocol already makes it size the work ("whether the scope is small, large, or very large") — the verdict just went nowhere. Now:

- **Protocol**: a `set-scope` signal block (`small` / `large` / `very-large`), documented next to `set-session-name` in `prompts/protocols/signal.md`. Omitting it means the full review runs.
- **Turn-gate**: `parseScopeVerdict` — last block wins, unknown spellings ignored (a guess could skip review work, so only the protocol's exact words count) — emitting a `scope-verdict` run event, deduped like the session name.
- **run.ts**: the Bootstrap fixes its scope before the build, but the agent only sizes the work *during* it — so the gate sits on the checklist step: a declared `small` resolves the pass clean without dispatching anyone, after a log line saying so. No verdict / `large` / `very-large` / hands-off: unchanged. The gate can only ever drop work the agent explicitly said was unnecessary.

## Notes

- `prompts/system_prompt.md` is deliberately untouched: `check-prompt-drift.mjs` pins it to issue #326's body. The signal protocol file is composed into every run's system channel regardless, so the block works even for "Disable system prompt" runs. Worth mirroring a sentence into #326's Scope section when convenient.
- New agent-facing protocol wording — worth a human read.

Tests: 3 new turn-gate tests + 2 run-gate tests; the-framework 1599 (0 fail), dashboard 652 green, root build green.

### DIFF
--- a/packages/the-framework/prompts/protocols/signal.md
+++ b/packages/the-framework/prompts/protocols/signal.md
@@ -5,6 +5,13 @@ When you call setSessionName(<name>) (after creating and checking out the `the-f
 ```
 You do not stop; re-emit it if you rename the session.
 
+## Scope verdict
+When you have sized the work the prompt asks for (e.g. your analysis's scope entry), emit a `set-scope` block naming the verdict, so the framework can right-size the rest of the run — a `small` verdict skips the app-wide production-grade review, whose cost would exceed the change it examines. The first non-empty line is the verdict, one of `small` / `large` / `very-large`:
+```set-scope
+small
+```
+You do not stop; re-emit it if the scope turns out bigger as you work. Omit the block when unsure — no verdict means the full review runs.
+
 ## Ready for merge
 When you call setReadyForMerge() — you believe the work is complete and ready for human review — emit an empty `ready-for-merge` block. This flips the dashboard status from building to ready; it does not stop your turn.
 ```ready-for-merge
--- a/.changeset/scope-gate.md
+++ b/.changeset/scope-gate.md
@@ -1,0 +1,5 @@
+---
+'@gemstack/the-framework': patch
+---
+
+A run whose build declares small scope skips the production-grade checklist (#1356). The signal protocol gains a `set-scope` block (`small` / `large` / `very-large`); a `small` verdict resolves the checklist pass clean without dispatching an agent, with a log saying why. Only an explicit `small` skips — no verdict (an older prompt, a run without the protocols) keeps the full checklist, so the gate can only ever drop work the agent said was unnecessary.
--- a/packages/the-framework/src/events.ts
+++ b/packages/the-framework/src/events.ts
@@ -179,6 +179,14 @@ export type FrameworkEvent =
    */
   | { kind: 'session-name'; name: string }
   /**
+   * The agent's own sizing of the work (#1356), from a `set-scope` block. `small` lets the
+   * run skip the production-grade checklist — an app-wide review would cost more than the
+   * change it examines. No verdict means no claim, and the checklist runs as it always has;
+   * the gate only ever drops work the agent explicitly said was unnecessary. Re-emitted when
+   * the agent revises the verdict mid-run.
+   */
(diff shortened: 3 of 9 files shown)


suleimansh:
@brillout 
And even this PR did not fixed my issue

packages/the-framework/prompts/protocols/signal.md
```diff
@@ -5,6 +5,13 @@ When you call setSessionName(<name>) (after creating and checking out the `the-f
 ```
 You do not stop; re-emit it if you rename the session.
 
+## Scope verdict
+When you have sized the work the prompt asks for (e.g. your analysis's scope entry), emit a `set-scope` block naming the verdict, so the framework can right-size the rest of the run — a `small` verdict skips the app-wide production-grade review, whose cost would exceed the change it examines. The first non-empty line is the verdict, one of `small` / `large` / `very-large`:
+```set-scope
+small
+```
+You do not stop; re-emit it if the scope turns out bigger as you work. Omit the block when unsure — no verdict means the full review runs.
+
```
suleimansh, line 14: I think this is not good

brillout: > the app-wide production-grade review

What is this?

suleimansh: Screenshot 2026-07-29 at 1 44 49

PR #1360, opened 2026-07-28
fix: hold the event loop at a parked gate, and end a run that dies holding one

### LINKED ISSUE #1359: A daemon run dies silently at the backlog gate: the question outlives the process, and picks are read by nobody
Observed live 2026-07-29 ~02:07 (run `2026-07-28T23-07-06-428Z`, a #1356 verification test).

## What happened

1. The run finished its build (one turn, $0.02), declared `set-scope small`, the checklist was skipped (#1356 working as intended) — and the todo-loop asked its per-item gate: *"Start the next backlog item? (23 open)"*.
2. **~2 seconds later the run process was dead**: `run.json` says `status: stopped` at 23:07:14, the PID is gone, `events.jsonl` ends at the `choice` event with **no `end` event**, and `stderr.log` is 0 bytes.
3. The dashboard kept rendering the question from the event log. The maintainer clicked "Stop the backlog loop" five times; all five picks landed correctly in the run's `control.jsonl` (`{"kind":"choice","id":"todo-next","pick":"stop","by":"user"}` ×5, written ~3 minutes after the process died). Nothing read them. No error anywhere.

## Why it looks like an empty event loop, not a crash

A silent exit-0 with no end event and empty stderr is the classic Node pattern: the choice await (`requestChoices` → `opts.requestChoice`) resolves off the control-channel watcher, and if nothing else holds the event loop open — or the watcher isn't actually armed for this spawn shape — Node simply runs out of scheduled work and exits mid-await. Compare #905/#922, which were this same channel silently unwired for daemon-spawned runs.

## Why nobody hit it before today

Until #1356, a dashboard run practically never *reached* the backlog gate: the production-grade sweep sat in front of it for minutes and users stopped the run first (three stopped runs on 2026-07-28 say exactly that). Skip the sweep and the gate is two seconds from the prompt — likely the first time it fired in a daemon-spawned interactive run.

Adjacent: #1225 (choices not working with CC web), #1164 (Queue button) — same "a control surface shows but doesn't act" family, and worth weighing against the standing instinct to simplify the AWAIT/choices mechanism altogether.

## Consequences

- The dashboard shows a live question for a dead run, indefinitely — clicks feel ignored (they are).
- The handoff (push / PR / auto-merge, #1216) never runs, because the death happens before it. On a repo with any open `TODO_AGENTS.md` items, every interactive daemon run now ends here.

## Fix directions

1. Find why the process exits mid-await: is the control watcher armed for daemon-spawned runs at the todo-loop stage, and does the await hold the event loop (a ref'd interval/watcher)?
2. Whatever the cause: a run that dies with an unanswered gate should write an `end` event, and the dashboard should render the question as expired rather than answerable.

Forensics preserved in the run worktree: `.the-framework/worktrees/2026-07-28T23-07-06-428Z/.the-framework/{events,control}.jsonl`, `run.json`.

suleimansh:
Fixes #1359.

## What was happening

In plain terms: the run asked its question, and then the program simply ran out of things to do — so Node quietly shut it down, mid-question. Nobody crashed, nothing errored; the process just had no reason left to stay alive, because the one thing keeping watch for your answer was explicitly marked "don't stay alive for me."

Concretely, a daemon-spawned run at a parked gate has nothing holding the Node event loop:

- the daemon spawns it detached with every stdio ignored and `--no-dashboard` (`daemon-runtime.ts:152`), so no stdin and no HTTP server;
- the claude-code driver spawns and reaps a child *per prompt* (`driver/claude-code.ts:143`), so nothing runs between turns;
- the control watcher is deliberately unref'd (`control.ts:89`, "never keep the process alive just for steering") — a deliberate choice from the actions-abort-hang fix, which this PR does not reverse.

So when the todo-loop's gate parked a bare Promise in `pendingChoices` (`cli.ts`), the loop emptied and Node exited 0 mid-await: no `end` event, empty stderr, `run.json` stuck. The five valid picks in the preserved `control.jsonl` landed after the death and were read by nobody. Verified with a minimal repro (parked promise + unref'd watcher): without the fix the process exits 0 silently; with it, it survives and consumes the late pick.

Until #1356, daemon runs practically never reached this gate — the production-grade sweep sat in front of it and users stopped runs first. Skip the sweep and the gate is two seconds from the prompt.

## The fix (three narrow pieces, no AWAIT/choices redesign)

1. **`gate-keepalive.ts` (new)** — a ref'd idle interval that exists only while at least one held promise is pending; counter-based, shared across overlapping holds, gone the moment nothing is parked. `cli.ts` passes the `requestChoice` promise and the live-chat queue's `next()` (the identical hazard at the settled park) through it. A Stop still releases everything: aborting resolves the parked choices and closes the queue.
2. **`run-store.ts` hardening** — a run that dies holding a gate anyway (kill -9, crash) no longer asks forever: every path that flips a dead `running` run to `stopped` (read-time self-heal, boot reconcile, worktree archive) now appends the surrogate `end` event the process never wrote, folded through `applyEventToMeta` so `pendingChoice` clears from the meta — previously the flip kept it, which is exactly the preserved `run.json`'s state.
3. **`live-state.ts` (dashboard)** — an `end` event expires every open gate, mirroring the meta fold's "a finished run is not awaiting anything" rule, so the question stops rendering as answerable once the death becomes visible.

## Tests

- `gate-keepalive.test.ts` (node:test, injected timer seam): shared timer across overlapping holds, release on resolve/reject, and `hasRef()` on the real timer — the ref IS the fix.
- `run-store.test.ts`: five #1359 cases across self-heal, worktree archive, and both reconcile branches — the surrogate end lands in the log(s) and `pendingChoice` clears; a run that wrote its own end is left alone.
- `live-state.test.ts`: an `end` expires open gates; a gate asked after a continued run's end still opens fresh.
- Full suites green: 1609 (the-framework) + 653 (framework-dashboard).

Changeset: patch `@gemstack/the-framework`.


suleimansh:
Superseded: this exact commit landed on main inside #1361's squash (a0c9db9) — a dogfood test run had branched from the fix checkout, so its auto-merge carried the fix with it. Content-identical, so closing rather than merging a no-op.

Issue #1327, opened 2026-07-27
Goal: fire 10 `Spike & plan` agents concurrently


brillout:
Let's make this work ASAP:

1. Settings:
   - Run in Claude Web
   - Routine work => only select "Spike & plan tickets"
   - Concurrent agents => `10`
1. Click [Trigger routine now]

Result: spins up 10 concurrent agents that spike & plan one ticket each.

We'll need a lock mechanism. I suggest a new one:
- The files `.plan.md` and `.spike.md` are created before the agent starts working on the ticket
- Placeholder content for `.plan.md` and `.spike.md`:
  ```md
  PENDING:<AGENT_ID>
  ```
- Pushed directly to `main` before agent starts working
- Agents refuses to work on a ticket if either `.plan.md` or `.spike.md` exists => agent picks another ticket to work on

(I'll update the `Spike & plan` prompt, stay tuned.)

suleimansh:
Blocker first: #1320. On my machine every CC-web session is a bundle upload with no push access. Yesterday 5 cloud agents did the work, none could deliver it. 10 spike agents here = 10 stranded spikes. I exhausted every local fix (web-setup, relogin, update) and filed it upstream: anthropics/claude-code#81776.

Your machine may be fine (your claude/* branches push). 2-min check from the-framework checkout, then open the session it prints:

```
claude --cloud "Diagnostic: run git remote -v and say: clone or bundle. Change nothing."
```

Clone = #1327 works today, on your box. Bundle = plan B: same 10 agents on the local driver.

The lock: right idea, 3 tweaks:
1. The daemon pushes the PENDING files, not the agents (agents can't push, see #1320) - and locks the whole batch in one commit.
2. Staleness rule: PENDING + no PR + N minutes old = reclaimable, else a dead agent bricks its ticket forever.
3. Composes with #1316 (merged): PR diffs claim work after a PR exists, your locks cover before. Both windows closed.

I can build the code side now (fan-out, one pinned ticket per agent). Waiting on your prompt update for the rest.

suleimansh:
Delivery-side update: the cloud gap that blocked this is closed. A launcher "Run on: Claude web" run now lands its own PR fully autonomously — first one is #1546, chain evidence on #1320. Fanning spike/plan agents out to cloud sessions is no longer blocked on delivery; each web session can push and open its own PR.

suleimansh:
Live 3-wide batch verified end-to-end today (2026-08-17, deliberately small before the full width of 10):

[...]

Issue #1320, opened 2026-07-27
Web runs complete their work but cannot push: cloud container has no GitHub write access unless the account's Claude GitHub app covers the repo


suleimansh:
Found on the 3-run parallel web drive (2026-07-27, sessions 01DRvsPfGzSMAbCP3NmMdiSr / 01SiSWRCyY73fduiXzRS5PKi / 01CWq28wQJw9wqrKavzRAPGs). All three cloud sessions finished their tasks, committed clean work on a branch, and then hit the same wall, quoted from the session:

> the proxy answers 403 "GitHub access to this repository is not enabled for this session" for both api.github.com and an authenticated push (anonymous read works)

So a `claude --cloud` hand-off from a repo the user's Claude GitHub app does not cover produces a session that can clone anonymously, do all the work, and never deliver it: no push, no PR, and from the local side pure silence (the run ended `done` at hand-off; nothing ever arrives). Cross-check: cloud sessions from an account whose app DOES cover the repo push fine (the repo's existing `claude/*` branches).

Why it is nasty for the parallel-agents story: N sessions x a repo without app access = N completed-and-stranded agents, zero local signal, discovered only by opening each session page by hand.

The fix itself is account configuration, not framework code: github.com/apps/claude -> Configure -> grant the org/repo. What the framework CAN do:

1. Document it as a web-run prerequisite (FEATURES.md web-run chapter + the #1318 launcher-warning family: a second preflight fact next to folder trust).
2. Possibly detect it: unclear whether any local, cheap signal exists for "does this account's Claude GitHub app cover this repo" — spike welcome. If one exists, the #1318 launcher warning is the natural surface.
3. The stranded-work recovery path: the session offers a git am patch + PR body; `claude --teleport <sessionId>` can pull the session back locally. Worth a documented recipe.

Follow-up to #1314 (folder trust): same class — a web-run precondition that fails silently after the hand-off.

suleimansh:
Full transcripts confirm (user made the sessions public). All three completed and committed clean work:

- `the-framework/quiet-conversation-commit-pathspec`: found the bug is WORSE than filed — git aborts the entire add/commit on a no-match pathspec, so a project without .the-framework/conversations never committed its pending session archives either. Fix passes only the pathspecs with pending files. 1497 tests pass in-container (the 10 daemon.test.ts failures are the known needs-dashboard-bundle trap, verified pre-existing on the base commit). Sent a git-am patch + PR body as session attachments.
[...]

Issue #1313, opened 2026-07-27
Queue claims are per-machine: two daemons draining the same queue double-assign an entry


suleimansh:
Reproduced live during today's two-machine queue-drain dogfood (2026-07-27, ~18:05Z).

Queue claims (#1253) are resolved against the local daemon's run metas only: `claimedQueueEntries` (`packages/the-framework/src/queue-promote.ts:195`) reads `listRuns` of this machine and asks `resolveRunPr` about *those runs'* branches. A drain running on another machine is invisible: its run meta lives on that machine, its check-off lives on its unmerged branch, and the shared `TODO_AGENTS.md` on main still lists the entry as open.

What happened: one maintainer's sweep worked "Settle-moment UX" and opened PR #1309 (17:29Z). Thirty-six minutes later the other machine's sweep, seeing the entry open and unclaimed, pinned a second agent to the same entry (run 2026-07-27T18-05-38-506Z; stopped by hand before it published a duplicate PR). Two machines draining one queue is the normal shape now, so this will keep happening.

## Fix directions

1. Consult the remote before pinning: for each candidate entry, check whether any open PR's branch already carries that entry checked off (one `gh pr list` + a per-branch queue read, riding the #1028 cache). Claims become "local live run, or any open PR whose branch retires the entry", machine-independent.
2. Or write claims into the shared medium itself: the drain marks its entry in-progress on the queue file (on main or via a claims file) before working it. Heavier: needs a push per claim and a staleness rule.

Direction 1 fits the existing machinery (`resolveRunPr` already reads PRs; `merged-worktrees` already reads branch state) and needs no new writes.

PR #1364, opened 2026-07-29
feat(auto-pm): fan [Spike & plan] out to the concurrency, one PENDING-locked ticket per agent (#1327)

suleimansh:
Closes nothing yet — the code half of #1327; the maintainer's Spike & plan prompt rewrite composes on top.

## What this does

Clicking **[Trigger routine now]** with `Concurrent agents = N`, the other routines unticked, and an empty queue now spins up **N concurrent Spike & plan agents, one ticket each** — the fan-out beat #1327 asks for.

- **Fan-out**: `AutoPmJob.fansOut` (declared on the spike job only) lets a rotation job batch like draining does (`auto-pm.ts`). Each agent gets a `pinnedSpikeJob` prompt — the pin is *appended* to the stock preset so the maintainer's coming prompt rewrite can't silently lose it.
- **PENDING locks** (new `spike-locks.ts`): before agents start, the daemon pre-creates each ticket's `.spike.md`/`.plan.md` as `PENDING:<AGENT_ID>` placeholders — one pathspec-scoped batch commit, pushed to the current branch. The daemon pushes, not the agents, per the #1327 thread agreement (cloud sessions can't push, #1320). Push failure keeps the batch (the commit still guards local forks) and logs.
- **Staleness** (thread rule): a lock still placeholder + older than 60 min (commit time) + untouched by any open PR is released before enumeration, so a dead agent can't brick its ticket. Conservative on every unprovable case (uncommitted placeholder, unreadable `gh`, open PR).
- **Reader awareness**: a PENDING sibling reads as `locked`, not `spiked`/`planned` (`dashboard/tickets.ts`), contributes no effort, and a placeholder plan promotes no quick-win (fails closed in `planned-quick-wins.ts`, now tested).
- **Composability**: locks cover the window *before* a PR exists; the merged #1313 PR-diff claims cover after. The stock (unpinned) spike prompt now also names the placeholder convention, so a single-agent run skips locked tickets too.

## Fallbacks

Missing seams, zero candidates, or a lock that claimed nothing all degrade to the exact pre-#1327 behaviour (one unpinned agent / rotation advance). No settings changes needed — `autoPmConcurrency` (#1204) already exists and caps the batch.

## Tests

- `spike-locks.test.ts`: acquire (batch commit, pathspec scope, push, lost-race skip, commit rollback, push-failure tolerance) and release (stale / young / open-PR / unreadable-PR / uncommitted / real-sibling cases).
- `auto-pm.test.ts`: fan-out to concurrency, partial locks, lock-failure fallback, seamless fallback, in-flight ticket pinning, rotation advance on nothing-to-spike, pinned-prompt shape, catalog flag.
- `tickets.test.ts` / `planned-quick-wins.test.ts`: PENDING placeholders read as locks and promote nothing.

the-framework: 1630 pass · framework-dashboard: 653 pass.


suleimansh:
@brillout This PR also touch the [spike_and_plan.md](https://github.com/gemstack-land/the-framework/pull/1364/changes#diff-fc3a2fea669100d598856d609bce26800442aede35fe7fd54df6e0fe6eae43ef)

packages/the-framework/prompts/presets/spike_and_plan.md
```diff
@@ -1,5 +1,7 @@
 Spike & plan every ticket that has no spike or plan yet, most important first.
 
+A `.spike.md` or `.plan.md` that already exists counts even when it only reads `PENDING:<agent>` — that placeholder is another agent's claim on the ticket, so leave both files alone and skip the ticket.
```
brillout, line 3: Good

brillout:
Couple of smells:
- `HEAD:<current branch>` (see spec.md) => should be `HEAD:origin/main`
- `SPIKE_LOCK_STALE_MS` 60 min => should be longer (spike & planning can take hours, and it's a bug recovery mechanism so, ideally, it's never needed)
- There is a lot of LoC addition, but few LoC removal => does this PR delete all code that is outdated?

Does this PR introduce AI slop (I think yes) and how much? Let's further gauge and keep on the radar how much AI slop the code has (helps us take the right decisions).

suleimansh:
Both code smells fixed in #1394

brillout:
It makes me think that reading the `spec.md` diff can be worth it to catch such issues early. (And my gut feeling tells me we need to know about all high-level flows that exist in the codebase, but let's see.)

That said, we can be less careful about AI slop for TF's experimental parts.

Issue #358, opened 2026-07-11
Turn-boundary gate for showMarkdown() + AWAIT (PLAN approval)


suleimansh:
The #326 prompt's `## Large scope` flow says: write `PLAN_<SESSION_NAME>.agent.md`, `<SHOW_MD>`, `<AWAIT>`. Today `showMarkdown()` alone works (the doc sidebar renders the file), but the AWAIT half has no gate: `showChoices()`/`showMultiSelect()` + AWAIT pause the run (#337/#339), while a markdown + AWAIT stop just ends the turn and the run rolls on without approval. Same gap the old pill had.

Proposed shape, mirroring the other gates: add an `await-markdown` block to the AWAIT protocol (`{ "file": "PLAN_<slug>.agent.md", "title": "..." }`), detect it at the turn boundary, surface an approve/revise single-select (`requestChoices`: Approve / Request changes with a comment), and re-prompt with the outcome. The sidebar already renders the PLAN file, so the gate is just the pause + answer.

One design question for Rom before building: is approve/revise the right answer set, or approve-only for the MVP?

brillout:
How about `showConfirmation()`?

brillout:
So:

showMarkdown(mdContent) => no AWAIT  
showConfirmation(mdContent) => same as showMarkdown() but with AWAIT and a green button `Accept` in the UI

brillout:
(Btw. it makes me think that your `request*` naming you mentioned earlier might fit better, e.g. `requestConfirmation` fits better. Feel free to rename all `show*` to `request*` — although we can keep the name `showMarkdown` because it doesn't request human intervention. As you wish, it's things we can polish now or later, no preference on my side.)

brillout:
> One design question for Rom before building: is approve/revise the right answer set, or approve-only for the MVP?

I think the best would be two buttons:
- Green `Approve` btn => proceed
- Red `Decline` btn => print "Plan declined, awaiting for user instructions" (e.g. in the log area)

Feel free to take shortcuts for the MVP

Issue #322, opened 2026-07-10
Retrieve usage limits


brillout:
## What

Being able tor retrieve the usage limits.

Alternative: we infer how much % the current session consumed from the amount of tokens/dollars consumed. 

## Context

With autopilot, the idea is that I make a prompt and let the AI run for many hours at a time — I only review a big chunk of work at the end.

That's why I think we'll eventually (maybe even already now?) need to be careful about token consumption.

Especially now that we found anti-lazy pills.

Idea: the user can set the maximum consumption per prompt (e.g. 10% of weekly limit).

## See also

- https://github.com/gemstack-land/gemstack/pull/300#issuecomment-4918256151

brillout:
> (e.g. 10% of weekly limit)

I think 20% is a better default. (Because it's 1 working day of 5 working days.)

PR #1373, opened 2026-07-29
Remove the built-in production-grade checklist gate (black-box principle, #1372)

### LINKED ISSUE #1372: Remove the built-in production-grade checklist gate (black-box principle) — and the set-scope machinery that existed to skip it
## Where this comes from

Rom's Discord ping (2026-07-29): "Where does this come from? Why do we have this?" about the production-grade checklist prompt at `src/steps.ts:90-98`, and his follow-up decision:

> I think we should remove it. I think we should leave everything as vanilla as possible, as unopinionated as possible. We treat Claude as black box. We assume it's clever. We don't babysit. And we don't tell it "make clean code". The problem with "make clean code" system prompt like the above is that it potentially conflicts with Claude's own system prompt. It isn't our responsibility to improve Claude. It's Anthropic's responsibility.

Suleiman agreed: remove it, keep Claude a black box.

Provenance, for the record: the checklist gate is day-one AI-written code — the outcome gate of the original build loop (#167, guardrail 3 of #113). It escaped review because #551 moved "all prompting" into `prompts/` but missed the prompt strings living in `steps.ts`.

## What to remove

1. **The built-in production-grade checklist** — `PRODUCTION_GRADE_PROMPT` + `driverChecklist` in `src/steps.ts`, and the run wiring that makes it every run's outcome gate. After removal a run with no domain preset and no serve config passes **no `checklist` step** to `Bootstrap`, which then runs scope → build and stops — exactly the "agent does the work then stops, nothing more" behaviour expected for trivial prompts.
2. **What stays**: domain-preset review loops (#252) — the user explicitly opts into those — and the boot-and-serve gate (#229), which is mechanical verification (does the app boot), not a quality prompt. `domainLoopChecklist` loses its built-in-checklist `fallback` ("never left silently unreviewed" inverts under the new principle: no opted-in review means no review).
3. **The `set-scope` machinery (#1356 / PR #1358)** — the signal protocol, `ScopeVerdict` parsing, the `scope-verdict` event, and the run's small-scope skip. It exists solely to *skip* the production-grade review on trivial runs ("I think there is better ways to do it" — this is the better way: with no default gate there is nothing to skip).
4. **Display fallout**: CLI/terminal success lines currently read `production-grade` vs `prototype`; a review-less run needs a neutral "done" line instead of a misleading "prototype ready".

## Audit: other babysitting layers (follow-ups, not this issue)

Swept `src/` + `prompts/` for similar "improve Claude" prompt layers:

- **`onBeforeMergeableQuality` defaults to `true`** (`registry.ts:394`): automatically applies the maintainability-refactor preset before merge — an automatic "make clean code" step, the clearest remaining violation of the black-box principle. Proposal: default off (or remove); needs Rom's call. → separate issue
- **Preset bodies** (`readability.md`, `ux.md`: "Work until it's exceptionally good…"): opinionated by design but **user-invoked buttons** — the user asked for that work, so they don't violate the principle. Keep.
- **`system_prompt.md` (#326)**: functional protocol text (analysis, session name, alternatives, ready-for-merge) — capabilities, not quality exhortations. No "clean code" language found. Leave to brillout's redesign.
- **`improvePrompt` / scaffold / extend prompts**: mechanical task direction, not quality babysitting. Keep (improve still drives preset/serve blockers).

suleimansh:
Closes #1372. From Rom's Discord call (2026-07-29): *"We treat Claude as black box. We assume it's clever. We don't babysit. And we don't tell it 'make clean code'."*

## What changed (plain language)

**Before:** every build run ended with an always-on "production-grade checklist" — the framework re-prompted the agent up to 5 times to review the whole app (correctness, error handling, auth, …) and looped until the agent reported no blockers. This is why a trivial prompt like *"create test03.md with one line"* went on reviewing the entire app after the file was written, even with the system prompt disabled.

**After:** a run with no domain preset and no `--serve` config ends when the build turn ends. Start run → agent does the work → done. Nothing more.

**What still reviews:** the two things the user explicitly opts into — a domain preset's review loop (`--preset`) drives the checklist exactly as before, and the boot-and-serve gate (`--serve`) still mechanically verifies the app starts. A preset with no loop for the build event now reviews nothing (previously it silently fell back to the removed built-in checklist).

**Gone with it:** the whole `set-scope` machinery (#1356 / #1358) — protocol section, parser, `scope-verdict` event, small-scope skip. It existed only to spare trivial runs from this gate; with no gate there is nothing to skip, and skipping a *preset* review the user asked for would be second-guessing the user.

## Two honesty fixes the removal forced

- **Stops:** budget (#322), quota (#529), and plan-decline (#358) abort the run *between phases*, and the review phase was what observed the abort. With no phase after the build, an aborted run would have settled as `done` / `ok: true`. The build step now throws when the run signal is aborted, and `runFramework` re-checks after the bootstrap — a stopped run still ends as *stopped*.
- **Wording:** the CLI/terminal/dashboard said `production-grade` vs `prototype ready`. A run that configured no review is neither — it now reports plain `✓ done` (dashboard `loopStatus` stays `null` instead of showing a 0-pass "prototype").

## How to manually verify

1. `the-framework --fake --no-dashboard` — the offline demo runs scope → build → deploy and prints `✓ done`, with no `checklist pass` lines.
2. A real trivial run (the #1358 repro): prompt *"Create a file test03.md containing the single line 'just test, delete me' nothing else."* — the run should end after the agent's build turn; no app-wide review prompt should follow (watch the dashboard prompt log).
3. A preset run (`--preset software-development`) still shows the preset's review loop driving checklist passes.
4. Stop-path: start a run with `--budget-usd 0.01` (or hit Stop) — it must end as *stopped*, not as a clean success.

## Breaking (pre-1.0 minor)

Removed exports: `PRODUCTION_GRADE_PROMPT`, `driverChecklist`, `MISSING_VERDICT_BLOCKER`, `parseScopeVerdict`, `ScopeVerdict`, the `scope-verdict` `FrameworkEvent`, `domainLoopChecklist`'s `fallback` option. Agents emitting a `set-scope` block are simply ignored (unknown fence).

## Test status

`@gemstack/the-framework`: 1602/1603 pass, 0 fail. `@gemstack/framework-dashboard`: 653/653. Prompt drift check in sync.

## Related follow-up (not in this PR)

The audit for other "babysitting" layers (#1372 lists it): `onBeforeMergeableQuality` defaults to **true** — an automatic maintainability-refactor before merge, the clearest remaining "make clean code" layer. Needs its own decision/issue.


suleimansh:
@brillout this is the removal you called for on Discord (steps.ts:90-98 and everything wired to it). Review surface per our agreed process: the PR description + a manual test of the trivial-prompt case; the "How to manually verify" section lists the four checks.

Issue #860, opened 2026-07-20
Unattended runs leave code on a branch nobody is told about


suleimansh:
Since #855, auto PM starts runs that write real code. That code has no automated path to a human: it lands on a local branch nobody pushes, and nothing surfaces it.

## What happens to an auto-PM run's work today

1. It gets a worktree and a `the-framework/<session-name>` branch.
2. It does the work and commits; stragglers are force-committed at teardown (`store/worktree.ts:135`).
3. The worktree is removed (`daemon.ts:635`).
4. The branch stays. Nothing deletes it, and nothing pushes it - there is no `git branch -d` and no automatic `git push` anywhere in the package.
5. Only `TODO_AGENTS.md` is read back off it (`queue-promote.ts:47`), so the check-off lands but the code does not.

## Why nobody finds out

- The overview's "Working now" list only includes `status === 'running'` (`dashboard/overview.ts:70`), so a finished run drops out entirely.
- The "needs you" interventions queue has exactly two kinds (`dashboard/interventions.ts:93-102`): an open PR **already on GitHub**, and a run parked on a choice gate. A finished run sitting on an unpushed branch produces neither. So the notification path only fires *after* a human has already clicked "Open PR".
- `RunHistory` rows show a status badge and no branch, no `readyForMerge`, no hint of unpushed work.
- `readyForMerge` is a display flag. Nothing branches on it to trigger a handoff.

The one affordance that exists, `RunHandoffPanel` (push branch / open PR), renders in a single place: after you click into that specific archived run (`RunReplay.tsx:56`). You have to already suspect there is something to look at.

## Not a bug in the handoff design

The click is deliberate, and the reasoning in `run-handoff.ts:231` still holds:

> Deliberately a click, not something teardown does on its own: pushing publishes the agent's work under the user's name to a shared remote, and that is the user's call to make, not a side effect of a run ending.

That is about **publishing**. The gap is **surfacing**: the human is never told there is a decision waiting. When a human started the run they know to go look. Auto PM's runs nobody watched, which is the whole point of them.

## Suggestion

A third intervention kind: a finished run whose branch has commits that are not pushed and not merged. Surfacing only, no publishing, so the existing design call is untouched - the dashboard offers, the human takes.

This overlaps with the "Auto evaluate whether human intervention needed" item on #538, so it may be a slice of that rather than its own thing. Filing it as evidence rather than picking the design.

## Note

[...]

PR #1377, opened 2026-07-29
Fix stranded session work: retry the pre-handoff commit and say "commit-failed" instead of a wrong "committed nothing" (#1376)

### LINKED ISSUE #1376: Trivial run's work is stranded: agent skipped committing, handoff says "session committed nothing", safety commit lands only after the publish decision (#860 redux)
Found live during the basic-use-case manual testing (2026-07-30, the matrix from the Discord agreement). First real break of the trivial-prompt flow.

## What the user saw

Prompt: *Create a file test04.md containing the single line "just test, delete me" nothing else.* Handoff armed: **push the branch**. The agent wrote the file, said it "skipped the usual branch/analysis ceremony" for a throwaway scratch file, signalled ready-for-merge, the run finished cleanly — and then:

```
handoff  ~ handoff skipped: the session committed nothing
```

No push. The worktree was torn down. To the user, the work simply vanished.

## What actually happened

The file survives — on a **local branch nobody is told about** (`the-framework/add-test04-file`, commit `[The Framework] uncommitted changes`), which is #860's exact failure shape resurfacing through a new hole:

1. The agent never committed (its own reasonable trivial-scope judgment; and with *Disable system prompt* it gets no commit instructions at all, so this path fires on essentially every trivial run there).
2. The handoff's empty-check runs at settle (`settleRun` success path, `cli.ts:905`; empty = zero commits or `.the-framework/`-only paths, `run-handoff.ts:311`, #1291) → zero commits → skip.
3. Only *afterwards* does teardown's safety commit (`worktree.ts:141`) capture the dirty state onto the branch — too late for the handoff to publish it.

So the safety net that exists precisely to save uncommitted agent work fires on the wrong side of the publish decision.

## Consequences

- "Push the branch" armed → silently skipped despite real work existing.
- Under **Auto-merge**, the merge half would merge nothing while the actual change sits stranded — the basic auto-merge scenario cannot work for an agent that skips committing.

## Proposed fix

At settle, before the handoff's empty-check: if the run worktree is dirty, make the safety commit **then** (same `[The Framework] uncommitted changes` message), and let the existing paths-based empty filter decide. Genuinely empty runs stay unpublished (`.the-framework/`-only changes still count as empty); real uncommitted work gets pushed/PR'd/merged per the armed intents. Teardown keeps its commit as a no-op backstop.

Repro is deterministic: any trivial prompt where the agent writes without committing, with push/PR/merge armed.

suleimansh:
Closes #1376. Found live during the basic-use-case manual testing: the trivial-prompt run's file **visibly disappeared from the dashboard when Stop was pressed**, with the handoff reporting "the session committed nothing" despite real work existing.

## Root cause (verified forensically)

The pre-handoff safety commit (`cli.ts` `maybeAutoHandoff` → `commitPendingWork`) **failed silently and its result was ignored**:

- `commitPendingWork` swallows every git error into `false`; the handoff call site never checked it.
- The daemon's conversation committer works in the same checkout and is busiest at exactly session end (committing the session's own files) — so the pre-handoff commit can lose git's `index.lock` race.
- With no commit made, `readRunHandoff` saw a branch whose only commits were bookkeeping → `empty` → skip `no-commits`.
- Teardown's *identical* commit succeeded seconds later (repro: commit `8ed1a28` at 00:47:22, after the skip), rescuing the file onto a local branch nobody is told about — the #860 failure shape through a new hole.

Evidence: re-running the empty-check against the surviving branch *with* the rescue commit present yields non-empty (`test04.md` is not a bookkeeping path) — so the skip can only mean the commit was missing at handoff time.

## The fix

1. **`commitPendingWork` retries** (3 attempts, short backoff, injectable for tests): outlasts transient lock contention; persistent failures (no identity, refusing hook) still return `false`.
2. **The handoff honors the result**: a still-failing commit becomes an explicit `commit-failed` skip — new `AutoHandoffSkip` variant, terminal message "the session's pending work could not be committed, so nothing was published (the work stays on its branch)" — instead of judging a branch that is missing the session's work. Publishing an incomplete branch and silently calling it empty were both wrong; saying why is the honest third option. Teardown still rescues the work afterwards, exactly as today.

## How to manually verify

1. Trivial prompt run (e.g. *create test06.md with one line*), handoff armed for push: at Stop, the branch is pushed **with the file's commit** (`[The Framework] uncommitted changes`) instead of "handoff skipped: the session committed nothing".
2. Same with Auto-merge: the merge now has real content to land — this was the blocker for the auto-merge basic scenario.

## Tests

- `commitPendingWork` retries past a transient `index.lock` failure and still reports persistent failure (new).
- Full suite: 1626 pass / 0 fail; dashboard 653/653; typecheck + prompt drift clean.


suleimansh:
Second fix on this branch, found while live-testing the first: **the header's handoff checkbox lied about a push-only run**. The run writes `handoff-armed` as its very first event — before the live channel attaches — so a live tab folds a stream without it, and `handoffState`'s armed-armed fallback showed a ticked **Open PR** on a session the launcher explicitly set to push-only (the run itself was armed correctly: `run.json` said `{push: true, pr: false}`). The armed pair was already mirrored onto the run record for exactly this reader; the view never used it. `handoffState` now takes a seed and `RunView` passes the record's mirror (a streamed `handoff-armed` still wins). Manual check: start a session with Open PR off in the composer → the header should read **Push branch** (checked), not Open PR.

Issue #1390, opened 2026-07-30
Sessions don't end themselves: finished work waits for a human to press Stop


suleimansh:
## In one sentence

When a session finishes its work it doesn't end — so someone has to press **Stop** to get the push / PR / merge that was already armed at launch.

## The pieces involved

Three things share the word "handoff", which makes this hard to talk about:

| name | what it is |
|---|---|
| `armedHandoff` | The **intent**. Built when the session launches from the Push branch / Open PR / Auto-merge options. Ticking the session's checkboxes updates it while the run is alive. |
| `maybeAutoHandoff` | The **gatekeeper**. Runs once, at session end. Decides *whether* to publish. |
| `runAutoHandoff` | The **doer**. Actually pushes the branch, opens the draft PR, and merges it. |
| `handoff` event | The **record** — what happened, shown in the dashboard and written to the journal. |

### How `maybeAutoHandoff` works

It runs once, when the session settles (`cli.ts:1660`, called from `cli.ts:930`), and asks five questions. A single "no" and nothing is published:

1. Is anything armed — were the boxes ticked?
2. Did the user press Stop? If yes, skip: stopping is read as "I didn't want this".
3. Is this a fake / offline run?
4. Did the final commit succeed? If not, skip — otherwise we would publish a branch missing the session's last edits.
5. Does the branch still exist, and what is it called now? (the agent may rename it mid-session)

If all five pass it hands over to `runAutoHandoff`: push → open PR → merge.

Note the question it never asks: **did the agent say the work was finished?** That one is #1363.

### Where the `ready-for-merge` signal fires

It is the last line of the system prompt, and the only place it is mentioned: *"After you're done, consider whether \<SESSION_NAME\> is finished and there isn't any work left to do — if that's the case then call `setReadyForMerge()`"*.

So it fires **mid-session**, right after the agent applies its changes, and it does not stop the turn. Real work can still follow it — the backlog loop runs after the main work settles. "The agent said it's done" is therefore not the same as "the session has nothing left", which matters for the proposal below.

## Why the session doesn't end on its own

Once the work settles, two things keep it alive — and on a trivial run neither needs a human:

- the backlog offer gate (#323), which parks asking whether to drain the queue
- the stay-open chat (F17 in #1375), which keeps the session available for follow-up

## Why this hurts

Because publishing only happens at the end, **Stop** comes to mean two opposite things: "cancel this" and "I'm done, wrap up". Which one you get depends on state that isn't visible:

[...]

PR #1392, opened 2026-07-30
Auto-merge requires the agent's setReadyForMerge signal (#1363)

### LINKED ISSUE #1363: Which session endings run the auto handoff (push / PR / merge)? The verified matrix — and whether Stop-at-a-gate skipping it is the design we want
While verifying the #1359 fix and the #1216 auto-merge live (run `2026-07-29T00-11-42-459Z` → PR #1361), we mapped which ways of ending an interactive daemon session actually trigger the end-of-session handoff. Traced in code and confirmed on :4200 — recording it here to discuss whether each cell is intended.

## The matrix

| How the session ends | Run ends as | Auto push → PR → merge? |
|---|---|---|
| **Stop** while a gate is parked, unanswered (e.g. "Start the next backlog item?") | `stopped` | **Skipped** (`skip('run-stopped')`) |
| Answer the gate ("Stop the backlog loop"), then **Stop** the settled session | clean (`ok`) | **Runs** — this merged #1361 |
| Answer the gate, never touch anything again | stays open for chat (#714) | Not yet — fires when the session ends |
| **Stop** mid-build (agent still working) | `stopped` | **Skipped** |

Mechanics: `maybeAutoHandoff` (cli.ts) skips when `journal.stoppedCleanly()` — i.e. the `end` event carried `stopped: true`, which only happens when the turn loop aborted. Stopping a *settled* session closes the chat queue, the run ends `ok`, and the handoff runs. The code's rationale for the skip: *"A stopped run is a session the user cut short; publishing what it happened to reach is the opposite of what stopping meant."*

## Questions

1. **Is row 1 right?** At the backlog gate, the session's own work is done and verified — the question is only about *starting more*. A user who Stops there arguably means "no more items, wrap up", which is what answering "Stop the backlog loop" means… yet one path publishes (and auto-merges, when armed) and the other silently doesn't. Two clicks that read almost identically, opposite outcomes.
2. On abort, the parked gate resolves to the **recommended** option `by: auto` (so nothing hangs), which logs a `choice-resolved: proceed` for a run the user just killed. Harmless, but the event log reads as if the user chose to proceed. Worth a `by: 'abort'`-ish marker?
3. Should the dashboard make the difference legible — e.g. the Stop button hinting "ends without publishing" while a gate is open and the handoff is armed?

Context: #1102 (handoff), #1216 (merge half), #1359 (the death that used to hide all of this), #846 (unattended semantics).

Filed so @suleimansh can get @brillout's read on rows 1–2 before anyone 'fixes' what may be intended.

suleimansh:
TLDR: an armed merge now runs only when the agent called `setReadyForMerge()` and its own session TODO file has no open entries. Push and PR are untouched; a withheld merge opens the PR as a draft and says why. The system prompt now makes the signal a required terminal action. Implements the rule settled on #1390 (fix #1363).

## The two halves (they must ship together)

1. **The gate** — `maybeAutoHandoff` withholds the merge unless (a) `journal.sawReadyForMerge()` (same signal the on-before-mergeable step already requires) and (b) `TODO_<SESSION_NAME>.agent.md` has no open entries. Never `TODO_AGENTS.md`: the global queue is decoupled from sessions, or auto-merge would never fire while any backlog exists.
2. **The prompt** — the closing "consider whether … then call setReadyForMerge()" is now a required yes/no terminal action. Without this half, gating on the signal silently stops auto-merge on simple runs — live matrix row 3 (#1363, run `00-21-32-410Z`) merged 3s after the PR opened with the signal never emitted.

## Built to be deleted

(b) is the temporary safety belt from #1390 — the agent's word should ultimately be enough. Deleting it = removing `sessionTodoOpen` from `withheldMerge()` and `sessionTodoPending()` in todo-loop.ts; both say so in their comments.

## Visibility

A withheld merge is not a skipped handoff: the branch pushes, the PR opens (as a draft, since `draft: !intent.merge`), and the `handoff` event carries `merge: { outcome: 'withheld', reason }`. The transcript now renders every merge outcome — before this, a merged/withheld/failed merge was invisible in a dashboard transcript.

## Tests

- `withheldMerge` decision table (run-handoff.test.ts)
- `sessionTodoPending` reads only the session file, ignores `TODO_AGENTS.md`, refuses path-escaping names (todo-loop.test.ts)
- transcript lines for all four merge outcomes (events.test.ts)
- Suites: the-framework 1637/0 (1 pre-existing skip), dashboard 655/655 (one PreviewBar flake on first run, clean on rerun)

Not yet done: a live matrix row (needs the main checkout's daemon — will dogfood after merge like rows 1–3).

@brillout this touches `prompts/system_prompt.md` — the new wording is the smallest change I could make that turns the suggestion into a required decision; rephrase freely.

### DIFF
--- a/packages/the-framework/prompts/system_prompt.md
+++ b/packages/the-framework/prompts/system_prompt.md
@@ -52,7 +52,9 @@ Measure "variability":
 
 ## After applying changes
 
-After you're done, consider whether <SESSION_NAME> is finished and there isn't any work left to do — if that's the case then call setReadyForMerge()
+After you're done, decide: is <SESSION_NAME> finished, with no work left to do?
+- Yes: call setReadyForMerge() — required, the work is never merged without it
+- No: don't call it; say what's left instead
 
 
 
--- a/.changeset/merge-requires-agent-signal.md
+++ b/.changeset/merge-requires-agent-signal.md
@@ -1,0 +1,5 @@
+---
+'@gemstack/the-framework': patch
+---
+
+Auto-merge now requires the agent's word (#1363, rule settled on #1390). An armed merge only runs when the agent declared the session done via `setReadyForMerge()` — previously the merge fired on config alone, landing work on main that the agent never said was finished. Second check, a temporary safety belt: the session's own `TODO_<SESSION_NAME>.agent.md` must have no open entries. The global `TODO_AGENTS.md` queue never withholds a merge — it is decoupled from sessions. A withheld merge is not a skipped handoff: the branch is still pushed and the PR still opens, as a draft, and the handoff event says why the merge did not run. The system prompt's closing instruction is now a required terminal action rather than a suggestion, so agents reliably signal when their work is done.
--- a/packages/the-framework/src/cli.ts
+++ b/packages/the-framework/src/cli.ts
@@ -24,10 +24,10 @@ import { hostExecutor } from './host-exec.js'
 import { startDashboard, singleProjectProvider, resolveDashboardBundle, type Dashboard } from './dashboard/index.js'
 import { startRelay, relayPublisher, type RelayPublisher } from './relay.js'
 import { randomUUID } from 'node:crypto'
-import { formatFrameworkEvent } from './terminal.js'
+import { formatFrameworkEvent, mergeWithheldWhy } from './terminal.js'
 import { CLAUDE_CODE_SESSION_LINK } from './session-link.js'
-import { type AutoHandoffSkip, type ChoicePick, type ChoiceRequest, type FrameworkEvent, type OnBeforeMergeableSkip } from './events.js'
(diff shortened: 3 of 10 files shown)


packages/the-framework/prompts/system_prompt.md
```diff
@@ -52,7 +52,9 @@ Measure "variability":
 
 ## After applying changes
 
-After you're done, consider whether <SESSION_NAME> is finished and there isn't any work left to do — if that's the case then call setReadyForMerge()
+After you're done, decide: is <SESSION_NAME> finished, with no work left to do?
+- Yes: call setReadyForMerge() — required, the work is never merged without it
+- No: don't call it; say what's left instead
```
brillout, line 57: Good

Issue #1398, opened 2026-07-30
CI flakes: three tests fail intermittently, unrelated to the diff under test


suleimansh:
PR #1394 (touches only spike-locks files) went red twice on two different tests, neither in its diff. Both pass locally. Known flakes so far:

1. `daemon-workspace.test.ts` — "a project-less topic run spawns in a neutral scratch dir (#1120)": `ENOTEMPTY: directory not empty, rmdir …/.the-framework` — teardown removes the scratch dir while the run's process is still writing it.
2. `RoutineWork.test.tsx` — "with auto-run off the trigger still sweeps once (#1210)": component race, run 30580356732 attempt 2.
3. `PreviewBar` (dashboard) — popup-mount race, seen across several PRs.

Each is a real race in test teardown/mounting, not in the code under test — but every occurrence costs a human a rerun-and-stare, and red CI on green diffs trains people to ignore CI.

suleimansh:
Reopening: #1401 fixed flake 1 of 3 (the daemon-workspace ENOTEMPTY teardown race). Flakes 2 (RoutineWork trigger-sweep race) and 3 (PreviewBar popup-mount race) are still in the code — fix incoming.

PR #1401, opened 2026-07-30
test: retry every teardown rm in daemon-workspace.test.ts (#1398 flake 1)

suleimansh:
Fixes the first (and most fixable) of #1398's three flakes: `daemon-workspace.test.ts` — "a project-less topic run spawns in a neutral scratch dir (#1120)" — failing with `ENOTEMPTY: directory not empty, rmdir …/.the-framework` when teardown removes the scratch dir while the run's detached process is still writing it.

The cure already existed in this same file: the re-home test's teardown passes `maxRetries: 10, retryDelay: 100` to `rm` with a comment explaining exactly this race (#1165's teardown-race tail — the body passes, the cleanup flakes). About half the file's teardowns had adopted it ad hoc; the flake lived in the ones that hadn't.

This hoists the options to one `RETRIED_RM` const with the explanation on it, and uses it in every teardown `rm` in the file — including the previously-plain ones and the inline duplicates. No production code touched; test bodies and assertions unchanged.

Not addressed here (tracked in #1398): the `RoutineWork.test.tsx` component race and the `PreviewBar` popup-mount race — both are dashboard component races, a different shape of fix.

Framework suite: 1638 tests, 0 fail.

### DIFF
--- a/packages/the-framework/src/daemon-workspace.test.ts
+++ b/packages/the-framework/src/daemon-workspace.test.ts
@@ -25,6 +25,15 @@ import { nodeGitRunner } from './project.js'
  * agent at the user's working tree and its uncommitted work.
  */
 
+/**
+ * Teardown options for every `rm` in this file. Starts spawn detached, and daemon-side writes
+ * (meta, sessions, registry) can still be landing when a teardown begins; a concurrent create
+ * inside a dir being removed fails the whole rm with ENOTEMPTY (#1165's teardown-race tail —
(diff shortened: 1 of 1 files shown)


suleimansh:
One more data point, minutes after this PR went up: #1402 (a dashboard-only diff, nowhere near `daemon-workspace.test.ts`) failed its first CI run on exactly this flake — `ENOTEMPTY: directory not empty, rmdir …/.the-framework` in "a project-less topic run spawns in a neutral scratch dir (#1120)". Rerun-and-stare, as #1398 predicted.

Issue #923, opened 2026-07-20
[The Framework] Orphaned runs: spawned runs outlive their daemon and nothing cleans them up


suleimansh:
## TLDR

Runs outlive the daemon that spawned them. A machine accumulates `bin.js` processes that nothing will ever clean up.

## Evidence

A single dev machine, mid-session:

```
14 x node .../packages/framework/dist/bin.js ...
```

Several over a day old (`ELAPSED` past `01-03:01:13`), and most with **`ppid 1`**: their parent daemon exited and they were reparented to init. They are still holding worktrees, and some hold headless Chrome children (`--browser` runs), which is the expensive kind.

## Why it happens

`spawnDetached` deliberately detaches a run so it survives the CLI that asked for it, and `runDaemon`'s teardown disposes previews and watchers but does not stop spawned runs. So a daemon killed with `Ctrl+C` (or one that crashed) leaves every in-flight run running, with nothing tracking them: the next daemon has no record of the previous one's children.

`RunMeta` already carries `pid`/`host`, and `readLiveMeta` self-heals a dead `running` meta to `stopped`, so the *display* recovers. The processes do not.

## Rough shape

Options, roughly in order of how much they change:

- On daemon shutdown, stop the runs it spawned (it knows their pids). Deliberate detachment is about surviving the *CLI*, not about surviving the daemon that owns them.
- On daemon startup, reconcile: read the live metas, and for any `running` meta whose pid is alive but whose owning daemon is gone, either adopt it or stop it.
- A `framework runs --prune` CLI verb, which overlaps #752 (CLI for retained worktrees).

Worth deciding which of these is the intended lifetime before building: "a run survives its daemon" may be a feature for the headless/24-7 direction (#605), in which case the fix is adoption and visibility rather than killing.

## Related

- #752 — CLI for retained worktrees, same cleanup family.
- #605 — the daemon/gateway split, which decides whether outliving the daemon is intended.

suleimansh:
Scoping, before anything is built.

Current count on the machine: 16 `bin.js` runs, 15 of them `ppid 1`, oldest 3d 8h, and 12 headless Chrome processes still attached to them. Restarting the daemon today adds to the pile: SIGKILLing it reparented its two children rather than stopping them.

[...]

Issue #762, opened 2026-07-19
Resuming a stopped run starts a new run instead of continuing it, and the resume often fails


suleimansh:
Reported from use, 2026-07-19. Two problems in the same flow.

Pick a run, stop it, then type another message into it. Expected: it continues that run. Actual: a brand new run appears in the rail.

## 1. It is a new run by design, and that design reads wrong

While a run is live, a message is a `message` control entry the run drains between turns, so it continues in place. Once the run has ended, `RunResumeChat` instead spawns a fresh run with `--resume-session <id>` (#720). The agent conversation does continue, but the dashboard shows an unrelated new row, so the continuity is invisible.

Worth deciding: keep spawning a new run but make the relationship visible (label it as continuing run X, or group them), or hold the row and present it as one continued run.

## 2. The resume frequently fails outright

Observed repeatedly in `test-project` (four of eight recent runs):

```
driver error: No conversation found with session ID: cdffc594-e95e-4fa5-a072-d64df151431e
[framework] claude-code exited (1)
```

The captured session id outlives the conversation the agent CLI is willing to resume. The run then just shows FAILED, with the reason only in the event log.

Either check the session can be resumed before spawning and fall back to a fresh run with a note, or surface the failure as "that conversation is gone" rather than a bare FAILED.

PR #1411, opened 2026-07-31
Sessions end themselves; Stop is a pause, Merge is the second action (fix #1390, fix #1391)

### LINKED ISSUE #1390: Sessions don't end themselves: finished work waits for a human to press Stop

### LINKED ISSUE #1391: Split Stop into distinct actions: pause/resume, cancel, and finish
## Problem

A run accepts five control actions: `stop`, `choice` (answer a gate), `message` (steer/chat), `handoff` (retick the boxes), `bind`.

`stop` is the only one that makes a run stop doing anything, so it has accumulated several unrelated meanings:

- **end the session** — it aborts the run and kills the turn's process tree
- **discard** — `maybeAutoHandoff` skips with `run-stopped`, so nothing is published
- **"I'm finished, wrap up"** — in practice, because a session whose work is done doesn't end by itself (#1390)

One action, three intents, and the framework has to guess which was meant from state the user cannot see.

## The four actions that exist conceptually

| action | what it should do | today |
|---|---|---|
| **Pause / Resume** | freeze the run, change nothing, come back later | **already implemented** — see below — but only the daemon can trigger it |
| **Interrupt** | stop what the agent is doing now, keep the run alive so it can be redirected | #1132 |
| **Cancel** | end the run, publish nothing | this is what `stop` does |
| **Finish** | end the run, publish what is done | absent |

## Pause / resume is already built — it just has no front door

This is the surprise, and it makes this issue much smaller than it looks. `resumeSuspendedRuns` restores a suspended run with:

- `continueRunId` — the **same run id**, so the same worktree and the same dashboard row
- `resumeSession: sessionId` — the **same agent conversation**, not a fresh one
- `queueEntry` — the drain's pin handed back (#1268)
- `RESUME_PROMPT` — telling the agent why it stopped mid-task (#923)

A staleness rule already drops suspended runs too old to resume. This is genuine pause/resume of one session, not the "leave a note for a later run" path the backlog uses (#529).

It is reachable only when the daemon suspends a run at a quota boundary. There is no user-facing way in.

### So what is actually missing

1. A control action to trigger it — there is no `{kind:'pause'}`
2. A Pause button, and a run state that reads *paused* rather than stopped
3. Resume on demand, rather than only when the daemon sweeps

### One real constraint

This works cleanly at a **turn boundary**. Pausing mid-turn means killing a one-shot process and losing that turn's in-flight work — which is the #1132 problem. So "pause between turns" is nearly free; "pause right now, mid-thought" depends on #1132 landing first.

## Proposal

- **Pause / Resume** becomes a first-class user action, exposing the machinery above rather than building anything new.
- **Cancel** keeps today's semantics but is named for what it does, so "publishes nothing" is a property of the action rather than a surprise.
- **Finish** either becomes explicit, or — per #1390 — stops being needed because a session with nothing left ends itself.
- **Interrupt** is #1132 and is not duplicated here.

## Why it matters

While one action means all of these, intent has to be inferred. That is how finished work went unpublished in run `2026-07-29T23-29-17-054Z`: the work was complete, Stop was pressed to mean "I'm done", and it was read as "discard". Details in #1390.

Related: #1132 (interrupt), #1390 (sessions should end themselves), #1363 (which endings run the handoff), #1173 (unclear ending UX).

suleimansh:
Builds the session-lifecycle design settled with @brillout on #1390 and #1391.

## What changes

**A finished session closes itself (#1390).** The only thing that held a healthy session open was the stay-open chat park (`runChatPhase`'s indefinite `messages.next()`). The chat phase now *drains*: a message that already arrived is processed as a turn; an idle queue is the session's natural end. The run ends `ok`, so the handoff — push, PR, and the #1363-gated merge — fires right there, with no Stop press. The `settled` ("waiting") limbo is no longer emitted, which also closes the "done shows early" seam: the pill flips when the session actually ends.

- A follow-up message reopens the conversation: the composer's existing resume path (#762) starts a `prompt` run with `--resume` + `continueRunId`, same row, same branch — exactly the CC-web model Rom asked to replicate.
- The backlog offer gate (#323) is untouched: "after the signal + backlog gate, if nothing needs a human, end" — the gate is still the one thing that may need a human at the end.
- **Carve-out:** a run whose *own terminal dashboard* is the only surface keeps the old stay-open park (`stayOpenChat`). It has no daemon to resume through, so ending would leave its composer a dead end. Daemon-spawned runs (`--run-id`) — the dashboard's sessions, the quick-wins — all self-end.

**Stop is a pause (#1391).** Name, icon, and `{kind:'stop'}` stay (CC-web alignment, per Rom). With the park gone, Stop always means: kill the current turn, publish nothing (`run-stopped` skip unchanged), worktree kept, resumable from the composer ("Session stopped — your next message resumes it"). Cancel = a pause never resumed. No new daemon plumbing needed — the #923 suspend machinery stays what it was (daemon-shutdown recovery).

**Merge is the second session action (#1391).**
- *Live run:* ⋮ menu → "Merge when finished" appends a `{kind:'merge'}` control. The run arms the full ladder and records the **human** authorization, which the merge gate honors instead of demanding the agent's `setReadyForMerge` — a human's word outranks the signal (and the session-todo safety belt). A parked backlog offer resolves to "stop the backlog loop", so one click means "wrap up and land it". Not an abort: the session still ends at its own natural end and merges there.
- *Ended run:* an open, unmerged PR gets a **Merge PR** button in the action bar (`sendMerge` → `mergeSessionPr` → `ghMergePr`, draft marked ready on the way). This is the direct answer to the withheld-merge ending — the Haiku 0/5 pattern (#1334) that leaves draft PRs like #1399/#1404 parked on a human.

## Notes for review

- A Stop pressed while parked in the (old) chat used to end the run `ok:true` and publish; that ambiguity is gone with the park — Stop now always lands on the abort path.
- Messages that arrive in the sub-second window between the final queue check and process exit are dropped, same as they already were on Stop (`close()` discarded pending messages). The composer flips to the resume path as soon as the `end` event streams, so the practical window is the control-file tail poll (~300 ms).
- `agentSettled` / the `settled` event / the "waiting" badge remain as renderers for archived logs and for the terminal-dashboard stay-open mode; nothing else emits `settled` anymore.
- No prompt files touched.

## Tests

- the-framework: 1644 pass / 0 fail (new: chat drain + self-end, `takeQueued`, `merge` control round-trip, `mergeSessionPr` ×3)
- framework-dashboard: 662 pass / 0 fail (new: Merge PR button on an open PR, none once merged; live-menu Stop+Merge pair, neither once ended)
- Live dogfood: PASSED — four runs on this build (self-end + auto-merge; one-click Merge-when-finished; draft-PR Merge button; resume). Evidence in the first PR comment.


suleimansh:
Live dogfood on this build (daemon on :4200 from this branch, main checkout, four runs, all Opus):

**Run A — `2026-07-31T00-17-00-134Z`, full ladder armed.** Agent signalled `ready-for-merge` unprompted, parked at the backlog gate; answered "Stop the backlog loop" — and that was the last click. Journal: `choice-resolved(stop, user)` → `end ok:true` → `handoff done, PR #1412, merge:{merged}`. **PR #1412 merged 66 s after launch, Stop never pressed, zero `settled` events in the stream.** The "done shows early" seam is gone — the pill flipped at the actual end.

**Run B — `00-22-56-033Z`, merge OFF.** While parked at the gate, clicked ⋮ → **Merge when finished**. One click did the whole chain: `handoff-armed merge:true` re-announced, the gate self-resolved to stop, session ended itself, **PR #1413 merged** (human-authorized).

**Run C — `00-25-54-606Z`, merge OFF.** Gate answered normally → self-end → **draft PR #1414** + no merge (correct: not armed). The ended run's bar showed **Merge PR**; clicking it marked the draft ready and merged it (test23.md on main), and the button withdrew once the PR read MERGED.

**Resume.** Messaged the ended run C: continuation ran in the same run row, resumed the same agent conversation (same sessionId), answered exactly `test23.md`, ended itself again, and the handoff skipped `already-open` — no duplicate PR.

Environment restored after: stock daemon back on :4200, local main rebased onto origin/main (2 bookkeeping commits remain unpushed, #1397). Cleanup owed: test21/22/23.md on main + their run branches.

Issue #1420, opened 2026-07-31
Improve lock mechanism


brillout:
The lock mechanism should be broadened: a ticket should be assigned to a single agent, regardless whether it's planning or implementation.

I suggest a new file `tickets/<DATE>_<SLUG>.lock.md` with this content: `CLAIMED:<SESSION_ID> (<SESSION_NAME>)`. (Feel free to change the content to whatever quick implementation you see.)

Notes:
- In the future, that single agent can spin up several child agents. So the assigned agent acts as coordinator.
- 6h is better than 60mn, but still too short.
  - A coordinator can live for multiple days if not weeks (e.g. if tasks are queued in `TODO_AGENTS.md` => sub-agents can be instantiated many days later).
  - I think we should remove this lock retrieval mechanism (ideally, we never need it because we ensure agents don't "hang", and it's the user's responsibility to watch over the AI queue).

@suleimansh Needed for proper spiking & planning => highest prio.

See also:
- https://github.com/gemstack-land/the-framework/issues/1384

brillout:
I went ahead with the following, but not sure if it's enough?
- https://github.com/gemstack-land/the-framework/commit/256f916cf58ef5135397c9e9a6d8bd98e6f320ae

Also we should remove the old mechanism.

suleimansh:
Not enough yet — the lock only exists in the prompts, no code creates or reads `.lock.md`. The daemon still uses the old `PENDING:` locks, so the two can't see each other and tickets can still get double-booked.

Also, the agent can't place the lock itself: it pushes at the *end* of a session, on its own branch — but the lock must be on **main before work starts**. So the daemon should write + push the claim at assignment time, like it does today for the old locks.

Plan (one PR): daemon writes/pushes `.lock.md`, dashboard reads it, old PENDING mechanism + 6h expiry deleted. Release = the lock file is removed in the session's own PR, plus a manual release button for dead agents.

Will start on this unless you object.

brillout:
Makes sense.

brillout:
> plus a manual release button for dead agents.

Only if it's a quick-win

suleimansh:
[...]

Issue #1383, opened 2026-07-30
Dashboard: live run view flickers blank mid-run; white screen for a while after closing a run


suleimansh:
Two occurrences during the manual test matrix on 2026-07-30 (daemon on main build, #1373/#1377 merged):

1. Run `2026-07-30T00-14-58-192Z` (matrix row 2): mid-run, the message area went blank for a stretch — "screen was showing no message, like flickering" — then recovered and showed the full transcript.
2. Run `2026-07-30T00-21-32-410Z` (matrix row 3): after closing the run view, the page went **white for some time** before rendering; on reload the full transcript (including the handoff line) was there.

No console logs captured yet, no repro steps — filing as an observation with two data points so it stops being folklore. Both runs were trivial single-turn builds, so this isn't a long-transcript rendering cost.

Plausible territory: live-channel attach/reattach and event replay (the same area as #1377's header seeding fix, where the opening events predate the live channel's attach). Next occurrence: grab the browser console + note whether it coincides with reattach.

suleimansh:
Third occurrence, now with a concrete repro (run `2026-07-30T13-52-08-019Z`, daemon on the main build): pressing **Stop** on a settled session flips the view to **"This session has no events."** (with a flicker), and it stays that way until a manual page refresh — after which the full transcript is back. So the blank state is a live-channel/view reset on stop, not data loss: the journal is intact throughout.

suleimansh:
Status check on the three observations, against today's main (post-#1402, post-#1411):

**Observation 3 (Stop on a settled session → "This session has no events.") — fixed, twice over.** #1402 landed the direct cure (an empty archive read never replaces a populated feed). And since #1411, the state that triggered it no longer exists: a settled session ends itself, so there is no parked-on-Stop moment left to press Stop in.

**Observation 2 (white screen after closing a run) — almost certainly #1194, not this issue.** The white-blank has its own ticket (`tickets/2026-07-25_bug-white-blank.md`, P8) with a deterministic repro — hovering the Serve menu item — and #1196 explicitly landed as a mitigation, not a fix. The ⋮ menu with its Serve state is mounted in the run view's action bar, which is exactly what gets touched when leaving a run. Suggest tracking the root cause there and dropping it from this issue's scope.

[...]

Issue #1067, opened 2026-07-23
Run on: run a session on a connected device without leaving the dashboard (remote-run relay)


suleimansh:
Part of #1049. The architectural half of the "Run on" UX. Today, picking "a device I have" **navigates** the browser to that device's daemon (the SPA is served by it, so everything stays same-origin). Users expect a connected device to be a **true peer run-target** like Current device / GitHub Actions: pick it, **stay on this dashboard**, type a prompt, submit, and the session runs on the remote device and streams back here, with the session appearing in the current view.

## The design fork the plan must resolve

The remote daemon's RPC (`/_telefunc`), event Channel, and `<img>` screencast are same-origin + token-guarded (cookie). Making a connected device a run target without navigation means the LOCAL dashboard has to drive the REMOTE daemon. Two ways:

- **(a) Cross-origin from the browser** — CORS on the remote's mount, the token as a bearer header on cross-origin `fetch`/`EventSource`, cross-origin screencast. This is the "repoint" the #1049 plan deliberately rejected for reworking all three transports.
- **(b) Server-side relay (recommended)** — the LOCAL daemon proxies run-submit + the event stream to the REMOTE daemon over the token (the local daemon holds the saved device's token). The browser stays same-origin with its own local daemon; no CORS; the token never leaves the two daemons. The local daemon becomes a thin relay for a run that executes remotely.

Recommend (b). It reuses the token+bind from #1051, keeps the browser same-origin, and is the natural shape of #230 (hosted run relay) and #806 (hosted mode).

## Rough scope (to be firmed by the plan)

- A "remote" run mode: when the selected target is a saved device, the local daemon dispatches the run to that device's daemon and relays its events/branch/PR back into the local run view.
- The run's identity, worktree, and PR live on the remote device; the local side is a viewer + control channel.
- Reuse the existing run-view pipeline (same `DriverEvent` shape) so the relayed stream renders unchanged.

Relates to #230 (hosted run relay: watch one run from multiple browsers) and #806 (hosted mode). Depends on the token+bind lane (#1051, merged).

Issue #426, opened 2026-07-12
[The Framework] retire page.ts: port the per-run dashboard + relay to the new dashboard


suleimansh:
The #405 keystone (daemon serves the new Vike + Telefunc dashboard) is done and shipped: PR-0 #419, PR1 #421, PR2 #423, PR3 #425. The daemon now serves the new dashboard at / by default, browser-verified. page.ts survives only as the daemon's /legacy escape hatch plus two consumers that still run on it and must be ported before it can be deleted.

## 1. Per-run foreground dashboard
`framework "prompt"` (without --no-dashboard) and `--resume` call `startDashboard({ cwd })` with no `clientBundleDir`, so they still serve page.ts (cli.ts ~811 and ~1341). Port them to the new dashboard: serve the prerendered bundle + mount /_telefunc for that run's project (register the run's cwd, or a single-project mode).

## 2. Relay (design-gated on Rom)
`relay.ts` serves `dashboardHtml` read-only at /r/:id/ (page.ts + SSE). The new SPA needs a live Telefunc backend, so the relay can't just swap to it. Needs a rethink, likely a static snapshot of the event log rendered without a server. This touches the self-contained / shareable design, so it needs Rom's input before building.

## 3. Delete page.ts (after 1 + 2)
Remove page.ts + sse.ts + dashboardHtml + the legacy routes in dashboard/server.ts handle() (/events, /api/*, /stop, /choice, /api/start) + isLegacyRoute + the /legacy branch + the FRAMEWORK_DASHBOARD=legacy escape hatch + the legacy DashboardOptions fields. Migrate or drop the server.test.ts tests that hit the legacy /api/* endpoints directly.

Resume in a fresh session from here. Part of #405.

suleimansh:
Part 2 (the relay, `--share`) is the last thing still on the old page.ts. My vote is option 1 below. Any objection?

Relay in plain terms: with `--share`, a run streams its events to a small shared server so a teammate on another machine can watch it live in the browser, read-only. It holds no files, just the event stream flowing through it.

Why it is not a straight swap: the old dashboard draws itself from the event stream alone, so the relay just forwards events. The new dashboard also reads files (run history, docs, project list) that the shared server does not have, so it would show up mostly blank. It needs a watch-only, events-only mode to work on the relay.

Options:
1. Live watch-only: serve the new dashboard on the relay with just the live event feed, rendering only from events, no history/docs rails. Keeps todays behavior with the least new code. (my vote)
[...]

PR #1423, opened 2026-07-31
Live feed: a reconnect never blanks a populated transcript (#1383)

suleimansh:
Fixes the last live observation on #1383 — the mid-run blank flicker. Mechanism (from the assessment comment there): every subscribe replays the whole log before following live, and `use-live-events` cleared its feed on each (re)subscribe. So after a transient channel drop, the "stream lost" banner cleared on resubscribe and the transcript sat **empty, with no banner**, until the replay re-streamed it line by line. A daemon restart mid-run did this to every open view.

## What changes

- **Server** (`events-tail.ts` / `events.telefunc.ts`): the on-disk tail owns its first pull and reports the replay boundary; `onEvents` sends a wire-only `stream-sync` marker at it, once per subscription. The marker is not a `FrameworkEvent`, is never journaled, and the follower only starts after it — no appended line can slip in ahead of it. It still fires on a missing/unreadable log (an empty replay is not a marker withheld; a client waiting forever would freeze its feed).
- **Client** (`use-live-events.ts`): the *first* subscribe of an effect is unchanged — the pane was just cleared, the replay streams straight in. A *reconnect* keeps the populated feed on screen, buffers the replay, and swaps atomically on `stream-sync` — the feed never shows less than it already showed, the same rule #1402 established for the archive read. In-memory sources (relay #426, relayed device runs #1067) have no replay boundary to report and send no marker; a reconnect to those swaps at a 1.5 s grace deadline instead. A reconnect that dies mid-replay drops its partial buffer rather than swapping in a shorter feed.

## Notes for review

- Only `onEvents` grows the union (`LiveFeedEvent = FrameworkEvent | StreamSync`); nothing downstream of the hook sees the marker.
- Duplicate-history protection is unchanged: the swap replaces (never appends to) the shown feed, and `currentRunEvents` still scopes mid-channel log truncations.
- Main currently has 3 unrelated failing tests (stale prompt-spec assertions from #1420's `ticketing_format.md` edit — the #1388 pattern); fix PR for those follows separately, after which this PR's CI should be green.

## Tests

- the-framework: +2 (replay boundary lands between replay and follow; fires even for a not-yet-created log)
- framework-dashboard: +4 (first-subscribe streaming + marker swallowed; atomic swap on reconnect; grace-deadline swap for markerless sources; mid-replay death keeps the old feed) — 666/666
- Suites otherwise green except the 3 pre-existing #1420 assertion breaks, reproduced on clean origin/main.

### DIFF
--- a/.changeset/live-feed-replay-sync.md
+++ b/.changeset/live-feed-replay-sync.md
@@ -1,0 +1,6 @@
+---
+'@gemstack/the-framework': patch
+'@gemstack/framework-dashboard': patch
+---
+
+The live run feed no longer blanks mid-run on a stream reconnect (#1383). Every subscribe replays the whole log before following live, and the client used to clear its feed first — so after a transient channel drop the "stream lost" banner cleared and the transcript sat empty, bannerless, until the replay refilled it. The on-disk tail now sends a wire-only `stream-sync` marker when its replay is delivered, and a reconnecting client buffers the replay and swaps atomically on it — the feed never shows less than it already showed, the same rule #1402 set for the archive read. In-memory sources (the relay, relayed device runs) send no marker and fall back to a short grace deadline; a reconnect that dies mid-replay drops its partial buffer rather than swapping it in.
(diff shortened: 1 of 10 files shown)


suleimansh:
Live verification on this build (daemon on :4200 from this branch), reproducing #1383's exact scenario — a channel drop under an open live-run view:

1. Started a trivial run (publishing disarmed), let it park at the backlog gate with a full transcript on screen.
2. `kill -9` the daemon: the view showed the "daemon is not answering" banner with the **transcript fully retained** underneath.
3. Restarted the daemon: on reconnect the feed **never blanked** — the replay swapped in atomically and the view (transcript, gate panel, handoff line) was byte-identical to before the kill. On the merged-main build this is the state that flashed empty and refilled line by line.
4. The parked gate still answered normally after the reconnect; the run ended itself and the disarmed handoff skipped `not-armed`.

One pre-existing quirk observed (not a regression): after a reconnect the feed's displayed times show the replay's arrival time (`stampReceived` re-stamps on the way in), so every row shares one timestamp. Same behavior before this PR; worth its own small issue if it bothers anyone.

PR #1426, opened 2026-07-31
remove planned-quick-wins.ts: the planning agent queues its own quick wins (#1420)

suleimansh:
Per the #1420 thread ([Rom's question](https://github.com/gemstack-land/the-framework/issues/1420#issuecomment-5144467262), agreed by @suleimansh): the machinery goes, the agent does it.

**Why removal is safe:** the module was already dead — it parsed the pre-#1420 keys (`Effort: quick-win`, `Consensus: consensual`), which the new 0-10 `Effort`/`Uncertainty` format never produces, so it has promoted nothing since the format changed.

**The replacement is one prompt line**, in `ticketing_format.md`'s plan notes (@brillout — your file, please reword as you see fit):

> If the plan concludes the ticket is an unambiguous quick win (low `Effort`, `Uncertainty: 0`), add `- "Implement tickets/<TICKET>.md"` to `TODO_AGENTS.md` in the same commit as the plan, under the right `## Priority` — a consensual quick win should not wait for another agent to re-read the plan and reach the same conclusion

Same pattern as the `.lock.md` release in #1425: the agent's own PR carries the state change, no daemon pass, no extra agent turn, no threshold hardcoded in code. Failure mode is benign — a missed quick-win just waits for triage, exactly as it does today.

**Removed:** `planned-quick-wins.ts` + test + spec (−507 lines), the `promotePlans` seam in `auto-pm.ts`, its wiring in `daemon-services.ts`, and its two loop tests. Net −562.

Tests: 1624 pass, 0 fail.


packages/the-framework/prompts/ticketing_format.md
```diff
@@ -74,6 +74,7 @@ Notes:
 - Covers both spiking (e.g. high-level research without implementation plan) and planning (e.g. concrete implementation proposal)
 - The `.plan.md` file can be modified multiple times over an extended period (e.g. a ticket requiring repeated human intervention, transitioning from spiking to concrete plan)
 - All sections are just proposals and optional: you can use any headings with any format
+- If the plan concludes the ticket is an unambiguous quick win (low `Effort`, `Uncertainty: 0`), add `- "Implement tickets/<TICKET>.md"` to `TODO_AGENTS.md` in the same commit as the plan, under the right `## Priority` — a consensual quick win should not wait for another agent to re-read the plan and reach the same conclusion
```
suleimansh: @brillout

brillout: No, bad: no clear separation, it isn't the job of ticketing_format.md, see the new `presets/spike_and_plan.md`

suleimansh: Agreed — moved. `ticketing_format.md` is untouched now; the rule rides in the queue entry `spike_and_plan.md` writes ("Create tickets/<TICKET>.plan.md — if the plan concludes … also add 'Implement tickets/<TICKET>.md' in the same commit"), and the fan-out's pinned prompt mirrors it, since pinned agents write plans without reading a queue entry.

packages/the-framework/prompts/presets/spike_and_plan.md
```diff
@@ -1,4 +1,4 @@
 For every ticket `tickets/<TICKET>.md` without `<TICKET>.plan.md` and `<TICKET>.lock.md`, add the following to TODO_AGENTS.md
-- "Create tickets/<TICKET>.plan.md"
+- "Create tickets/<TICKET>.plan.md — if the plan concludes the ticket is an unambiguous quick win (low Effort, Uncertainty 0), also add \"Implement tickets/<TICKET>.md\" to TODO_AGENTS.md in the same commit"
```
brillout: Still bad. This is conflating *planning* (all tickets) with quick-win *implementation*.

brillout: The quick-win task should be a separate prompt. So there are two tasks here. See `presets/triage_quick.md`.

suleimansh: Done — the PR is now a pure removal: spike_and_plan.md and ticketing_format.md are both untouched, and the pinned fan-out prompt lost the quick-win sentence too. The separate prompt already exists in the catalog: `presets/triage_quick.md` picks and queues quick wins on its own rotation turn, and with #1420's format it now has plans' Effort/Uncertainty to read. If you want, one line there ('a ticket whose .plan.md concludes low Effort and Uncertainty 0 is exactly such a pick') would make that explicit — your file, your wording.

brillout: I agree the triage_quick.md could be better aligned, but let's keep it like that for now (I think it's fine, and I'm curious to see what tickets agents will pick).

PR #1427, opened 2026-07-31
sync system_prompt.md from #326 (drift check red on every PR)

suleimansh:
Issue #326's system-prompt block was edited today (~15:26), so the `drift` check now fails on every PR (first seen on #1426; the #1425 merge CI at 15:17 was still green). This is the documented fix: copy block 1 into `prompts/system_prompt.md` verbatim.

What the issue edit changed (all followed verbatim here):
- `TODO_FILE: \`TODO_AGENTS.md\`` glossary line restored
- the very-large-scope bullet uses `<TODO_FILE>`/`<SHOW_MD>` placeholders again
- the terminal instruction is back to "consider whether <SESSION_NAME> is finished … then call setReadyForMerge()" — i.e. the #1392 "required, the work is never merged without it" tightening is gone from the issue

@brillout flagging that last one for visibility: if the softening was deliberate, nothing to do (the merge gate itself is code — `withheldMerge` still blocks a merge without the signal, so the worst case is more withheld draft PRs on weaker models, per the #1334 tier table). If it was an accidental restore of an older block, edit the issue and this file can be re-synced.

Drift check passes locally against the live issue; suite 1650/0.

### DIFF
--- a/packages/the-framework/prompts/system_prompt.md
+++ b/packages/the-framework/prompts/system_prompt.md
@@ -31,10 +31,9 @@ Do the following before applying your first change.
 
 ### Session name
 
-1. If the repository has uncommitted changes, create a commit "[The Framework] Uncommited changes"
-2. Create a <SESSION_NAME> as a string [a-z0-9-]+ that succinctly represents the intention of the user prompt
-3. Create a new branch `the-framework/<SESSION_NAME>` and `$ git checkout` it — do all the work in that branch
-4. Call setSessionName(<SESSION_NAME>)
+1. Create a <SESSION_NAME> as a string [a-z0-9-]+ that succinctly represents the intention of the user prompt
+2. Create a new branch `the-framework/<SESSION_NAME>` and `$ git checkout` it — do all the work in that branch
+3. Call setSessionName(<SESSION_NAME>)
 
 
 ## Before applying changes
@@ -52,7 +51,9 @@ Measure "variability":
 
 ## After applying changes
 
-After you're done, consider whether <SESSION_NAME> is finished and there isn't any work left to do — if that's the case then call setReadyForMerge()
+After you're done, decide: is <SESSION_NAME> finished, with no work left to do?
+- Yes: call setReadyForMerge() — required, the work is never merged without it
+- No: don't call it; say what's left instead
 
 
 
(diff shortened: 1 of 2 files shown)


packages/the-framework/prompts/system_prompt.md
```diff
@@ -51,9 +52,7 @@ Measure "variability":
 
 ## After applying changes
 
-After you're done, decide: is <SESSION_NAME> finished, with no work left to do?
-- Yes: call setReadyForMerge() — required, the work is never merged without it
-- No: don't call it; say what's left instead
+After you're done, consider whether <SESSION_NAME> is finished and there isn't any work left to do — if that's the case then call setReadyForMerge()
```
brillout: Mistake?

suleimansh: Yes, I think its mistake

packages/the-framework/prompts/system_prompt.md
```diff
@@ -19,7 +20,7 @@ If it isn't clear what you should do (e.g. unclear scope, unclear user prompt),
 ### Scope
 
 - If the scope of what you'll work on is *large*, create a `PLAN_<SESSION_NAME>.agent.md` of what you'll work on, <SHOW_MD>, <AWAIT>
-- If the scope is potentially *very large* (e.g. spans over many hours/days of work), consider adding entries to `TODO_AGENTS.md` and show the new entries to the user via `showMarkdown()`
+- If the scope is potentially *very large* (e.g. spans over many hours/days of work), also create a <TODO_FILE> (backlog of follow-up tasks) and <SHOW_MD>
```
brillout:
```suggestion
- If the scope is potentially *very large* (e.g. spans over many hours/days of work), consider adding entries to <TODO_FILE> (backlog of follow-up tasks) and show new entries <SHOW_MD>
```

packages/the-framework/prompts/system_prompt.md
brillout: Smells like a bug though... do agents always make sure their work on top of origin/main after git fetch?

Btw. the whole commit that commits uncommitted changes isn't needed anymore since we have worktrees. (This is the kind of bugs we should manage to fix without much discussion. Such major bug (a blocker really) should always be fixed.)

suleimansh: Root cause found, and it's not an agent-rebase bug: issue #326 hadn't been edited since Jul 27 — #1392 tightened the wording in the repo file only, so the repo was *ahead* of the issue, and this PR's first sync followed the drift check's "issue is the source of truth" instruction backwards into a revert. Fixed properly now: the issue is updated (tightened wording restored, your <TODO_FILE> suggestion, and the uncommitted-changes step removed per your comment — worktree runs start clean, the step only made junk commits), and the branch re-synced to it. Drift green, suite 1650/0.

brillout: Hm, I still consider it a bug (maybe a merge conflict bug), but let's see if this occurs again.

PR #1440, opened 2026-07-31
spike & plan queues the <COUNT> most important tickets, default 10 (fix #1421)

### LINKED ISSUE #1421: Parameterized `Spike & plan tickets` task
The new prompt (https://github.com/gemstack-land/the-framework/commit/bd2d5b55bb2e2075c945ce4999cca446f8ae40fc) makes the agent add all tickets to the AI queue.

Can we parameterize the prompt instead? Default value: `10` (agent will ad 10 tickets instead of all).

Highest-prio, but we can postpone if not a quick-win.

suleimansh:
Fixes #1421 — the prompt made the agent queue plan work for *every* open ticket; it now caps at the `<COUNT>` most important, default 10.

The parameter is prompt-native: `<COUNT>` in the body, defined in a footer line (`COUNT: 10`), following the prompts' own placeholder convention (`AWAIT:`, `SESSION_NAME:`, …). So changing the default — or overriding it from a derived prompt by appending a different `COUNT:` line — is a one-line prose edit, and no launcher UI or param machinery was added (that half stays postponable per the issue's "postpone if not a quick-win").

@brillout — prompt-file change (`prompts/presets/spike_and_plan.md`), your wording call: happy to rephrase.

Cost context from the first live fan-out: a plan run costs ~$4.70, so an uncapped sweep over a large backlog compounds fast; 10 ≈ $50 per sweep.

Suite: the-framework 1623/0 (1 skipped). Changeset: patch. `prompts.generated.ts` is gitignored and regenerates on build, so the diff is the prompt + changeset only.

### DIFF
--- a/packages/the-framework/prompts/presets/spike_and_plan.md
+++ b/packages/the-framework/prompts/presets/spike_and_plan.md
@@ -1,4 +1,6 @@
-For every ticket `tickets/<TICKET>.md` without `<TICKET>.plan.md` and `<TICKET>.lock.md`, add the following to TODO_AGENTS.md
+For the <COUNT> most important tickets `tickets/<TICKET>.md` without `<TICKET>.plan.md` and `<TICKET>.lock.md`, add the following to TODO_AGENTS.md
 - "Create tickets/<TICKET>.plan.md"
 
 Put the entries in the right `## Priority` following a mix of sensible criteria after reading `tickets/<TICKET>.md` (e.g. if ticket seems low effort => higher priority).
+
+COUNT: 10
--- a/.changeset/spike-plan-top-n.md
+++ b/.changeset/spike-plan-top-n.md
@@ -1,0 +1,5 @@
+---
+'@gemstack/the-framework': patch
+---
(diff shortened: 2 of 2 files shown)


packages/the-framework/prompts/presets/spike_and_plan.md
```diff
@@ -1,4 +1,6 @@
-For every ticket `tickets/<TICKET>.md` without `<TICKET>.plan.md` and `<TICKET>.lock.md`, add the following to TODO_AGENTS.md
+For the <COUNT> most important tickets `tickets/<TICKET>.md` without `<TICKET>.plan.md` and `<TICKET>.lock.md`, add the following to TODO_AGENTS.md
```
suleimansh, line 1: @brillout

brillout: Good

brillout: And thanks for the ping

Issue #1455, opened 2026-07-31
Dashboard UX notes: rail width, Docs/History placement, inline choices, composer Stop/Resume


suleimansh:
A batch of UI/UX notes from live use (2026-08-01), grouped by page.

## Launcher / overview page

* [x] **1.** Right sidebar (tabs area) is too wide: `w-[27rem]` → `w-[22rem]`.
* [x] **2.** Move the **Docs** tab (the plan/todo markdown files at the workspace root) out of the right rail, into the main column under the launcher.
* [x] **3.** Same for **History** — main view, under the launcher.
* [ ] **4.** With several sessions running and some parked on a human: surface **all open questions** in one place. Today each question lives inside its own session view; the Overview badge only counts them.
* [ ] **5.** Show the **active project's tickets** on the launcher page — same presentation as the `/tickets` page, but scoped to the current project.
* [ ] **11.** The tickets list from item 5 should live in the **main column under the launcher** as part of one consistent arrangement with Docs (2) and History (3) — not as today's full inline list, which dominates the page (67 open tickets push everything below the fold).

## Session page

* [x] **6.** Transcript entries that represent an interaction should BE the interaction. The transcript prints each event as a kind chip + text; today interactive things render as inert text there, with the real UI off in a rail tab.

  **Example 1 — `choice`** (today plain text; should be interactive UI, the ●/○ answers hoverable and clickable, so the question is answered inline from the same chat window):

  ```
  (choice)  ? Start the next backlog item? (23 open)
                ● Work on: [Dry-run the Human Queue demo flow and fix glaring breaks](tickets/2026-07-25_nice-demo-human-queue.…
                ○ Stop the backlog loop
  ```

  **Example 2 — `browser`** (today the agent's browser lives in a rail tab; render it live in the flow, at the point the agent used it):

  ```
  (browser)
      ┌──────────────────────────────────────────┐
      │ ◀ ▶ ⟳  http://localhost:3000/…           │
      ├──────────────────────────────────────────┤
      │                                          │
      │        the interactive browser,          │
      │        live where the agent used it      │
      │                                          │
      └──────────────────────────────────────────┘
  ```

* [ ] **7.** Once inline answering exists, consider hiding the Choices panel from the right rail (simpler UI).

## Composer / session controls

* [ ] **8.** **Stop session** should live at the composer (where Start lives), not only in the ⋮ menu — with a stop icon, like Claude Code's UI (the send ↑ becomes a stop while a turn runs).
[...]

PR #1482, opened 2026-08-02
Transcript choice rows become the interaction; rail Choices tab retired (#1455 items 6/7)

suleimansh:
Items 6a + 7 of the #1455 batch, one PR as agreed (item 6 was split: this is 6a — inline `choice` gates; 6b, the inline browser, is design-first on the thread).

## What changed

**Item 6a — a transcript entry that represents an interaction IS the interaction.** A `choice` event row now renders the real thing instead of inert formatter text:

- An **open gate** mounts the same interactive `ChoicePanel` the right rail held (new `inline` container variant — rounded card in the flow, behaviour identical: three shapes, autopilot countdown, parked-after-post). The newest open gate is `active`, keeping Ctrl+Enter bound. Answering happens right where the question was asked.
- A **resolved gate** collapses to a single expandable ✓ line — the `AnsweredChoice` card, extracted from the launcher hub's `AnsweredCard` (#1481) and now shared by both surfaces. The redundant `✓ chose …` log line is folded into the card.
- A gate closed by `end` without an answer stays plain text (#1359 — its audience is gone), as does a superseded earlier firing of a re-fired gate; a resolution is only worn by a firing it follows.
- `EventList` derives all of this from the same events array it renders — no new state, no extra fold in the shell. The read-only relay watch passes no `projectId` and keeps today's text rows.

**Item 7 — with the gates inline, the rail's Choices panel is retired.** The tab, `ChoicesRail`, the fresh-gate focus-yank, and the shell's `pendingChoices` fold are gone. The launcher hub (#1455 item 4) is untouched — it polls server-side and now reuses the shared card.

## Verification

- `pnpm --filter @gemstack/framework-dashboard typecheck` clean; full suite 736 tests / 77 files green (6 new EventList choice-row tests).
- Live-verify next: cheap Sonnet run parked at the backlog gate, answered from the transcript (will report on this PR).

Part of #1455.


suleimansh:
Live-verified on the daemon (:4200) serving this branch's build (checked the served `pages_index` chunk by grep: inline-variant marker present, retired rail-Choices help text absent from every chunk).

End-to-end with a cheap Sonnet run (`2026-08-02T01-28-40-367Z`) parked at the backlog gate:

- The `choice` transcript row rendered the real inline ChoicePanel ("Your call · Start the next backlog item? (23 open)", both options, Recommended marker, Ctrl+Enter bound, autopilot checkbox).
- Right rail showed Files/Docs/History only — no Choices tab, and nothing yanked focus when the gate fired (item 7).
- Clicked "Stop the backlog loop" **in the transcript**: pick posted, run logged "Backlog loop stopped by you (23 item(s) left)", ended clean, handoff skipped (nothing committed — no debris).
- The row collapsed to the ✓ AnsweredChoice line; the redundant "✓ chose" log line is folded into it; Expand shows the options with the pick marked, Collapse returns to one line.

PR #1483, opened 2026-08-02
This session was stopped before it finished, not because the work was do

suleimansh:
This session was stopped before it finished, not because the work was done. Look at what you had already done, then carry on from there. The session lifecycle still applies: once the work is genuinely finished with nothing left to do, call setReadyForMerge() — without it the finished work is never merged.

Opened from The Framework session `2026-08-01T22-12-40-791Z`.


suleimansh:
Resume-after-quota-reset worked end-to-end (this PR is its product: session resumed, found test27/28 already committed, declared ready-for-merge, published via the full ladder). One wrinkle worth knowing about: the branch sat paused across the quota boundary while main landed its own sessions, so both sides had appended to `.the-framework/LOGS.md` and the PR arrived **DIRTY** — the armed merge-on-green cannot resolve a conflict, so it would have stalled silently. I merged main in with a chronological union of the log entries (no entries dropped) to unblock it.

Pattern to watch: any run paused long enough for main to move will hit this on its bookkeeping files. A merge-driver/union strategy for `.the-framework/LOGS.md` (or the daemon rebasing before merge-on-green) would close it.

PR #1491, opened 2026-08-02
Per-run quota guard honors the spend slider when it loosens the gate (fix #1490)

### LINKED ISSUE #1490: Per-run quota guard ignores the Usage slider (autoSpendOffset): runs pause while the bar shows room
## What happened

Dashboard-started Fable runs pause at the first turn boundary with `Quota boundary reached (Current week (Fable))` even after the user drags the Usage panel's limit handle well above the model's weekly usage. Live repro tonight (runs `2026-08-02T21-52-24-189Z`, `21-26-44-426Z`, `21-23-18-530Z`): week ~25% elapsed, Fable week 40% used, slider at ~65% — run still paused.

## Why

The slider writes `preferences.autoSpendOffset` (#960), and the daemon side reads it in ONE place so surfaces can't disagree — `dashboard/quota.ts` (`defaultQuotaSource`):

> One source, so the bar the user reads and the line auto PM obeys cannot disagree (#960): the slider is read here rather than in each consumer.

But the **per-run guard** never got the memo: `consumption-guard.ts:70` calls `quotaBoundaryStatus({ …, limitOffset: DEFAULT_SPEND_OFFSET })` — the hard-coded half-day cushion (≈7.1 points) — so the line a *run* stops at is always `boundary + 7.1`, regardless of the slider. Tonight that line was ~32% < Fable's 40% → pause, while the bar (correctly, per the preference) showed plenty of room. The exact disagreement the #960 comment forbids.

Sonnet runs pass only because the all-models week (22%) sits under the default line — not because anything honors the slider.

## Fix direction

Thread the preference into `startConsumptionGuard`: an injectable `limitOffset` supplier (mirroring `QuotaSource`'s), read from the registry the same way `defaultQuotaSource` does, refreshed per gate check so dragging the slider unblocks a parked run's next boundary check without a restart. `DEFAULT_SPEND_OFFSET` stays the fallback (fresh install, unreadable registry).

suleimansh:
Fixes #1490 — live repro tonight: Fable week 40% used, boundary ~25%, slider dragged to ~65%, and the run still paused with `Quota boundary reached (Current week (Fable))` while the Usage bar showed room.

## What

- `startConsumptionGuard` gains a `limitOffset` supplier. `cli.ts` feeds it `preferences.autoSpendOffset` from the registry — the same single source `defaultQuotaSource` reads for the Usage bar and auto-PM, per #960's "cannot disagree" rule.
- **The slider only ever loosens the gate**: the line in force is `max(DEFAULT_SPEND_OFFSET, slider)`. The guard's existing rule — lowering the slider (to stand down unattended work) must never tighten the gate on work the user asked for — is kept and now has a test pinning it.
- The gate stays synchronous (it answers from cached poller readings): the supplier is refreshed in the background around each between-turns check, and the check uses the last value that landed — one turn behind at worst, same freshness as the quota reading itself. So dragging the slider unblocks a *parked* run's next boundary check without a restart.

## Tests

Three new cases in `consumption-guard.test.ts`: slider loosens (45% window passes under boundary+20), slider never tightens (hard-left slider leaves a within-cushion window ungated), and a mid-run drag is picked up between checks. Suite 1691/0. Spec + changeset updated.

## Verify live

Drag the Usage handle above the Fable weekly "% used" (show-all-limits tooltip), start a trivial Fable run — it should now sail past the turn boundary into the ending (synthesize → backlog gate) instead of pausing.

### DIFF
--- a/.changeset/guard-honors-spend-slider.md
+++ b/.changeset/guard-honors-spend-slider.md
@@ -1,0 +1,5 @@
+---
+'@gemstack/the-framework': patch
+---
+
+The per-run quota guard now honors the Usage panel's spend slider when it loosens the gate (fix #1490): the line a run pauses at is `max(default half-day cushion, autoSpendOffset)`, read fresh around each between-turns check so dragging the slider unblocks a parked run without a restart. Before, the guard hard-coded the default cushion and a run paused on a window the Usage bar showed as having room — the exact bar/gate disagreement #960 forbids. The slider still never tightens the gate on work the user asked for.
--- a/packages/the-framework/src/cli.ts
+++ b/packages/the-framework/src/cli.ts
@@ -66,6 +66,7 @@ import { createGateKeepalive } from './gate-keepalive.js'
 import { nodeGitRunner } from './project.js'
 import { addProject, ensureDaemonToken, listProjects, readDaemonToken, readPreferences, resolveProjectPath } from './registry.js'
 import { startConsumptionGuard } from './consumption-guard.js'
+import { DEFAULT_SPEND_OFFSET } from './preference-defaults.js'
 import {
   planMaintenanceSweep,
   maintainSweep,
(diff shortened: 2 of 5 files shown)


suleimansh:
@brillout 
Will check this issue more..

Issue #778, opened 2026-07-19
Resuming a session fails when its agent conversation is gone


suleimansh:
Split out of #762, which auto-closed when its PR merged. This half is unfixed and is the cause of the FAILED sessions seen while dogfooding.

## What happens

Message a finished session to continue it, and the spawned process dies immediately:

```
driver error: No conversation found with session ID: cdffc594-e95e-4fa5-a072-d64df151431e
[framework] claude-code exited (1): No conversation found with session ID: …
```

The session shows **FAILED** with no explanation on screen; the reason is only in the event log. Observed repeatedly in test-project — four of eight sessions in one stretch.

## Cause

Continuing passes the captured `sessionId` to the agent as `--resume-session`. That id outlives what the agent CLI is willing to resume (its own retention, a cleared history, a different machine), and there is no check before spawning.

## Suggested fix

Do not pre-flight it — there is no reliable way to ask whether a conversation still exists. Instead let it fail once and recover:

- Detect the `No conversation found with session ID` exit in the driver.
- Retry the same prompt **once, without** the resume flag, so the session continues as a fresh conversation rather than dying.
- Say so in the event log ("that conversation is no longer available; continuing without its history"), so the user understands the agent lost prior context instead of silently wondering.

The session row itself should stay one row either way, which #762 already handles.

## Alternative

Surface it as a clear terminal state ("that conversation is gone — start a new session") rather than recovering automatically. Worse, I think: the user's message is already typed and they just want it to go somewhere.

PR #1517, opened 2026-08-05
Fix a batch of bugs across ai-sdk, drivers, autopilot, dashboard, and mcp

brillout:
A repo-wide bug hunt across every package. Each fix is verified against the provider SDK types, the adjacent tests, or the code's own documented intent. The whole workspace stays green after the changes — typecheck clean, and every suite passes (the-framework 1691, ai-sdk 956, ai-autopilot 306, dashboard + mcp + connectors + examples all passing; 21/21 turbo tasks).

## ai-sdk — provider adapters (wrong SDK shapes that were silently dropped or threw)

- **google/chat**: put `systemInstruction` under `config` — the `@google/genai` SDK only reads `config.systemInstruction`, so a top-level sibling was silently dropped and **every Gemini request ran with no system prompt**. Also emit `googleSearch` instead of `google_search` (the snake_case key serialized to an empty tool, so native web search never activated), and map the non-streaming `finishReason` so a truncation/safety stop isn't reported as a clean `stop`.
- **google/embeddings**: use `contents`/`embeddings` — the singular `content`/`embedding` forms threw, and even fixed would have returned empty vectors.
- **google/files**: iterate the returned `Pager` with `for await` — it has no `.files` array, so `list()` always threw.
- **google/images**: send an enumerated Imagen `aspectRatio` (`1:1`/`16:9`/…), not a pixel pair like `1024:1024`, which the predict endpoint rejects.
- **anthropic**: the Files API lives at `client.beta.files.*` (`client.files` is `undefined`) and takes a named uploadable, not a bare `Blob` with a `purpose`; map `max_tokens`/`refusal` stop reasons in both the response and stream mappers so a truncated Claude answer isn't indistinguishable from a complete one.
- **openai**: `files.delete` / `vectorStores.delete` (v4's `del` was removed in v5/v6) and the v5 `vectorStores.files.retrieve(fileId, { vector_store_id })` signature; send PDFs as a `file_data` data URI, not a bare `data`/`mime_type` (rejected by the API).
- **cohere**: v2 `rerank` takes plain-string documents, not `{ text }` objects (which the v2 client rejects before the network call).

## ai-sdk — agent loop

- An `onBeforeToolCall` middleware returning `{ type: 'abort' }` now **stops the whole loop** and pairs the aborted `tool_use` with a result. Previously it halted only the tool phase, so a guard middleware "blocking" a dangerous tool didn't actually stop the agent — the next provider call carried an unanswered `tool_use` (a 400 on Anthropic). Adds a regression test.
- `buildResumeSnapshotMessages` now includes `resumedToolMessages`, so an approval resume doesn't drop the executed tool's result and leave a dangling `tool_use` for the next resume to choke on.
- `fake.assertPrompted` matches the last user message (the actual prompt), not the first (stale history).

## the-framework — drivers

- **actions**: honor the per-prompt abort signal — it was ignored, so an abort could leave the driver polling GitHub for up to an hour.
- **claude-code-quota / agent-cli**: collect stdout/stderr as `Buffer`s and decode once — a per-chunk `String(chunk)` corrupts a multibyte codepoint split across chunks (the quota readout's `·`, and turn error details).
- **claude-code**: on the conversation-gone retry (#778), suppress the failed attempt's `error` event and the retry's duplicate `start`, so a recovered turn's transcript doesn't show a phantom agent error and the user's message twice.
- **cloud**: report a user abort as "prompt aborted", not the misleading "no cloud session was created".

## the-framework — cli / core

- Validate `--deploy` **before** wiring the dashboard / store / control channel, and release every handle (interrupt trap, dashboard, LOGS.md) in `abortBeforeDriver` — a bad `--deploy dokploy` or `--run-on actions` left the process hung with the run stuck `running`.
- `--browser` is gated on a local run: a `--run-on web`/`actions` session can't reach this machine's browser, so the system prompt must not claim one, and no headless Chrome is leaked per remote run.
- A topic run skips the handoff commit (its scratch cwd is not a git repo), so it no longer reports an alarming `commit-failed` after ~600ms of pointless git retries.
- **loopback-host**: match the `127.0.0.0/8` range as an *address*, so a rebound `Host` like `127.evil.com` no longer bypasses the DNS-rebinding guard (fixed in the shared helper and the inline same-origin check). Adds regression assertions.
- **queue-promote**: `landPinnedEntry` retires a no-checkbox open entry too, matching the "open" grammar the rest of the sweep uses — otherwise it stayed open and was re-drained forever. Adds a regression test.
- **file-read**: reject any `.git` path segment, not just a leading one.
- **project**: give `git ls-remote` the network timeout, not the local-write one.

## ai-autopilot

- **overview/markdown**: only skip the top-level `#` title, so a `# ` line inside a body or fenced code block survives the round-trip.
- **loop/verdict**: try the bare-object fallback only when no fenced verdict parses, so brace-shaped prose after the real fence can't outrank it.
- **surface/events**: an `EventStream` iterator woken with nothing left re-waits instead of resolving a spurious `done` on a still-open stream (protocol violation under concurrent `next()`).
- **decisions/markdown**: only a real status counts as the `[status]` prefix, so a hand-written `[Frontend]`-style tag keeps its place in the title instead of being swallowed.

## framework-dashboard

- **use-notifications**: an enable-toggle flip no longer consumes a warmup slot, so enabling a feed mid-session doesn't fire a notification burst for the entire pre-existing backlog.
- **use-run-handoff**: `keepPrevious` across the poll-cadence flip, so the action-bar summary doesn't blink back to the live counts for a beat.
- **use-device-status**: key the poll on tokens too, so a refreshed device token isn't ignored until remount (which showed the device offline and blocked Start against it).
- **format-date**: `formatAge` no longer renders "0y ago" for an age of exactly 364 days.

## mcp

- **node-handler**: split a comma-joined `X-Forwarded-Proto` (a proxy chain otherwise 500s every request), and cancel the body reader on client disconnect so an abandoned SSE session doesn't leak its transport indefinitely.

## the-framework.ai

- **Hero**: the yarn install command no longer pairs Berry-only `yarn dlx` with the removed `yarn global add`.

## Deferred (found, not fixed — flagged for a follow-up)

- **ai-sdk** `prepareStep`'s `tools`/`toolChoice`/`providerOptions` and middleware `onConfig`'s `temperature`/`maxTokens`/`tools`/`providerOptions` are declared but silently ignored by `runFailover`. Real, but the fix threads many fields through the shared streaming + non-streaming failover path and wants its own dedicated tests — left out of this batch to keep it low-risk.
- **ai-sdk** OpenAI `fileSearch()` on chat-completions needs routing through the Responses API (a design change, not a shape fix).
- **ai-sdk** `newMessagesFromTurn` persists an empty `user('')` on a messages-mode continuation; low impact and the helper is documented to mirror the old persistence shape exactly, so it wants a design call rather than a patch.

https://claude.ai/code/session_016iof4vGKX2nrtPk8xPQU8y

---
_Generated by [Claude Code](https://claude.ai/code/session_016iof4vGKX2nrtPk8xPQU8y)_

### DIFF
--- a/packages/ai-autopilot/src/decisions/markdown.ts
+++ b/packages/ai-autopilot/src/decisions/markdown.ts
@@ -41,7 +41,9 @@ export function serializeDecisions(decisions: readonly Decision[]): string {
   return `${HEADER}\n\n${INTRO}\n\n${blocks.join('\n\n')}\n`.replace(/\n{3,}/g, '\n\n')
 }
 
-const HEADING_RE = /^##\s+(?:\[(\w+)\]\s*)?(.+?)\s*$/
+// Only a real status counts as the bracket prefix: a hand-written tag like `## [Frontend] …`
+// keeps its bracket in the title (and takes the default status) instead of being eaten.
+const HEADING_RE = /^##\s+(?:\[(rejected|accepted|superseded)\]\s*)?(.+?)\s*$/i
 const META_RE = /^-\s+([a-z-]+)\s*:\s*(.*)$/i
 
 /** Parse `DECISIONS.md` contents into decisions. Malformed sections are skipped. */
--- a/packages/ai-autopilot/src/loop/verdict.ts
+++ b/packages/ai-autopilot/src/loop/verdict.ts
@@ -42,16 +42,18 @@ export function parseVerdict(text: string): Verdict | undefined {
 
   const candidates: string[] = []
   for (const m of text.matchAll(FENCE)) candidates.push(m[1]!)
-  // Fallback: a bare object somewhere in the text (take the last `{`-to-`}` run).
-  const lastOpen = text.lastIndexOf('{')
-  const lastClose = text.lastIndexOf('}')
-  if (lastOpen !== -1 && lastClose > lastOpen) candidates.push(text.slice(lastOpen, lastClose + 1))
 
-  // Later candidates win, so scan from the end.
+  // Later fences win, so scan from the end.
   for (let i = candidates.length - 1; i >= 0; i--) {
     const verdict = coerce(candidates[i]!)
     if (verdict) return verdict
   }
+  // Fallback — only when no fence parsed: a bare object somewhere in the text (the last
+  // `{`-to-`}` run). Tried after the fences so brace-shaped prose trailing the real fenced
+  // verdict can never outrank it.
+  const lastOpen = text.lastIndexOf('{')
+  const lastClose = text.lastIndexOf('}')
+  if (lastOpen !== -1 && lastClose > lastOpen) return coerce(text.slice(lastOpen, lastClose + 1))
   return undefined
 }
 
--- a/packages/ai-autopilot/src/overview/markdown.ts
+++ b/packages/ai-autopilot/src/overview/markdown.ts
@@ -54,7 +54,10 @@ export function parseOverview(markdown: string): CodeOverview {
       current = { title: heading[1]!.trim(), body: [] }
       continue
     }
-    if (/^#\s+/.test(line)) continue // the top-level `# Code Overview` title
+    // The top-level `# Code Overview` title — only before any real content, so a `# `-prefixed
+    // line inside a body or summary (a shell comment in a fenced block, a hand-written h1)
+    // survives the round-trip instead of being dropped as a second title.
+    if (/^#\s+/.test(line) && !current && sections.length === 0 && summaryLines.every(l => !l.trim())) continue
     if (current) current.body.push(line)
     else summaryLines.push(line)
   }
--- a/packages/ai-autopilot/src/surface/events.ts
+++ b/packages/ai-autopilot/src/surface/events.ts
@@ -74,8 +74,13 @@ export class EventStream<E = SupervisorEvent> {
         return new Promise(resolve => {
           const wake = (): void => {
             pending = undefined
-            if (done || index >= stream.buffer.length) resolve({ value: undefined, done: true })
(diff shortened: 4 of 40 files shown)


suleimansh (approved):
Reviewed the full diff plus the risky call sites in context. Everything holds up:

- **Verified against surrounding code:** `mapGeminiFinishReason` already exists in `google/chat.ts` (the stream path uses it, non-stream now shares it); `usePolled` has the `keepPrevious = false` 5th param the `use-run-handoff` call now passes; `AgentResponse.resumedToolMessages` is populated at `agent.ts:1576`, so `buildResumeSnapshotMessages` prepending it is sound.
- **EventStream re-wait fix:** the re-registered `wake` is the same function reference, so `return()`'s `waiters.indexOf(pending)` cleanup still finds it, and `close()`'s splice-and-wake resolves it done — no leak, no false-done.
- **mw-abort:** pairing the aborted `tool_use` with a result *and* stopping the loop is the right combination — either half alone leaves a 400 or a live agent. Regression test covers both.
- **loopback regex:** correct `127.0.0.0/8` match; note the shorthand `127.1` is now rejected as a Host — strictly a behavior change, but the right side of the guard to err on.
- **cli.ts deploy-validation move:** pure hoist within the same function, no identifier used before its new position; `abortBeforeDriver` now releases interrupt trap, dashboard, and LOGS.md as claimed.
- CI's single `build` check runs install + build + typecheck + full `pnpm test`, so green covers the whole workspace, matching the PR body's claim.

No blocking findings. Merging.

Issue #1496, opened 2026-08-03
CC web doesn't work


brillout:
It's still stuck:

<img width="1366" height="722" alt="Image" src="https://github.com/user-attachments/assets/c2765a18-9ed7-4d4c-acdc-e95539d819c7" />

<img width="1366" height="719" alt="Image" src="https://github.com/user-attachments/assets/4b3860b1-adff-4dda-adf1-bec2e5c8217f" />

Highest-prio since it's a blocker for YC application.

brillout:
Also a blocker for dogfooding TF: I'd like to have agents create a `.plan.md` for all open tickets in parallel.

suleimansh:
Re-test evidence (2026-08-17): this looks fixed by #1518 + #1544 together.

Plainly: a launcher run with "Run on: Claude web" now goes all the way to a merged PR with no human step after send. Live run `2026-08-17T10-52-38-535Z`: the driver pushed a slash-free ref, the session provisioned **repo-bound** (not a bundle — its work commit sits directly on the pushed ref's tip), and instead of parking on effort/model questions it decided alone, committed, and opened PR #1546 itself, 39 seconds after hand-off. Merged; full chain evidence on #1320.

The two halves that closed it:
- #1544 fixed the provisioning side (the screenshots' "stuck" sessions were bundle-provisioned and couldn't push);
- #1518's hands-off closure instruction fixed the behavior side (no questions to nobody — the session lands its work).

Worth one re-test on the parallel-`.plan.md` shape from the comment here (N concurrent web runs — that's #1327's fan-out), but the single-run path this issue shows is working end to end now.

suleimansh:
For the record, since the close came from #1551's merge: the parallel shape this issue still needed is proven. The 2026-08-17 batch locked 3 tickets in one pushed commit (a0b430e6) and fired 3 concurrent "Run on: Claude web" sessions — all three delivered autonomously (PRs #1551/#1552/#1553, 2–4 min after hand-off, clean two-file diffs, no human step, no cross-claims), meeting all four pass criteria in this ticket's merged plan. The full-width 10-agent observation is #1327's scope and runs next.

Issue #1234, opened 2026-07-26
A Claude web run deadlocks on any ambiguous prompt: we tell it to await a human who cannot answer


suleimansh:
Send `hi` to a run with `Run on: Claude web`. The session starts, decides the prompt is ambiguous, asks a question, and stops forever:

```json
{
  "title": "What would you like me to do?",
  "options": [
    { "label": "Work on the next TODO" },
    { "label": "Tell you about current state" },
    { "label": "I'll tell you what to do" }
  ],
  "recommended": "Work on the next TODO"
}
```

Nobody can answer it. The session is spent, nothing was built, and the dashboard cannot show the question.

## Why

Our own prompt tells it to:

> Ambiguous prompt: make a list of interpretations sorted by plausibility, `<SHOW_CHOICES>`, `<AWAIT>`

It obeyed. But a cloud session cannot be answered, and that is structural:

- **Nothing we can use reads it back.** `claude agents --json --all` lists local background agents only, and a `session_01...` id is not among them. A private session API does exist (see #1237 for the endpoints), but it is undocumented and authenticates with the user's subscription credential, so it is not something this fix can lean on.
- **Nothing we can use writes to it.** `--cloud` starts a session; no CLI flag sends another message to one, and `--teleport` is interactive only (silently ignored under `-p`, see #1237).

So every ambiguous prompt sent to this target deadlocks by design, and burns a cloud session for nothing.

## Fix

Tell a hand-off run that the await gates are not available in this session, so it picks the most plausible reading, says which assumption it made, and carries on.

`composeRunSystem` already composes optional blocks around `AWAIT_PROTOCOL` (the browser protocol, the topic bind). Add one keyed off `Driver.handsOff` from #1231. `prompts/system_prompt.md` stays untouched, so the #326 byte guard is unaffected.

Open question: should the same block cover the plan approval and alternatives gates too? They await for the same reason.

## Word it as availability, not as a rule

The better long term shape is capabilities rather than instructions: let the prompt say "show the choices to the user" and let the agent reach for a real capability, instead of spelling out a protocol it has to follow. Availability then becomes a property of the session, and a cloud session simply is not handed that capability. It cannot park on a gate it was never offered.

That is a separate piece of work, but it decides the wording here. Phrase this block as "these gates are not available in this session" rather than "do not await, do X instead". Availability is the true statement either way, and it deletes itself cleanly once choices become a real capability.

Two things that will bite whoever builds the capability version:

[...]

PR #1518, opened 2026-08-05
Cloud hand-off: task-first prompt with labeled separators + tell the session to land its work (#1496, #1497)

### LINKED ISSUE #1497: Improve prompt for Claude Web
- Let's move the user prompt all the way to the top, because that's what the user sees first when he opens the CC web link
- The prompt structure is hard to read for humans, but humans should be able quickly understand what's going on when they read the prompt on CC web
   - The markdown headers of the various injected system prompt conflict, it's quite hard to follow for a human => how about we use a visually very clear separator between injected prompts, e.g. `\n\n\n===============================`

suleimansh:
Refs #1496, fixes #1497. @brillout — this touches a prompt (`prompts/protocols/hands_off.md`), so it wants your read before merge.

## Why

#1496's screenshots show the CC-web hand-off itself working: the session URL is captured, the session runs. What fails is everything after — the session does its triage, leaves the results in the conversation (and in `ANALYSIS_RESULT.md`, which is gitignored), never commits, and the run ends with "handoff skipped: the session committed nothing." Root cause: **nothing ever tells a cloud session to publish.** On a local run the framework publishes at settle; a `--run-on web` run ends at the hand-off and no local machinery can reach the cloud VM's workspace. The session is the only party that *can* land the work, and its instructions never said to.

## What

**1. `hands_off.md` gains the publishing half** (the "don't park on await gates" half already existed from #1234):

- commit on the session branch and open a pull request
- analysis/plan/decision deliverables go into *committed* files — results that live only in the conversation, or in gitignored files like `ANALYSIS_RESULT.md`, reach nobody
- ending without a PR is allowed only when the task genuinely required no repo change, said explicitly

This is what makes "plan hundreds of tickets in parallel" possible: each planning session must come back as a PR carrying `.plan.md` files, not as prose in a tab.

**2. #1497 as asked** — the hand-off prompt is the one prompt a human actually reads (it's the first thing on the claude.ai page), so `cloudHandOffPrompt()` now puts the task first and each injected block behind a hard rule:

```
<task>

===============================

Instructions from The Framework, the tool that started this session:

<session framing>

===============================

<per-call system>
```

Cloud driver only — every streamed driver keeps its invisible system channel exactly as-is; `composeRunSystem` untouched, so #547's byte-identical guarantee for local runs holds.

## Tests

Suite 1715/0. The prompt-order test re-pinned to the new format; new test pins the no-injection case (bare task, no rule/label). Prompt drift check green.

## Not covered here

The screenshot session also surfaced that hands-off runs still route analysis into `ANALYSIS_RESULT.md` per the base system prompt — the protocol now redirects the *deliverable* copy into committed files, but if you'd rather special-case the base prompt for web runs, that's your call (prompts are yours).


suleimansh:
Live-tested on an isolated daemon (`session_014uogAzEHAKxxuJtYBQpFga`): the prompt renders task-first with the labeled rules as intended, the session stated its assumptions instead of parking, picked the smallest unplanned ticket, wrote a real `.plan.md`, committed it on a session branch, and attempted push + PR — i.e. both halves of this PR behaved as designed. The push itself was denied by the cloud sandbox's git proxy, which is #1320 (Claude GitHub app doesn't cover this repo yet) — evidence posted there. So after merge, #1496 reduces to #1320's one-time settings click.

packages/the-framework/prompts/protocols/hands_off.md
```diff
@@ -8,3 +8,13 @@ AWAIT, that capability is not available here. Do not emit an await block and do
 - state in one line which assumption you made
 - carry the work through to the end
 The non-blocking blocks (show-markdown, set-session-name, ready-for-merge) are unaffected.
+
+## Land your work before the session ends
+The machine that started this run cannot publish for you either: nothing local sees this
+session's workspace, so work you do not land yourself does not exist to anyone. Before ending:
+- commit your work on your session branch and open a pull request for it
+- if the deliverable is analysis, a plan, or a decision, write it into committed files — a
+  result that lives only in this conversation, or in a gitignored file such as
```
brillout:
```suggestion
  result that lives only in this conversation, or in a gitignored file
```

packages/the-framework/prompts/protocols/hands_off.md
```diff
@@ -8,3 +8,13 @@ AWAIT, that capability is not available here. Do not emit an await block and do
...
 - carry the work through to the end
 The non-blocking blocks (show-markdown, set-session-name, ready-for-merge) are unaffected.
+
+## Land your work before the session ends
+The machine that started this run cannot publish for you either: nothing local sees this
+session's workspace, so work you do not land yourself does not exist to anyone. Before ending:
+- commit your work on your session branch and open a pull request for it
+- if the deliverable is analysis, a plan, or a decision, write it into committed files — a
+  result that lives only in this conversation, or in a gitignored file
+  ANALYSIS_RESULT.md, reaches nobody
```
brillout:
```suggestion
  reaches nobody
```

packages/the-framework/prompts/protocols/hands_off.md
```diff
@@ -8,3 +8,13 @@ AWAIT, that capability is not available here. Do not emit an await block and do
 - state in one line which assumption you made
 - carry the work through to the end
 The non-blocking blocks (show-markdown, set-session-name, ready-for-merge) are unaffected.
+
+## Land your work before the session ends
+The machine that started this run cannot publish for you either: nothing local sees this
+session's workspace, so work you do not land yourself does not exist to anyone. Before ending:
+- commit your work on your session branch and open a pull request for it
+- if the deliverable is analysis, a plan, or a decision, write it into committed files — a
+  result that lives only in this conversation, or in a gitignored file
```
brillout:
```suggestion
  result that lives only in this conversation, or in a gitignored file,
```

brillout:
Isn't it surprising that `## Land your work before the session ends` lives at packages/the-framework/prompts/protocols/hands_off.md ?

suleimansh:
Both sections follow from the same fact: nothing is attached to a hands-off session — no one to answer a gate, no local machine to publish the workspace. Can split out a `land_work.md` if you'd rather one rule per file.

brillout:
It would probably be cleaner (more DRY) to have a single "closure" instruction?

suleimansh:
Folded into a single closure instruction in 42982362: the detachment premise is stated once, with the two consequences (decide without awaiting, land your work) as bullets under it. The heading question dissolves with it.

suleimansh:
@brillout folded as you suggested in 42982362 — one closure instruction, premise stated once. Good to merge?

brillout:
Ok

Issue #682, opened 2026-07-18
Move tickets/TODO.md to TODO-AGENTS.md


brillout:
So that tickets/* only holds tickets.

Issue #410, opened 2026-07-12
`~/$HOME/.the-framework.json` instead of localhost


brillout:
I'm inclined to save preferences at `~/$HOME/.the-framework.json` instead of using localStorage — so, AFAICT, we actually never need localStorage.

Suggestion:
- `~/$HOME/.the-framework.json` => user preferences
- `~/$HOME/.config/the-framework/*` => internal files (e.g. global cache)

Where `~/$HOME/.config` is an example for Linux — on MacOs and Windows, the directory for config files lives somewhere else.

suleimansh:
Agree on dropping localStorage. We are already most of the way there: the registry resolves `$XDG_CONFIG_HOME/the-framework.json`, falling back to `~/.the-framework.json` (registry.ts). I would just put the global-options prefs in that same file next to the project list. One user-owned file, re-created per machine, no localStorage.

Two notes:
- Do not hardcode `~/.config` for the internal-files bucket. That is Linux-only, and cache is not config. Config/prefs -> XDG_CONFIG_HOME (mac `~/Library/Preferences`, win `%APPDATA%`), cache -> XDG_CACHE_HOME, data -> XDG_DATA_HOME. Simplest is to lean on `env-paths` and stop hand-rolling paths: keep the single top-level file for the user-owned stuff (registry + prefs), and use the cache/data dirs only for derived things (daemon.json liveness, global caches).
- With the Vike SPA rebuild (#405) the browser cannot touch the filesystem. So "never need localStorage" holds only because the daemon owns the file and the SPA reads/writes it over a Telefunc RPC. Fine, just noting prefs become daemon-owned, not browser-owned.

Net: one user file for registry + prefs (what we have), `env-paths` for the cache/data dirs, localStorage gone.

PR #1543, opened 2026-08-17
Zero migration code — and the one removal that says so out loud

brillout:
Two asks from #1536, and they are the same decision seen from both ends: nothing reads what a rename left behind, and the one deletion whose direction is *permissive* says so instead of being silent.

Based on #1536's head (`bf1b0f7`) and targeted at its branch, so it can land before the merge.

## 1. Zero migration code (`10f280c`)

Per [AGENTS.md](https://github.com/gemstack-land/the-framework/blob/main/AGENTS.md#L1-L2): the project has no users, so a rename is a break to take rather than a compatibility branch to carry. Every back-read on that branch — and the two that predate it — goes, along with the code that existed only to serve them. **+137 / −344.**

- **`run.json`.** `readMetaIn` is gone; every live/worktree read goes straight to `agent.json`, exactly as it did before the back-read.
- **The archive directories' old names.** `sessions/` and `runs/` are no longer read by `archiveDirs`, no longer un-ignored by the seeded `.the-framework/.gitignore`, and no longer staged by the committer. That last one takes the per-commit pathspec filter with it: two specs needed filtering because git fails the whole command when one matches nothing, and with a single spec there is nothing to decide.
- **The handoff triple.** `autoPushBranch` / `autoOpenPr` / `autoMerge` are unknown keys now — in the committed `the-framework.yml` and in the stored preferences alike.
- **`antiLazyPill` → `vanilla`** (and its inversion), and the stored **`agent` → `driver`**.
- **The registry's bare `ProjectRecord[]`** (pre-#410). `readRegistry` accepts one shape; anything else is the empty registry it already answered for a malformed file.
- **Two `@deprecated` aliases** kept for callers that no longer exist: `VersionProbe` (also dropped from the public barrel) and `RunClaudeOptions`.
- **A `mkdir`** whose only reason was a `tickets/TODO.md` the backlog stopped reading when the pre-#682 spellings went.

The tests that pinned each migration now pin its absence — the house style for a compat cut, the same shape `findFlatTodo` already has for the retired backlog locations. Five SPECs carry the decision where SDD puts it, and `MEMORY.md` carries it once for the repo so the next rename does not grow a reader.

## 2. The per-project tier says it is gone (`e0ba45f`)

Every other retirement above is a rename, and a renamed key is simply unknown. B5's deleted per-project tier is not a rename, and its direction is the one this review kept turning up: those repos fall through to the user's default, and the default publishes. A repo set to "don't publish" starts pushing branches and opening pull requests under the user's name.

So it gets a notice — a report, not a migration. Nothing is honoured, carried, or written back; the block goes on being ignored, and the notice names it.

- **Read at boot, first statement in `runDaemon`.** `writeRegistry` rewrites the file from the fields it knows, so the daemon token, the project registration or any preference write erases the block. A notice raised at the first publish would race that erasure and lose.
- **Each row names the repo, what its block held, and the exact `handoff: <rung>` line** for that repo's `the-framework.yml` — with what happens without it. A block that never set publishing is told the user's own settings apply instead.
- **A pre-ladder block is recognized without being interpreted.** It spells publishing as the three booleans B5 replaced, and turning those back into a rung needs the mapping commit 1 deleted. The row says to restate it and offers the ladder's own words. Reassuring that reader is the failure the notice exists to prevent.
- **Both surfaces, one wording.** The daemon's log prints it; the dashboard is handed each row already written, since what a removal did to a repo is not a judgement two surfaces should each phrase for themselves. It reaches the dashboard because the premise is that nobody is at the keyboard — which makes stdout at boot the one place it would not be read.

Nothing is withheld pending an acknowledgement (that is the deleted tier under another name) and nothing is written into the user's repo on their behalf (a bigger outward action than the one being prevented). The banner clears itself: the block is gone after the first preference write, so the next boot has nothing to say.

## Consequences, stated out loud

An existing checkout is now a checkout of a different product. A `.the-framework/` written before D5 reads as empty — no live agent, no history — rather than being healed. That is the trade being asked for, and it costs nothing while nobody has one. The per-project tier is the single exception, and it is an exception about being told, not about being kept.

## Verification

Both typechecks, a full build, **1432 node tests**, **776 dashboard tests (81 files)**.


suleimansh:
Checked against the code and ran it: both typechecks, a clean build, and **1432 node / 776 dashboard, zero failures** on `e0ba45fa` (your 1422/772 predates the notice commit — its tests account for the difference).

The removals are symmetric and complete: every read site follows its constant, the two non-migration fixes are untouched (the relay/browser-proxy origin guard, and the stop-decline publish leak — only the legacy `mkdir` left `todo-loop.ts`), and the committer's single-pathspec shape plus the halved `archiveDirs` probes are real wins beyond the deletions. Your notice also covers what mine didn't: the pre-ladder boolean publish spellings, repos named by path, and one advice string shared by the boot log and the banner, so the two surfaces can't drift.

Two notes, neither blocking:

1. **The banner is lost on restart-after-first-write even if never seen.** Boot captures, a preference write erases the block, and the next boot has nothing to read — so a user who didn't have the dashboard open during that first session never learns. Your comment owns this ("the banner goes when the fact does"), and a preference write almost always means the dashboard was open, so the window is narrow. Stating it so the trade is on record.

2. **The yml parser now silently ignores unknown keys**, so a committed pre-B5 `autoPushBranch: false` would silently start publishing — the one direction this branch kept naming — with no notice, since the banner covers only the registry tier. I grepped the org: no yml anywhere carries the old spelling, so it's theoretical today. If you want the manual-migration story to be discoverable without migration code, a loud unknown-key warning in `parseFrameworkConfig` would do it; equally fine to leave.

Good to merge from my side.

brillout:
Thanks — CI came back green on `e0ba45f` here too. Neither note pushed, and here is why for each.

**1. The unseen banner.** Agreed, and there is no version of this that closes it without keeping what was deleted. Remembering "this was shown" is a write, and remembering "this was pending" means carrying the block past the write that drops it — which is the tier's data living on under another name. The window is the one you describe: a preference write with no dashboard open, which is close to a contradiction since preference writes come from the dashboard. Worth having on record rather than fixed.

**2. The yml's unknown keys.** You are right that it is the same direction, and it is the last silent-permissive path left on the branch. One thing I would want settled before writing it: **it has to warn, not throw.** `loadFrameworkConfig` turns a throw into a warning *and an empty config* — so failing on an unknown key would drop the file's `handoff` along with it and publish more than the file says, which is the failure we would be trying to name. That means threading an `onWarn` through `parseFrameworkConfig` (it has none today; only `loadFrameworkConfig` does), then deciding where that warning surfaces for an unattended agent, since stdout at boot has the same nobody-is-reading problem the banner exists for.

That is a real change to the parser and its callers, on a branch that is otherwise at "good to merge" and for a key that — as you checked — no yml in the org carries. So I have left it, deliberately. It is also independent of both commits here: it names nothing retired, it warns about any unknown key, so it stands or falls on its own merits and would land just as well as a follow-up.

Happy to write it now if you would rather it went in with this — say the word and it is a small commit on top. Otherwise this is ready as it stands.

---

Issue #697, opened 2026-07-18
`tickets/` view


brillout:
Also:
 - A button "import tickets from GitHub" when empty => it's just a small prompt
 - Button "Add to Queue (TODO_AGENTS.md)" for each ticket

PR #1549, opened 2026-08-17
One GitHub sync: the import preset goes, the update's empty branch is the first import (fix #1501)

### LINKED ISSUE #1501: Remove prompt to import GH issues
Let's remove it in favor of the following udpate prompt that also covers when `tickets/` is empty.

https://github.com/gemstack-land/gemstack/blob/dbf2d445ec2d7083e5056bc08289351f566cc7e5/packages/the-framework/prompts/presets/update_tickets.md

suleimansh:
Builds #1501 as decided there: the `import-tickets` preset is removed in favor of `update_tickets`, whose `[Empty]` branch already treats a bare `tickets/` as a first import (every open issue comes across).

## What changes

- **`import_tickets.md` deleted**, `importTickets` preset removed from the catalog and launcher list. `update-tickets` is now the only `newAgent`-marked preset.
- **Every surface that offered the import now offers the update**, under the preset's own label, sending its text verbatim — the empty backlog panel (project + all-tickets pages), the onboarding checklist's "Populate tickets/" step, and the composer's preset menu. One label, one instruction, wherever pressed (#697's rule, now with one fewer prompt to keep honest).
- The update preset's tooltip now says an empty `tickets/` gets a full first import.
- Tests moved with it: the catalog's pinned-names list, the newAgent pinning, and the prompt-content test now assert the `[Empty]` first-import branch; panel/page/onboarding/composer tests target the one button.

## Notes for review

@brillout — prompt surface, so flagging per the usual: no prompt text changed, only `import_tickets.md` deleted; `update_tickets.md` is untouched and carries the whole load, exactly the shape you linked in the issue.

One wording call worth a look: the empty-state and onboarding buttons now say **"Update from GitHub"** even on a never-filled `tickets/`. That keeps label = preset label everywhere; if "update" reads oddly there, relabeling is a one-liner (at the cost of two labels for one preset).

Suite: 1428 node + 774 dashboard, 0 fail.

### DIFF
--- a/.changeset/one-github-sync-preset.md
+++ b/.changeset/one-github-sync-preset.md
@@ -1,0 +1,5 @@
+---
+'@gemstack/the-framework': minor
+---
+
+The separate "Import tickets from GitHub" preset is gone; "Update from GitHub" is the one GitHub sync. The update prompt already treats an empty `tickets/` as its first import — every open issue comes across — so the import preset was a second button for the same work, and two entry points meant two prompts to keep honest. Every surface that offered the import (the empty backlog panel, the onboarding checklist) now offers the update under its own label, sending the same preset text verbatim wherever it is pressed.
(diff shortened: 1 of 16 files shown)


suleimansh:
Merging: this implements exactly what #1501 asked for ("remove it in favor of the update prompt"), and today's dogfood run validated the replacement path live — the **Update from GitHub** routine handled the import-from-empty case itself and opened #1574 (merged). @brillout FYI — happy to revisit the wording if you want changes post-merge.

Issue #1580, opened 2026-08-18
Save all branches at `.the-framework/branches/`


brillout:
It's gitignored and includes all active worktrees generated by TF-driven agents.

Also add `branches/` (repo root) as a symlink to `.the-framework/branches/`

highest-prio because it's a structural improvement.

Issue #1575, opened 2026-08-18
Cloud web runs execute the published npm build, so their bookkeeping predates repo-side renames


suleimansh:
Caught live by the first post-#1573 dogfood sweep: the \"Update from GitHub\" routine ran as a cloud web session (\`target: web\`), and its PR #1574 failed CI because the session archive landed in \`.the-framework/<user>/sessions/\` — the directory #1536 renamed to \`agents/\` — tripping main's zero-migration guard test.

Cause: the cloud environment installs \`@gemstack/the-framework\` from npm, and the last published version predates the rename. Local daemons run the repo build (v0.0.0 self-report), so the skew only shows on web runs.

Options:
- Publish a release (merge the standing \`chore: version packages\` PR #1285) whenever repo-side layout changes land — makes the skew window short but reintroduces it on every such change.
- Pin cloud runs to the repo's own build rather than npm (heavier, but removes the class of bug).
- At minimum: the version gate that #1519 added for the extension has no counterpart for the cloud environment's framework version.

Workaround applied to #1574: a fixup commit re-homing the two archive files under \`agents/\`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1547, opened 2026-08-17
Cloud hand-off leaves scratch refs on origin — sweep them


suleimansh:
Since #1544, every "Run on: Claude web" run leaves refs on origin that nothing consumes afterward.

What the first live run (#1546, session_01VTAkK8iU4FhDr5EkQPbTV3) left behind, both at the same commit 5ef9f285:

- `cloud-1-3955352b` — the slash-free ref the driver pushes pre-hand-off so the session can clone at it (anthropics/claude-code#87235 workaround)
- `the-framework/agent-2026-08-17T10-52-38-535Z` — the run branch

The session does its work on its own `claude/*` branch and opens its PR from there, so once provisioning settles the scratch refs are dead weight — one pair per web run, accumulating forever.

Sweep options:

1. Driver deletes the ref once session creation settles. Mind the provisioning race: the clone happens at an unknown time after hand-off — deleting too early strands the session (it would need `claude --teleport` recovery). There is no signal for "clone finished", only "session created".
2. Daemon sweep: delete `cloud-*` refs (and run-branch refs whose run is ended with no open PR) older than some safe age, e.g. a day.

Option 2 is safer given the race. Low priority — the refs are tiny — but the list grows with every web run.

PR #1588, opened 2026-08-18
Mint slash-free branch names: tf-agent-<id> and tf-<slug>

### LINKED ISSUE #1581: Avoid `/` (and other FS problematic chars) in branch names
So that `branches/` is flat and each dir matches the name of its branch.

See:
- https://github.com/gemstack-land/the-framework/issues/1580

suleimansh:
Closes #1581.

## Why
A `/` in a branch name breaks two consumers:
- the flat `branches/` dir planned by #1580 — the dir can no longer be named exactly like its branch;
- cloud sessions: a slashed ref never resolves as the session revision (anthropics/claude-code#87235), which is the whole reason the `cloud-*` scratch-ref push in `cloud-scratch-refs.ts` exists.

Agent slugs are already `[a-z0-9-]`-only, so the `/` from the two prefixes was the only FS-problematic character we ever emitted.

## What
- New `src/branch-names.ts` (pure, node-free — `auto-pm.ts` is reachable from the client barrel) is now the single mint site: run branches start as `tf-agent-<id>` and rename to `tf-<slug>`; the triage pins are `tf-triage-quick` / `tf-triage-consensual`.
- **Prompts** (@brillout — your call per the prompt-ownership rule, happy to adjust): `system_prompt.md`, `triage_quick.md`, `triage_consensual.md`, `signal.md` now tell the agent to create `tf-<SESSION_NAME>` instead of `the-framework/<SESSION_NAME>`. Naming scheme is a two-constant change if you prefer another spelling (asked on the issue).
- Legacy `the-framework/…` spellings are still **recognized, never minted**: the scratch sweep ages out legacy run branches left on origin (test added), `isAgentBranch` classifies both, and the archive branch fallbacks keep guessing legacy names since every run they apply to predates the rename.
- Repo hygiene rider (first commit): dropped the stale `!*/sessions/` un-ignore rules from this repo's committed `.the-framework/.gitignore` and untracked the four session files they let back in today (same intent as 1785a01c, which only fixed the seeded file).

## Follow-ups (not in this PR)
- With run branches slash-free, the `cloud-*` scratch-ref push + daily sweep (#1547) become removable — the cloud hand-off can clone at the run branch directly.
- Old-era `the-framework/run-<ts>` branches (minted by the outdated npm daemon build, #1575) are a third legacy spelling nothing sweeps; worth folding into the scratch-ref follow-up.

## Tests
Full suite green: 1463 node tests + 773 dashboard tests, 0 failures.


brillout:
Prompt changes LGTM

PR #1590, opened 2026-08-18
CLAUDE.md imports AGENTS.md so Claude Code sessions load the rules

suleimansh:
One line. Claude Code injects `CLAUDE.md` into every session's context; this repo only had `AGENTS.md`, so its rules were literally absent from those sessions — which is how #1589 grew compat bloat despite the zero-users rule. The `@AGENTS.md` import makes AGENTS.md the single source of truth and puts it in context everywhere.

(TF-driven runs get the #326 system prompt instead, which also doesn't carry AGENTS.md — that half is yours, flagged in the AC issue.)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,1 @@
+@AGENTS.md


suleimansh:
@brillout verified empirically before asking you to look: in a test dir with a canary rule in `AGENTS.md`, a fresh `claude -p` session answers **NO** when asked if the rule is in its project instructions (AGENTS.md alone is never loaded — exactly how #1589 grew compat bloat), and **YES, quoting it verbatim** once this one-line `CLAUDE.md` (`@AGENTS.md`) sits beside it. So this makes AGENTS.md actually reach every Claude Code session, with AGENTS.md staying the single source of truth. The TF-run half (injecting it into the #326 prompt) is tracked in #1591.

brillout:
Great idea :100:

brillout:
(FYI I'd go for `Read ./AGENTS.md`, so it's more readable for humans. But just a slight preference, you can ignore, just FYI.)

suleimansh:
Kept the `@` import deliberately: it's mechanical context injection (verified in the A/B above), while a `Read ./AGENTS.md` sentence depends on the model choosing to obey it — the same context-rot failure mode we're closing. The `@path` syntax is Claude Code's documented import form, so it reads as "include this file" once you know the convention.

Issue #1397, opened 2026-07-30
Daemon commits bookkeeping to local main but never pushes it


suleimansh:
The daemon commits session records (`.the-framework/sessions/*`) to the local main checkout but never pushes main. Local main drifts ahead of origin until a human pushes by hand.

Two bites so far:
- 2026-07-30 ~03:15: local main was 15 commits ahead; a run's PR (#1380) carried 29 stale session files because run worktrees fork the local checkout's HEAD.
- 2026-07-30 night: 2 unpushed bookkeeping commits again, pushed manually.

Fix direction: either push main after the bookkeeping commit (best-effort, like the spike-lock push), or stop committing bookkeeping to main and keep it out of run baselines.

suleimansh:
🤖 *agent comment*

**A third bite, and this one leaves permanent litter on `origin`.**

Found by looking at the two yellow "had recent pushes" banners on the Pull requests page. `tf-data` is just the data branch and is noise. The other was `cloud-1-ee53dc6e` — which led here.

`origin` currently carries five `cloud-*` scratch refs. Three of them can never be swept, and this issue is why:

| ref | tip commit | tip holds | parent on `main`? |
|---|---|---|---|
| `cloud-1-46418f8c` (Aug 17) | `[The Framework] 3 sessions` | 7 changed lines | yes |
| `cloud-1-97fb88cb` (Aug 17) | `[The Framework] 10 sessions` | 21 changed lines | yes |
| `cloud-1-a0500167` (Aug 18) | `[The Framework] a session` | 3 changed lines | yes |
| `cloud-1-dd8ca393` (Aug 21) | `web hand-off …` | nothing (empty) | **no** |
| `cloud-1-ee53dc6e` (Aug 21) | `web hand-off …` | nothing (empty) | yes |

The first three are the bookkeeping commits this issue is about — `.the-framework/sessions/*` committed to local `main` and never pushed. A cloud hand-off pushes the local checkout's HEAD to its scratch ref, so those unpushed commits **do** reach `origin`, riding a ref that nothing ever merges.

Then #1547's sweep does exactly the right thing and keeps them forever: its work gate refuses any ref whose tip is not reachable from the default branch, and #1601's empty-anchor exemption does not apply because these tips are not empty — they carry real session records. So they are kept as `holds-work`, correctly, and permanently.

[...]

PR #1595, opened 2026-08-18
All framework data moves to the tf-data branch (fix #1582)

### LINKED ISSUE #1577: Queue promote can clobber a newer main: an agent's stale TODO_AGENTS.md copy lands over externally-merged entries
Caught live in the 2026-08-18 dogfood run. Sequence:

1. The quick-win triage ran as a cloud web session and delivered its queue additions via PR #1576, which merged into main (three effort-1 entries in `TODO_AGENTS.md`).
2. The same routine's local run row still had its own branch copy of `TODO_AGENTS.md` — forked *before* those entries existed (empty sections).
3. The next sweep promoted that run (`landed the queue from 1 run(s)`) and committed the agent's stale, empty copy over the populated file (`[The Framework] queue updates from 2026-08-18T14-21-50-687Z`, −6 lines) — silently deleting all three just-queued entries from the checkout.

The promote design (#852) assumes the agent's branch copy is the newest truth because the agent is the only queue writer. Web runs broke that assumption: their queue edits arrive via merged PRs on origin/main, so the local promote of the same logical run is a second, older writer.

Possible directions:
- Promote by merge rather than copy (3-way merge the agent's copy against the fork point, so external additions survive).
- Skip the promote when the agent's copy is identical to its fork-point version (nothing was taken off → nothing to land).
- Treat a run whose changes landed via PR as already-promoted.

Recovered manually (dropped the clobber commit). Related: #1575 (web runs execute the published npm build).

### LINKED ISSUE #1582: Save data in special branch `the-framework_data`

suleimansh:
Implements #1582: **main is 100% code, 0% framework data.** Everything The Framework writes — `tickets/**` (plans and locks included), `TODO_AGENTS.md`, the session archives — moves to one dedicated branch, `the-framework_data`, checked out at `.the-framework/branches/the-framework_data` with a root `tickets` symlink into it.

**Net −1,300 LoC.** Three mechanisms stop needing to exist:

- `queue-promote.ts` — one queue, one branch, one local writer (the daemon). A drained entry is checked off by the daemon at settle, once the run's ending says the work was published. The #1577 clobber can't happen anymore.
- `agent-commit.ts` — archives are committed *and pushed* the moment a session settles, so the debounced committer and its flush phase go. #1397's stranded bookkeeping dies with it.
- ticket-locks' own commit/push machinery — a lock batch is just another funneled data write.

**How writes work** (`data-branch.ts`): every local write goes through one serialized cycle — sync with origin, apply, commit, push. A lost push race re-syncs and re-applies; a failed push keeps the commit local and the next cycle carries it out. The daemon also pulls the branch eagerly (~1 min), so machines and cloud sessions converge. Reads work from anywhere: the checkout, `git show the-framework_data:…`, or `origin/…` in a cloud clone.

**Prompts** (@brillout — yours to review, see [my comment](https://github.com/gemstack-land/the-framework/pull/1595#issuecomment-5336172195)): one new "The data branch" section on the system channel; all data-touching presets point at it. Two deliberate changes: drain agents no longer check their entry off (the daemon retires it), and **a plan is now a data write, not a PR** — plan review moves to the human queue.

**Tested:**
- Suites: 1455 node + 771 dashboard, both tsconfigs clean, CI green.
- Manual end-to-end smoke on a scratch repo with the daemon on this build: first tick creates branch + checkout + symlinks + seeded queue; a dashboard queue-add lands as a pushed data commit; two real drain runs followed the new write protocol unprompted; check-off and archives landed as data commits; a run whose hand-off failed correctly left its entry open.
- The smoke caught one real bug, fixed here: retiring the last ticket removes `tickets/` itself, and the lock write didn't create parents — every later drain claim failed silently and the sweep stood down with a wrong message. Fix + regression tests in `5e3785b4`; the whole retire loop is now also covered by three integration tests (`1fa12a6b`).
- Synced with main past #1594 (one test needed to learn to wait for the archive commit — the archive is a git cycle now, not a file copy).

**Hand-run migration** (zero migration code, per the no-compat rule) — after merge, on the main checkout:
1. Let the daemon birth the branch (or `git checkout --orphan the-framework_data && git rm -rf . && git commit --allow-empty -m "[The Framework] the-framework_data"`), then work in `.the-framework/branches/the-framework_data`.
2. Copy `tickets/**` and `TODO_AGENTS.md` into the data checkout; move `.the-framework/<user>/agents/**` to `agents/<user>/**`; commit + push.
3. One commit on main: delete `tickets/`, `TODO_AGENTS.md`, and the committed archives; add the `tickets` symlink and the simplified `.the-framework/.gitignore`.
4. Restart the daemon; the reconcile creates the checkout + symlink on every other machine.

Closes #1582. Structurally closes #1577 and #1397.


suleimansh:
@brillout prompt changes in `4052c45c` want your look — one new "The data branch" section (tickets/queue/archives on `the-framework_data`, read via `git show`, write via detached worktree + direct push), all data-touching presets point at it. Drain agents no longer check their entry off — the daemon retires it at settle. Big call: **a plan is now a data write, not a PR** (plan review moves to the human queue).

Manually tested end to end on a scratch repo: dashboard queue-add, two real drain runs following the new protocol unprompted, check-off + archives landing as data commits. Objections to plans-as-data-writes, or wording tweaks?

brillout:
The prompt changes are messy, not DRY enough, and not succinct enough. WDYT? Ideas?

FEATURES-SPEC.md
```diff
@@ -123,7 +123,7 @@ happens while nobody is at the keyboard.
 | 82 | Empty agents publish nothing |
 | 83 | Handoff panel: push / open PR / merge, as buttons |
 | 84 | A withheld merge is reported with its reason |
-| 85 | Agent history archived in the repo under per-user directories |
+| 85 | Agent history archived on the `the-framework_data` branch under per-user directories — pushed the moment a session settles |
```
brillout: I thought we now prefix with `tf-`?

suleimansh: Right, missed #1581's convention — renaming to `tf-data`. All occurrences are new in this PR, so it's a pure rename.

suleimansh:
Agreed — fix is deletion. The system prompt already says it twice (the tickets/TODO context bullets + the "The data branch" section), so all ten per-preset "(on the data branch — see …)" parentheticals go. Only behavior survives: drain's check-off step and update_tickets' one-commit rule. Prompt diff shrinks to the protocol file + two context-bullet tweaks.

packages/the-framework/prompts/data_branch_protocol.md
```diff
@@ -0,0 +1,24 @@
...
+
+1. `$ git fetch origin tf-data` (skip without a remote)
+2. `$ git worktree add --detach .tf-data tf-data` (or `origin/tf-data`)
+3. Edit the files inside `.tf-data/`
+4. `$ git -C .tf-data add -A && git -C .tf-data commit -m "<what you changed>"`
+5. `$ git -C .tf-data push origin HEAD:tf-data` — if the push is rejected, `$ git -C .tf-data pull --rebase origin tf-data`, then push again
+6. `$ git worktree remove --force .tf-data`
+
+Rules:
+- Never commit these files to your session branch, and never mix a data commit with code commits — a data change is pushed directly, it does not ride your PR
```
brillout: Is that true? I think there are use cases when a PR also modifies data?

brillout: Actually, because a PR always targets a non-data branch, it kinda must be true, I think?

packages/the-framework/prompts/data_branch_protocol.md
```diff
@@ -0,0 +1,24 @@
...
+- With no remote, push is skipped: the commit on the local `tf-data` branch is the write
```
brillout: That's a weird exception, isn't it?

packages/the-framework/prompts/data_branch_protocol.md
```diff
@@ -0,0 +1,24 @@
...
+
+- `$ git fetch origin tf-data` first (skip when the repo has no remote)
+- `$ git show tf-data:TODO_AGENTS.md` / `$ git show tf-data:tickets/<FILE>` — use `origin/tf-data:` instead when the local branch doesn't exist
+- List tickets: `$ git ls-tree --name-only tf-data tickets/`
+
+## Writing
+
+Only when the instructions tell you to edit tickets or the queue:
+
+1. `$ git fetch origin tf-data` (skip without a remote)
```
brillout: It feels like too much like micro management? Is it needed? Agents are quite proficient with git, so not sure it's needed? (I'd be curious how Claude Code Web handles git operations, but I'm inclined to think it doesn't micro manage agents.) WDYT?

packages/the-framework/prompts/data_branch_protocol.md
suleimansh: Yes, must be true — main carries zero data files now, so a PR with a ticket edit would re-land data on main at merge. The rule is there because committing tickets next to code is the old habit this branch breaks.

packages/the-framework/prompts/data_branch_protocol.md
suleimansh: Agreed, it only reads weird — the write *is* the commit; push is just propagation, and no remote means nothing to propagate to. Folding it into the write sentence.

packages/the-framework/prompts/data_branch_protocol.md
suleimansh: Agree — dropping the command recipe, keeping just the contract: commit on `tf-data`, push immediately, rebase on rejection, never via your checkout or PR. Claude Code's own prompts don't micromanage git either. The recipe did pass the E2E drain runs, so if trimmed agents fumble, the cheap fallback is one worktree hint.

brillout:
Brainstorming:
1. We mention the full path `.the-framework/branches/tf-data/TODO_AGENTS.md`
2. TF automatically rebases & pushes data

I think we don't need `1.` and explaining where TODO_AGENTS.md lives is enough.

Not sure about `2.`. For logs it seems clearly that eagerly pushing data is the way to go. Pushing ticket locks as well, but then we need to handle the situation there is a conflict — in that case I think it's better the agent tries to push and fails?

packages/the-framework/prompts/data_branch_protocol.md
brillout: > old habit 

There shouldn't be any old habit left, since we should remove all old code. Zero compat code, as always.

packages/the-framework/prompts/data_branch_protocol.md
brillout: Why not ignoring this egde case? (AFAICT it would almost never happen.)

If anything such edge case is an error state that should be surface to the UI, right?

Swallowing is actually the worst way to handle this erroneous state?

suleimansh:
> Brainstorming:
> 
> 1. We mention the full path `.the-framework/branches/tf-data/TODO_AGENTS.md`
> 2. TF automatically rebases & pushes data
> 
> I think we don't need `1.` and explaining where TODO_AGENTS.md lives is enough.
> 
> Not sure about `2.`. For logs it seems clearly that eagerly pushing data is the way to go. Pushing ticket locks as well, but then we need to handle the situation there is a conflict — in that case I think it's better the agent tries to push and fails?

1: agree, the trimmed protocol has no paths anymore.
2: the daemon already does this for its own writes. For agents push-and-fail is right, on locks the conflict IS the claim mechanism, loser backs off. And cloud runs have no daemon, so agents must push themselves.

Will test this and report back.

packages/the-framework/prompts/data_branch_protocol.md
suleimansh: Bad wording from me, there is no compat code — "habit" = the model's default to commit tickets next to code. The rule guards against that.
Tested e2e on a scratch repo: drain agent with the trimmed prompt did clean data commits, nothing leaked on its code branch.

packages/the-framework/prompts/data_branch_protocol.md
suleimansh: Agree, dropping it, protocol just says commit + push, no special case. 
No remote becomes an error surfaced to the UI. 
Removing the clause + small ticket for the UI part.

suleimansh:
Tested the lock race on a scratch remote: two agents claim the same ticket → second push rejected, rebase conflicts on the lock file → loser backs off, winner's lock intact. 
Different tickets → both land cleanly. 
So push-and-fail works as the claim mechanism, like you said. Added one line to the prompt so the loser knows the conflict means "back off", not "resolve it".

suleimansh:
🤖 *agent comment*

One more commit after the review round: the drain E2E showed the new root `tickets` symlink was untracked, so any sweeping `git add -A` would commit it onto a code branch. `a95e097b` hides it through the repo-level exclude the moment it is created (mechanism extracted to `git-exclude.ts`, shared with the node_modules-links case; a user's own `tickets` path stays visible and untouched). The sibling leak — auto-onboarded repos never getting `.the-framework/.gitignore` at all — predates this PR and is filed as #1600.

Amendment (`9e4d0324`): the first cut used a bare `/tickets` rule, and the repo-level exclude speaks for every worktree — including the tf-data checkout itself, so data commits silently dropped ticket files (CI's control.test caught it). The rule is now the pair `/tickets` + `!/tickets/`: a trailing slash never matches a symlink, so the root link stays hidden while the data checkout's real directory keeps committing. Regression-tested in data-branch.test; full suites green (1468 + 772).

suleimansh:
Before the migration runs after this merges: can you hand-publish `@gemstack/the-framework` from current main? #1598's layout gate only exists in unpublished code — latest on npm is 1.4.2 (Jul 27) — so until a gate-carrying build is out, a stale installed build would happily operate on the migrated layout with nothing to refuse it. Order would be: publish → merge this → migrate.

Issue #1519, opened 2026-08-05
Extension/backend version mismatch: throw and block, not a warning


suleimansh:
From Discord (brillout, 8/3): now that the extension updates without manual steps —

> Maybe quick win: show a warning if extension version mismatches what the backend expects. Actually, I'd say throw error and block.

Agreed there that it should be **error + block** — a warning would just get ignored, and a silently stale extension produces exactly the confusing half-working behavior the dogfood loop keeps tripping on.

## Seam that already exists

The extension already reports its own version to the bridge: `bridge-endpoints.ts` carries a `version: string` field, sanitized on ingest (`raw.version`, capped at 32 chars, `'unknown'` fallback). What's missing is the backend's half:

- the daemon declares the extension version (or minimum version) it expects
- on hello/heartbeat with a mismatched (or `'unknown'`) version, the bridge **rejects the connection with an error naming both versions and how to update** — no degraded mode
- the dashboard surfaces the blocked state (the relay/bridge indicator should say "extension vX, backend expects vY" instead of looking merely disconnected)

## Why block rather than warn

A version-skewed extension doesn't fail loudly — it half-works (missed messages, silently ignored new fields), which reads as a TF bug and burns a debugging session. Refusing to pair makes the fix (update the extension) the only path forward, which is cheap now that the update is automated.

PR #1598, opened 2026-08-19
A skewed framework build is refused by the repo's layout marker (fix #1575)

### LINKED ISSUE #1575: Cloud web runs execute the published npm build, so their bookkeeping predates repo-side renames

suleimansh:
## The problem, plainly

Cloud web runs don't use your local build — the cloud environment installs the framework from npm, and the published version can be **older than the repo it works on**. That happened in #1574: the repo had renamed its archive folder from `sessions/` to `agents/` weeks earlier, but the published build didn't know, saved a session archive under the old name, and the bad commit sailed into a PR — caught only hours later by main's guard test, with an error about symptoms rather than the cause.

## The fix, plainly

The repo now carries a small committed file, `.the-framework/LAYOUT` — 7 lines saying where the framework keeps its bookkeeping:

```
framework-dir: .the-framework
archive-dir: agents
branches-dir: branches
events-file: events.jsonl
meta-file: agent.json
tickets-dir: tickets
queue-file: TODO_AGENTS.md
```

Before starting any session, the framework compares this file against what the running build would write. If they disagree, the session is **refused immediately** — exit 1, both layouts printed side by side, the fix named — before a single file is written. Same stance as the extension version gate (#1519): refuse, don't warn, because a skewed build doesn't fail loudly on its own — it half-works, and its wrong-layout commits look plausible to a human.

The file's content is **generated from the code's own constants**, and a lockstep test pins the checked-in file to that generation (the `EXPECTED_EXTENSION_VERSION` ↔ `manifest.json` trick). So nobody has to remember to bump anything: renaming a folder in code breaks the test until the file is regenerated in the same PR.

## What changed where

- `src/layout.ts` (new) — the marker derivation + `checkLayout()`. A repo *without* the marker passes ungated (how test fixtures and pre-gate repos keep working; install writes it, so every newly activated repo is gated).
- `src/cli.ts` — the gate runs in `driveAgent` right after the workspace is known, before any write.
- `src/install.ts` — seeds the marker; the seeded `.gitignore` gains `!LAYOUT` (its `*` rule would otherwise keep the marker out of the install commit).
- `src/layout.test.ts` — unit coverage + the lockstep test; `control.test.ts`'s committed-files guard now allows `LAYOUT`.
- `.the-framework/LAYOUT` + `.the-framework/.gitignore` — the repo's own marker, checked in.

## How this fits #1595 (data branch)

Compatible — and complementary. #1595 keeps every name this marker derives from (`tickets`, `TODO_AGENTS.md`, `agents`, …) but changes **where** they live (the `the-framework_data` branch). Two touchpoints:

1. When #1595 lands, add one derived line to the marker — `data-branch: the-framework_data` from its `DATA_BRANCH` constant — so the placement change itself becomes a layout change the gate catches. One line in `layoutMarker()`; the lockstep test then forces the checked-in marker to regenerate in that same change.
2. Ideally this PR merges and gets **published** before the #1595 migration: then a stale pre-#1595 cloud build refuses to run in the migrated repo instead of writing main-style bookkeeping into it — the gate is exactly the protection that migration wants.

Only textual overlap: both touch `.the-framework/.gitignore` (`!LAYOUT` must survive #1595's simplification).

## Limits

- The gate only protects skews **after the next npm publish** — a build published before this PR contains no gate and can't refuse. Publishing once after merge (e.g. the standing version-packages PR #1285) arms it for good.
- Tests: full node suite 1496/1496; live E2E in the comment below (install seeds it, skew refused with zero writes, no marker = ungated, fresh clone passes).

### DIFF
--- a/.changeset/layout-gate.md
+++ b/.changeset/layout-gate.md
@@ -1,0 +1,5 @@
+---
+'@gemstack/the-framework': minor
+---
+
+The layout gate (#1575): a build refuses to run in a repo whose committed layout marker records a different bookkeeping layout, instead of committing files under names the repo no longer uses. Install records the marker; the repo's own is pinned to the build by a test, so a layout rename cannot land without regenerating it.
--- a/.the-framework/.gitignore
+++ b/.the-framework/.gitignore
@@ -1,6 +1,7 @@
 # The Framework: the committed project DB; session state is transient.
 *
 !.gitignore
+!LAYOUT
 !LOGS.md
 !conversations/
 !conversations/**
--- a/.the-framework/LAYOUT
+++ b/.the-framework/LAYOUT
@@ -1,0 +1,7 @@
+framework-dir: .the-framework
+archive-dir: agents
+branches-dir: branches
+events-file: events.jsonl
+meta-file: agent.json
+tickets-dir: tickets
+queue-file: TODO_AGENTS.md
--- a/packages/the-framework/src/cli.SPEC.md
+++ b/packages/the-framework/src/cli.SPEC.md
@@ -8,6 +8,7 @@ The framework command: serves the dashboard in the foreground, and runs one agen
 - Three shapes share that wiring: the full build flow, a verbatim prompt (research included), and transparent mode — the driver completely raw.
 - Every agent settles identically: the quality follow-up if asked for and earned, then the handoff — commit what it left, push the branch, open a draft PR, merge only when authorized — then a project-log entry, written even for agents that stopped or crashed.
 - Settings that cannot apply say so before the spending; a stopped agent never publishes — publishing what it happened to reach is the opposite of what stopping meant.
+- A build whose bookkeeping layout differs from the one the repo records is refused before it writes anything, with both layouts and the fix named — the stale published build a cloud environment installs must fail in seconds with the cause, not hours later at the main branch's guard.
 - Once an agent starts it is never interrupted for quota. The gate is on starting, and it lives with the daemon that decides whether to start unattended work at all.
 - Ctrl+C aborts the agent itself, the driver's process tree included; a second press force-quits.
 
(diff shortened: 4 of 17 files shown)


suleimansh:
Live-tested end to end against a scratch repo, using the built CLI (`dist/bin.js --agent`, fake driver, isolated `XDG_CONFIG_HOME`):

1. **Install seeds the gate**: `installProject` on a fresh repo wrote `.the-framework/LAYOUT` with the derived content, the seeded ignore carries `!LAYOUT`, and both land tracked in the install commit.
2. **Matching marker → runs**: a session in that repo completed normally (exit 0).
3. **Skewed marker → refused**: with the marker edited to `archive-dir: sessions` (the exact #1574 skew), the session exits 1 immediately, printing both layouts side by side and the fix for each direction. Checksummed every file under `.the-framework/` before and after: **byte-identical — the refused run wrote nothing.**
4. **No marker → ungated**: with the marker deleted, the session runs (pre-gate repos keep working).
5. **Fresh clone carries the gate**: a `git clone` of this branch (how a cloud VM gets the repo) contains the tracked marker, and a session in the clone passes the gate.

PR #1609, opened 2026-08-20
Migrate all SPEC.md files to the new upstream sdd.md template

brillout:
Upstream [sdd.md](https://github.com/brillout/sdd) redefined the SPEC.md template (commit `f54547c`): `## TLDR` is no longer a legal section, flows live under `## Flows` with an optional `## Flows — TL;DR` summary, why/history clauses belong in `## Rationales`, and `## User Stories` / `## Glossary` are new optional sections. This PR migrates the whole tree — all 558 SPEC.md files checked, 231 changed (+967/−451) — and applies the new writing style (product-manager reader, zero ambiguity, self-contained sentences, no history framing in flows). Only `*.SPEC.md` files are touched.

What the migration does:

- **`## TLDR` is gone everywhere** (212 files): where the bullets described behavior they became `## Flows` (byte-identical where already compliant); where a real flows section existed they were merged; history/why clauses were extracted into `## Rationales`, reworded in present tense.
- **`## User Stories`** added only where a file's flows directly realize something a user does, each story traceable to `FEATURES-SPEC.md`: root, `packages/the-framework`, `src`, `cli`, `auto-pm`, `tickets`, `todo-loop`, the dashboard root and components directory, the daemon's dashboard directory, the chrome extension, the website, and the Actions agent workflow. Leaf plumbing and single UI components get none.
- **`## Flows — TL;DR`** only where a Flows section is long enough to earn a summary: `packages/the-framework`, `dashboard`, `src/dashboard`.
- **`## Glossary`** where a spec owns coined vocabulary: `packages/the-framework/SPEC.md` (driver, location, attended/unattended, gate, ticket, queue entry, empty agent, preset) and `driver/SPEC.md` (driver).
- **Root `SPEC.md`** restructured per the new Hierarchy rules: product story, how the top-level subsystems relate, product-level user stories.
- **Canonical footer everywhere** — two `.github` specs carried a stale `## Before writing SPEC.md files` variant.
- A few grounded one-word corrections where the spec was unambiguously wrong against the code (e.g. the Claude Code usage question is answered by the agent, not the driver; each *driver*, not each agent, declares its CLI surface).
- No mermaid diagrams added or removed.
[...]

PR #1610, opened 2026-08-20
The watch relay's orphans are removed: the view, the ?run= route, and the specs that still promised it

suleimansh:
🤖 *agent PR*

Follow-up to #1609, which deliberately left these for this pass ("the stale flow sentences stay preserved for the queued relay-orphan cleanup", b560f9d3).

**What the user sees.** Nothing changes. That is the point — every line removed here was already unreachable.

The watch relay was deleted in #1536 (`f395f0c1`), but its client half stayed on main: `RelayView`, a `?run=<id>` early return in `App.tsx` that swapped the whole shell for it, and a `local` flag threaded through `useWorking` / `useDaemonHealth` / `useFavicon` purely to switch them off on a host that no longer exists. Nothing served any of it: the id landed in the `projectId` slot, resolved to nothing, and the page said *"This shared session isn't available."*

**Two things were dead by inspection once the view went.** `AgentFeed`'s `showSessionLink` / `showName` / `showStatus` were always `false` from `AgentView`, its one remaining caller — and with all three false the `AgentOverview` they fed hit `if (!sessionLink && !status) return null` every time. Both go. `AgentOverview` itself stays: `ProjectHome` renders it with its defaults, where it shows something.

**The specs that still promised the feature go too:**

- root `SPEC.md` — "A shared link re-serves one agent's event stream read-only, from the same daemon that owns it"
- `dashboard/SPEC.md` — the "Watch mode" flow, and the actions menu's "copy a shareable watch link" (`AgentActionsMenu.tsx` offers *Copy resume command* / *Copy session id* and never offered this)
- `dashboard/App.SPEC.md` — the shared-watch-link flow, keeping its daemon-banner half
- `dashboard/components/SPEC.md` — "a read-only shared watch view" in the shell's page list
- `src/dashboard-rpc/events.SPEC.md` — "the relay's own agents streaming from memory"
- `AgentOverview.SPEC.md`, `favicon.SPEC.md`, `FileTree.SPEC.md`, `use-daemon-health.test.SPEC.md` — the clauses describing what the relay opted out of

The last of those were findings on #1609's own contradiction list, closed here at the source rather than one spec at a time.

**Left alone on purpose: the device relay (#1067).** `/_relay`, `relay-endpoints.ts` and `relay-dispatch.ts` are a live, token-guarded feature that happens to share the word. Cutting on the name rather than on what the module does is exactly what nearly went wrong in #1536.

`FEATURES-SPEC.md` needs no edit — the watch link was never listed there, which is part of how the stale specs went unnoticed.

**Tests.** Suite green: 1489 framework, 774 dashboard, typecheck clean. The two tests that only existed to prove the relay's opt-outs (`useFavicon(true, false)`, "disabled (the relay) never probes") are removed with the parameters they tested; every other test is untouched and still passes, which is the check that nothing user-visible moved.

All nine edited specs were verified against the new `sdd.md` template — allowed sections only, correct order, byte-exact footer.


suleimansh:
🤖 *agent comment*

SPEC walkthrough — nine files, each as intro + Flows, then the diff, then what it means against the rest of the file. Every change is a removal: a sentence describing the watch relay, which #1536 deleted.

---

**1. `packages/the-framework/SPEC.md`** — the product. The paragraph that changed, `**Surfaces.**`, is one of the prose blocks after the Flows; it lists every way the product is reached.

Diff:

```diff
-… the device's token is saved only in the user's own browser and handed to the local daemon per call. A shared link re-serves one agent's event stream read-only, from the same daemon that owns it. An agent can also run elsewhere: …
+… the device's token is saved only in the user's own browser and handed to the local daemon per call. An agent can also run elsewhere: …
```

Meaning: the surfaces list is now exactly the surfaces that exist — the dashboard, a saved remote device, the cloud/Actions locations, the Chrome the agent drives, Discord. The removed sentence sat between the device relay and the remote locations, which is why it read as one of the family; it is the only member of that list with no code behind it.

---

**2. `packages/the-framework/dashboard/SPEC.md`** — the dashboard. Two removals.

Flows, the line that goes:

```
- Watch mode: opened against a shared link, the same app renders one agent read-only.
```

Diff: that bullet is deleted outright.

And in the `**The agent view**` paragraph:

```diff
-… an actions menu (stop, open in editor or on GitHub, remove worktree, delete it, copy a resume command, copy a shareable watch link).
+… an actions menu (stop, open in editor or on GitHub, remove worktree, delete it, copy a resume command).
```

Meaning: the Flows bullet described a second mode the whole app could enter; there is no such mode, and the `?run=` route that reached it is deleted in this PR. The menu item never existed at all — `AgentActionsMenu.tsx` offers *Copy resume command* / *Copy session id* and nothing else — so the list now matches the menu.

---

**3. `packages/the-framework/dashboard/App.SPEC.md`** — the one page.

```diff
-- A shared watch link renders that one agent read-only; a daemon that stops answering gets a banner, so a dead backend never looks like a quiet agent.
+- A daemon that stops answering gets a banner, so a dead backend never looks like a quiet agent.
```

Meaning: one bullet was carrying two unrelated facts. The banner half is live and stays; the watch-link half was the page's `?run=` early return, removed here. The file's first Flows bullet — "the address is the selection" — is now true without exception: there is no query parameter that replaces the whole shell.

---

**4. `packages/the-framework/dashboard/components/SPEC.md`** — the components directory. The shell's page list:

```diff
-Its pages are the Overview board, the project home/launcher, one agent's view, the cross-project tickets pages (list, per-ticket detail, per-ticket plan), Settings, a read-only shared watch view, and not-found.
+Its pages are the Overview board, the project home/launcher, one agent's view, the cross-project tickets pages (list, per-ticket detail, per-ticket plan), Settings, and not-found.
```

Meaning: the list is the routes `App.tsx` can render. `RelayView` is deleted in this PR, so it leaves the list with it.

---

**5. `packages/the-framework/src/dashboard-rpc/events.SPEC.md`** — the live event stream.

```diff
-- An agent with no file on this host — the relay's own, or one relayed from a device — streams from memory instead.
+- An agent with no file on this host — one relayed from a device — streams from memory instead.
```

Meaning: the flow itself is unchanged — an agent whose journal is not on this disk is served from memory. What changes is how many sources that covers: one, the device relay (#1067). This was finding 2 on #1609's own contradiction list; it is closed here at the source.

---

**6. `packages/the-framework/dashboard/components/AgentOverview.SPEC.md`** — the overview cards above a feed.

```diff
-The agent-overview cards projected from the event stream — the status line and an honestly-labelled link to the live session — each rendered only once its data has arrived, with embedding views opting out of the parts their own chrome already shows.
+The agent-overview cards projected from the event stream — the status line and an honestly-labelled link to the live session — each rendered only once its data has arrived.
```

Meaning: the opting-out clause described three flags (`showSessionLink`, `showName`, `showStatus`) that existed so the relay watch could keep the cards while the agent's own view suppressed them. With the relay gone, the agent view suppressed all three — so the component returned `null` on that path every time. The flags and that render are removed; the component keeps its one live caller, `ProjectHome`, which always wanted the full cards.

---

**7. `packages/the-framework/dashboard/lib/favicon.SPEC.md`** — the tab icon.

```diff
-The tab icon follows the work: the still logo while nothing is running, the animated one while an agent is working — and a view that cannot know (the relay) leaves the icon alone.
+The tab icon follows the work: the still logo while nothing is running, the animated one while an agent is working.
```

Meaning: the hook had an `enabled` parameter whose only false caller was the shell handing the tab to the relay view. One caller, one behavior now.

---

**8. `packages/the-framework/dashboard/components/FileTree.SPEC.md`**

```diff
-- Localhost-only: with no checkout to list (the relay), the tree renders nothing.
+- With no checkout to list, the tree renders nothing.
```

Meaning: the behavior is unchanged and still reachable — the parenthetical named the relay as the case that produced it, and that is no longer the reason. The empty-tree path stays.

---

**9. `packages/the-framework/dashboard/lib/use-daemon-health.test.SPEC.md`** — what the liveness-probe tests cover.

```diff
-Covers the liveness probe: an answering daemon reads healthy, a failing one flips to down, and the shared watch view (which has no daemon of its own) never probes.
+Covers the liveness probe: an answering daemon reads healthy, and a failing one flips to down.
```

Meaning: the third case tested the `enabled=false` path, which existed only for the relay. The test is removed with the parameter; the two real cases stay.

---

`FEATURES-SPEC.md` is unchanged, and that is itself part of the story: the watch link was never listed there, which is one reason six specs went on describing a feature that had been deleted.

suleimansh:
🤖 *agent comment*

Superseded by #1611, which does this cleanup and much more (the website copy bug and the 13 spec-drift findings). Closing this in favour of it — no need to review it.

Recording the two things #1611 leaves behind, so they are not lost:

**Dead code the relay's removal created.** `AgentFeed`'s `showSessionLink` / `showName` / `showStatus` are always `false` from `AgentView`, its only remaining caller — and with all three false the `AgentOverview` they feed hits `if (!sessionLink && !status) return null` on every render. #1611 updates AgentFeed's comment but keeps the props and that render. (`AgentOverview` itself stays either way: `ProjectHome` renders it with its defaults, where it shows something.)

**A dead parameter rather than a defaulted one.** #1611 has `useWorking` / `useDaemonHealth` / `useFavicon` "fall back to their defaults". With the relay gone nothing passes `false` at all, so the `enabled` parameter is dead — along with the two tests that exist only to prove it (`useFavicon(true, false)`, and "disabled (the relay) never probes").

**Four specs still carrying relay clauses** that #1611 does not touch: `AgentOverview.SPEC.md`, `favicon.SPEC.md`, `FileTree.SPEC.md`, `use-daemon-health.test.SPEC.md`.

All of it is removal of code nothing can reach, so nothing is lost by it landing later. Will pick it up as a small follow-up once #1611 is in.

Issue #1623, opened 2026-08-21
A boot that cannot reach GitHub arms a Discord flood: an empty first poll cannot be told from a failed one


suleimansh:
🤖 *agent issue*

Found while grounding the spec-vs-code list on #1613. That list names this one, but locates it in the wrong place — fixing where it points would change nothing.

**`keyed-watcher.ts` is correct.** Its own doc states the rule and its code keeps it:

> a failed project scan or projection just yields no new items that cycle, and never counts as a poll: the baseline must come from a real read, or a failed first poll would make the next good one announce everything pre-existing as new

```js
try {
  items = await opts.build(await opts.projects())
} catch {
  return            // <- returns before observe(), so warmedUp stays false
}
```

**The problem is that neither call can throw.** Both of the watcher's inputs fail soft, one layer up:

- `listSummaries` (`daemon-services.ts:130`) — `await listProjects(undefined, env).catch(() => [])`
- `buildInterventions` (`interventions.ts:89`) — `await prs(project.path).catch(() => [])`, per project

So on a boot where the registry cannot be read, or `gh` is not yet authenticated, or GitHub is unreachable, the first poll does not fail. It *succeeds, empty*. `SeenTracker.observe([])` then flips `warmedUp` against an empty baseline, and the first poll that does reach GitHub reports every already-open PR as new.

Both Discord watchers are wired this way — the "needs you" queue and the New activity feed — so the failure is one message per pre-existing item, at once, to the user's Discord.

## Why it stays invisible

Every layer behaves exactly as designed. The watcher refuses to warm up on a failure it never sees; the builders refuse to let one unreachable project sink the whole poll; the registry read refuses to make a missing file fatal. The bug lives in the seam: **an empty read and a failed read arrive as the same value**, and only the watcher needs to tell them apart.

## What a fix has to decide

The watcher cannot distinguish them from `T[]` alone, so one of:

- **The builders report failure.** `build` returns items *and* whether the read was whole (or throws, which the watcher already handles). Most honest, and the information exists at the `catch` that currently discards it.
- **An empty first poll does not warm up.** One line in `SeenTracker`, and it costs the case it is wrong for: a genuinely empty first poll on a quiet machine stays un-warmed until the first item appears, which then announces as new. That is a notification nobody asked for — the milder failure of the two, but still a wrong one.

The first is the real fix. The second is what you would ship today if the first were too large.

[...]

PR #1626, opened 2026-08-21
The dashboard's notifications keep their baseline the same way the daemon's do

### LINKED ISSUE #1625: The dashboard floods the same way on a bad boot, and its notifications are on by default
🤖 *agent issue*

Found while fixing #1623. Same seam, the other surface — and this one is **on by default**, where the Discord half is opt-in.

`dashboard/lib/use-notifications.ts` decides its baseline by *counting observations*:

```js
/** Observations to treat as baseline before notifying: the initial `[]` plus the first fetch. */
const WARMUP = 2
...
if (observations.current < WARMUP) {
  observations.current += 1 // still warming up: fold these into the baseline, don't notify
}
```

Two observations, whatever they contained. So on a page loaded while GitHub is unreachable:

1. the initial `EMPTY_INTERVENTIONS` — baseline slot 1
2. the first fetch, which resolves `[]` because every read underneath forgives its own failure — baseline slot 2
3. the first fetch that reaches GitHub — **past warm-up**, so the entire pre-existing queue is announced as new

The user gets one notification saying *"N items in your Human Queue"*, listing every pull request that was already open before they opened the tab.

## Why this one matters more than #1623

The defaults are not symmetric, and `preference-defaults.ts` says so deliberately:

```js
methods: { browser: true, discord: false },
categories: { humanIntervention: true, newActivity: false },
```

The browser bell and the "needs you" category are both **on unless you turn them off**. So the browser flood needs no configuration at all to reach someone — while the Discord one requires a webhook plus two opt-ins.

## The fix is the daemon-side one, one layer up

#1623's fix makes the builders report which projects they read *whole*, and keeps the baseline per project. The browser cannot use that yet: `onInterventions` / `onActivity` hand it only the item array, so the RPC has to carry `whole` through, and the hook has to warm up per project instead of by count.

That also retires the observation counter and the guard it needed — the effect currently has to special-case `items === lastItems.current` because a toggle flip would otherwise burn a warm-up slot and replay the whole backlog. With a per-project baseline there is nothing to burn.

The `SeenTracker` #1623 leaves behind is already the right shape and is browser-safe (its only import is a type). Sharing it would put both surfaces on one baseline rule, which is what `use-notifications.ts` says it is aiming at: *"One engine — the identity half … is the same code the daemon's Discord notifier runs … this is the client side of the same move."* The baseline half is the piece that never made the trip.

## Note

Depends on #1624 — it defines the `{ items, whole }` shape this would carry through the RPC.

suleimansh:
Closes #1625. **Stacked on #1624** — that PR is the base, and defines the `{ items, whole }`
shape this carries through the RPC. Review or merge that one first.

The dashboard had the same bug as #1623, on the surface that is **on by default**.

## What was wrong

`use-notifications.ts` decided its baseline by counting:

```js
/** Observations to treat as baseline before notifying: the initial `[]` plus the first fetch. */
const WARMUP = 2
```

Two observations, whatever they held. On a page opened while the daemon could not reach GitHub,
both went on empty lists — and the first fetch that did reach GitHub announced every already-open
pull request as new.

The defaults make this the more reachable of the two floods. `preference-defaults.ts` is explicit
that the polarities are not uniform: `browser: true`, `humanIntervention: true`, against
`discord: false`. The Discord flood needs a webhook and two opt-ins; this one needs nothing.

## The fix

The same rule as #1623, carried one layer up:

- `onInterventions` / `onActivity` return the `{ items, whole }` the builders already produce. The
  panels read `items`; only the notifier needs the rest.
- The hook keeps its baseline **per project**, using the daemon's own `SeenTracker`. Its only
  import is a type, so it is browser-safe as it stands.

`use-notifications.ts` already said this is where it was heading — *"One engine — the identity
half … is the same code the daemon's Discord notifier runs … this is the client side of the same
move."* The baseline half is what had never made the trip.

## What this deletes

- **The counter**, and the guard it needed. The effect had to special-case `items ===
  lastItems.current` because a toggle flip would otherwise burn a warm-up slot and replay the
  backlog. With a per-project baseline there is no slot to burn; the identity check stays only to
  skip pointless work, and its comment no longer describes a hazard that no longer exists.
- **`pickNewInterventions` / `pickNewActivity`**. `SeenTracker` subsumes them — it picks and folds
  in one step — and the browser hook was their only caller. Their re-exports and their test go too.

## Tests

Three new ones on the hook, and I checked they fail for the right reason rather than trusting that
they pass: flipping the `unread()` helper to claim a whole read (the one signal the fix turns on)
makes four of them fail, including both #1625 tests.

- a poll that reached no project is not a baseline, so the backlog stays quiet
- a project that cannot be read does not hold back a project that can
- turning notifications on does not replay what arrived while they were off (the behaviour the
  deleted guard used to protect, now protected by the design)

Full suite green: 1520 node, 777 vitest, typecheck clean.

## Checked in the real dashboard

Built and served the branch against a project with one pre-existing open PR: the RPC returns
`{"items":[…#3…],"whole":["tf-1334-dogfood-f8nvbw"]}`, the Human Queue renders it, the tab-title
badge counts it, and the console is clean on load and reload — the RPC shape change was the real
risk here, and it does not reach the UI.


suleimansh:
Superseded by #1627 — GitHub closed this automatically when #1624 merged and deleted its base branch. Same commit, rebased onto `main`.

Issue #1506, opened 2026-08-05
Add tooltips to prompt buttons


brillout:
That shows essential information:
- What's going to happen
- What model is going to be used
- Where it's run (locally, CC web, ...)
- More?

See also:
- https://github.com/gemstack-land/the-framework/issues/1507

Issue #1066, opened 2026-07-23
Run on: unify the execution-target picker into one flat list (drop redundant Local, don't lose the prompt)


suleimansh:
Part of #1049. Live-testing the shipped "Run on" gear (#1050 + #1052) surfaced a UX problem: the menu presents **two different axes as one list**, which confused a real user.

- **Current device / GitHub Actions / Claude web** = *where this run executes* (per-run driver choice; keeps the prompt; checkmark-select).
- **A device I have → Local / \<box\>** = *which daemon the dashboard talks to* (a full-page navigation that reloads and **discards the typed prompt**, and starts no session).

Two concrete confusions observed:
1. Picking a device deleted the in-progress prompt and started nothing (the user expected "run my prompt on that device").
2. **"Local" and "Current device" are effectively the same machine**, so the list reads as having a duplicate.

## Scope (UX only, no new transport)

- Collapse into **one flat "Run on" list**: `This machine ✓ · GitHub Actions · Claude web (coming soon) · \<each saved device\> · + Add a device`. Remove the separate "A device I have" section header and the redundant "Local" row (it duplicates "Current device").
- **Preserve the composer's typed prompt across a device-connect navigation** (carry it to the remote origin, e.g. sessionStorage or a query param, and rehydrate the composer there) so switching devices never nukes unsaved input.
- One clear selection model: the checkmark reflects the actual current target/connection (the driver target when on the local daemon; the connected device otherwise).

Files: `OptionsMenu.tsx` (merge `RunTargetSub` + `ConnectionSection`), `Composer.tsx` (fold `runTargets` + profiles into one list; prompt preservation), `lib/profiles.ts` (connect flow carries the draft).

Out of scope: making a connected device a true no-navigation run target (that is the remote-run relay, tracked separately). This issue keeps the existing navigate-on-connect mechanic; it only unifies the presentation and stops losing the prompt.

PR #1634, opened 2026-08-22
A routine can be configured before it is run (fix #1507)

### LINKED ISSUE #1507: Add "Configure first, then run" to prompt buttons
Hovering a prompt button shows a secondary action "Configure first, then run" => opens the new session page with the textarea prefilled with the prompt.

Not sure how the UI would look like, maybe as a dropdown?

See also:
- https://github.com/gemstack-land/the-framework/issues/1506

suleimansh:
Closes #1507.

**The problem in one line:** `Run now` on the Routine work card spends an agent on settings you cannot see from that card. The model and where it runs come from the Global options, so changing them meant leaving, editing preferences, and coming back.

## What it looks like

Each `Run now` is now a split button. The chevron opens one item:

> **Configure first, then run**
> Opens the launcher with this prompt, so you can set the model and where it runs.

Clicking it hands the routine's prompt to the picked project's launcher, where the model, the location, and the prompt itself can all be changed before anything is spent.

## Two decisions worth your eye

**A split button, not a hover-revealed control.** The issue floats "hovering a prompt button shows a secondary action ... maybe as a dropdown?" — I went with the dropdown half and made its trigger always visible. A control that exists only under a mouse is reachable by neither keyboard nor touch, and the chevron keeps the common click at one click. Say the word if you'd rather it were hover-gated.

**The drain row says what it costs.** "Spin up agents working on the AI queue" is the one row whose `Run now` fires a fan-out sweep, and the launcher can only ever send one agent — so its two halves really do different jobs. Its menu item reads *"Opens the launcher with this prompt — one agent, not the fan-out."* rather than looking like the same action.

## How it's built

No new mechanism: the prompt travels on the draft stash a hot ticket already uses (#1066/#1139), which the launcher takes once as it mounts. Verified in the browser that a reload leaves the composer empty rather than re-seeding.

`RoutineWork` gains one prop, `onSelectProject` — the same one `DashboardPage` already passes to the Human Queue and the hot tickets.

## Verification

**Dogfooded on the real daemon**, not only jsdom — the lesson from #1500 was that a component can render correctly and still be mounted where nobody sees it:

- Overview → "Update from GitHub" chevron → menu opens with the item and its description
- clicking it lands on the `gemstack` launcher with the routine's whole prompt in the box, model picker showing Opus, no agent started
- reload → composer empty, so the draft is taken once
- the drain row's chevron shows the fan-out caveat instead

**Tests: 5 new, each proven by breaking the code it guards.** Dropping the stash, pointing at the first project instead of the picked one, flattening the two descriptions to one, inverting the drain check, and gating the menu item on `busy` each failed exactly the intended test and nothing else. Full suite green: 1536 node + 788 dashboard.

`RoutineWork.SPEC.md`, its test SPEC, and `FEATURES-SPEC.md` all updated.

Next up is #1506 (tooltips on these same buttons), which is where "what model / where it runs" gets said on the card itself rather than only after you arrive at the launcher.

### DIFF
--- a/FEATURES-SPEC.md
+++ b/FEATURES-SPEC.md
@@ -24,6 +24,7 @@ happens while nobody is at the keyboard.
 - Start an agent from a ticket row
 - Start an agent from a queue entry's play button
 - "Run now" on a routine
+- "Configure first, then run" on a routine — the launcher opens with its prompt, so the model and location can be set first
 - The whole CLI is one command: `the-framework` serves the dashboard — four options, no verbs
 - `--host` / `--port`, the two things a browser cannot be asked; `--help` / `--version`
 - Reach the dashboard from another machine — non-loopback bind behind a generated shared token
--- a/packages/the-framework/dashboard/components/DashboardPage.tsx
+++ b/packages/the-framework/dashboard/components/DashboardPage.tsx
@@ -69,7 +69,7 @@ export function DashboardPage({
 
         {/* Routine work sits below the AI Queue (#1139/#1159): the scheduled jobs and the button
             that fires one now. */}
-        <RoutineWork onAgentStarted={onAgentStarted} />
+        <RoutineWork onAgentStarted={onAgentStarted} onSelectProject={onSelectProject} />
 
         <HotTickets onSelectProject={onSelectProject} onSelectAgent={onSelectAgent} />
       </div>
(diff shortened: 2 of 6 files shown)


suleimansh:
🤖 *automated*

**Merging this — but one decision in it is yours, and it is cheap to reverse.**

`Run now` on the Routine work card is now a split button. The chevron opens one item, *"Configure first, then run"*, which hands the routine's prompt to the picked project's launcher, so the model and where it runs can be set before an agent is spent.

The issue floated *"hovering a prompt button shows a secondary action … maybe as a dropdown?"*. I built the dropdown half and made its trigger **always visible**: a control that exists only under a mouse is reachable by neither keyboard nor touch. That is the one thing worth overruling if you disagree — it is a chevron, so switching it is a small follow-up, not a revert.

- **Option A** — chevron always visible (what shipped).
- **Option B** — chevron appears on hover only.

Second, smaller one: *"Spin up agents working on the AI queue"* is the only row whose `Run now` fires a fan-out sweep, and the launcher can only ever send one agent. Its menu item therefore reads *"one agent, not the fan-out"* instead of looking like the same action as the other rows. Say if that is more wording than it needs.

Next is #1506 — tooltips on these same buttons, where "what model / where it runs" gets said on the card itself instead of only after you arrive at the launcher.

Issue #1279, opened 2026-07-27
Routine runs started from the dashboard park forever; the same routine from the sweep completes and hands off


suleimansh:
Found by driving the demo flow end to end on a real daemon with real agents (isolated daemon, scratch repo `suleimansh/tf-drive-scratch`, main build as of 7aedf24). This is the dead end the quick-win beat hits.

## What happens

Clicking a routine card fires `sendStart` with the preset's prompt. That run:

1. did the triage correctly and committed to its branch 69 seconds in,
2. then entered the stay-open chat phase (`run.ts:456`, the #714 live-chat loop) and **parked as "running" indefinitely** (observed 35+ minutes, `settledAt` set, nothing after),
3. its armed push+PR handoff never fired, because the handoff runs when the process ends and a parked conversation never ends,
4. so no PR, the queue on main never fills, and the dashboard shows an eternal session. The flow's next beat cannot happen.

The identical routine started by the auto-PM sweep behaves correctly: the rotation run it fired in the same session emitted `ready-for-merge`, had its choice auto-answered, **ended on its own after 3.5 minutes**, and its handoff pushed the branch. So the machinery all works; only the entry point decides whether a run ever finishes.

The one way out today is Stop: on a *parked* run the chat loop exits cleanly, the run ends `done`, and the handoff fires (verified live; the push landed). But nothing tells the user that, Stop reads as "abort", and a Stop during actual work goes through the abort path where the handoff is deliberately skipped (`cli.ts:1606`). Expecting the user to Stop a run to get its PR is not a flow.

Also observed: the hand-fired run never emitted `set-session-name` or `ready-for-merge` (the sweep-fired one did), so even the ready-for-merge machinery never saw it.

## Fix direction

A run started from a routine/preset card is not a conversation. It should run the way the sweep runs it: unattended, ending at settle, handoff firing. Keep the stay-open chat for hand-typed prompts, which is what #714 was for. (Alternative: fire the armed handoff at settle rather than at exit, and let the park continue afterwards.)

Issue #1143, opened 2026-07-25
Bug: cannot select Fable


brillout:
Low-prio, post-MVP.

<img width="620" height="273" alt="Image" src="https://github.com/user-attachments/assets/dc78d88f-9c7a-4e0d-aa09-31174c057f05" />

suleimansh:
<img width="438" height="233" alt="Image" src="https://github.com/user-attachments/assets/9b398abb-721d-48cb-9536-b90e2c151674" />

Seems to have been fixed.
But I would like to remove the 'Default' option, its useless.
Default = unknown model

suleimansh:
🤖 *curated*

Done on `main`: Fable is in the model list (`dashboard/lib/agent-settings.ts:16`), and the meaningless **Default** entry from the follow-up comment is gone — `Composer.test.tsx:149` asserts no Default button exists; the unpinned state is now described as "the CLI's own default" (`agent-settings.ts:29`). Closing.

PR #1636, opened 2026-08-22
A routine's Run now says what it will spend, before it spends it (fix #1506)

### LINKED ISSUE #1506: Add tooltips to prompt buttons

suleimansh:
Closes #1506.

**The problem in one line:** `Run now` on the Routine work card spends an agent on a model and a place that are nowhere near the button. #1634 gave you a way to go and set them; this says what they currently are, before you press anything.

## What it looks like

Hovering any `Run now`:

> Bring `tickets/` up to date with the GitHub issues. An empty `tickets/` gets a full first import.
> Claude Code · Opus · This machine
> Starts one agent in gemstack, unattended — nothing is asked mid-run.

Three lines for the issue's three questions — what happens, which model, where it runs — plus the one fact that is arguably the biggest hidden cost: a card-fired routine runs unattended (#1279), so nothing will be asked mid-run.

## Two decisions worth your eye

**The first line is the preset's own sentence, not new copy.** Presets already carry a one-liner that the launcher's menu shows on hover; the routine now carries it too, exactly the way it already carries the preset's label. So the launcher and the card describe the same work in the same words, and rewording a preset rewords both.

Two presets had no such sentence and now do — **Plan tickets** ("Turn `tickets/*.md` into costed plans (`tickets/*.plan.md`)") and **Drain queue** ("Work the entries already on the queue (TODO_AGENTS.md)"). Plan tickets is in the launcher menu, so that one is user-visible there too. Reword either if you want them said differently.

**The drain row reports something else entirely, and it has to.** Its `Run now` is the sweep, not a start, and two things the other rows promise are simply untrue of it:

- it visits **every project the daemon watches**, not the one picked in "Run in";
- each of those resolves its own committed `the-framework.yml` on top of these preferences (`resolveProjectAgentOptions`, `daemon-services.ts:149`), so naming a model here would be a guess.

So it says:

> Work the entries already on the queue (TODO_AGENTS.md)
> Each project's own settings decide the model and where it runs.
> Sweeps every project the daemon watches, up to 2 agents each, unattended.

The number is the live concurrency setting from the same card.

## How it's built

The model lists, the run-target names and the "nothing pinned" wording moved into `dashboard/lib/agent-settings.ts`. They were spread across four surfaces — the launcher's driver tree, the gear's "Run on" list, the Settings page and now this card — and a tooltip that reports settings is only worth anything if it says exactly what the launcher says. One side effect: the Settings page called the local machine "This device" while the gear called it "This machine"; there is one name now.

A model is still only named inside its own driver's list (#1143): nothing pinned, or a model pinned on the other driver, reads as "the CLI's own default" rather than borrowing the first entry.

## Verification

**Dogfooded on the real daemon**, both rows and both refactored surfaces:

- Overview → hover `Run now` on "Update from GitHub" → the three lines above, with `Claude Code · Opus · This machine` matching the actual preferences
- hover the drain row's `Run now` → the sweep wording instead, with the live concurrency of 2
- the launcher's model menu still lists Fable / Opus / Sonnet / Haiku with Opus checked, i.e. the moved catalog still feeds the tree it came from

**Tests: 4 new, each proven by breaking the code it guards.** Dropping the preset's sentence, falling back to the first model when none is pinned, hardcoding the run target to local, giving the drain row the ordinary settings line, and ignoring the concurrency in its fan-out line each failed exactly the intended test and nothing else. Full suite green: 1536 node + 792 dashboard.

`RoutineWork.SPEC.md`, its test SPEC, `preset-catalog.SPEC.md`, the new `agent-settings.SPEC.md` and `FEATURES-SPEC.md` all updated.


suleimansh:
🤖 *automated*

**A second commit: the launcher's preset hints were being painted behind the menu that triggered them.** Not a regression from this branch — I found it by hovering the two preset one-liners the branch adds, and it was equally true of the ones that shipped months ago.

`components/ui/tooltip.tsx` put `z-50` on the popup, and Base UI leaves that element `position: static`, where a z-index does nothing at all. The menu and the popover both carry theirs on their **positioner**, which is absolutely positioned — so both won, every time.

Measured in the page before changing anything, with the "Suggest new features" hint open inside the presets menu:

| | |
|---|---|
| tooltip rect | x 264→546, y 454→480 |
| menu rect | x 277→533, y 168→858 |
| `elementFromPoint` at the tooltip's own centre | `SPAN` of the **menu item** |

So the hint was rendering correctly and losing the paint. Setting the z-index on the positioner in the live page flipped that same probe to the tooltip itself, which is exactly what the one-line change does.

Fixed, rebuilt, and re-checked in the browser: `Turn `tickets/*.md` into costed plans (`tickets/*.plan.md`)` now reads over the menu. Full suite still green (1536 node + 792 dashboard).

**No test.** jsdom computes no stacking and the classes carry no CSS there, so the only assertable thing would be the class name — which would pass whether or not the tooltip is visible to a human, and would have passed for the whole time this was broken. The reason it must sit on the positioner is written where the change is instead.

PR #1640, opened 2026-08-22
Run now on the planning routine spends the concurrency, not one agent (#1204)

suleimansh:
🤖 *curated*

**Draft: this is Option B from #1204, built so the choice can be seen rather than imagined. One word merges or closes it.**

@brillout — you wrote *"as long as there isn't any redundant work being done, I don't see a problem with concurrent agents."* This is the one place that condition was met and the setting still did nothing.

## What was wrong

Two routines can fan out safely — a drain takes one entry *off* the queue, a plan agent writes one ticket's *own* sibling files — so their agents do disjoint work. Only one of them did it from the card:

| | the daemon, unattended | clicking Run now |
|---|---|---|
| **drain** | up to `Concurrent agents` | up to `Concurrent agents` |
| **Plan tickets** | up to `Concurrent agents` | **exactly one, always** |

Same routine, two behaviours depending on who fired it. The tooltip said "Starts one agent" and was telling the truth about the click — which is why this reads as a missing feature rather than a bug.

## What changed

The sweep's narrowing stops being drain-shaped: `drainOnly: true` becomes `only: 'drain' | 'plan'`. A plan-only tick reaches the same claim-then-start path the rotation already takes — candidates enumerated, ticket locks pushed *before* any agent starts, one agent per locked ticket.

**The locks are the point.** Without them two clicks a second apart plan the same ticket twice, which is the exact double-work this thread is about. This reuses the sweep's existing lock path rather than adding a second one.

`projectId` scopes a sweep to one project. The drain sends none and visits them all — what its tooltip promises. Planning is the picked project's own work and carries one.

## Three things a narrowed click must not do

Each is its own test:

- **work the queue instead** — the queue-picked mode would send a plan click to the drain whenever entries are waiting
- **override a switched-off routine** — an unticked box stands the click down and says so, rather than firing a routine the card shows as off
- **cost the rotation its turn** — a click borrows the tick for a routine it named, so the cycle stays where it was; the rule a due maintenance sweep already follows

## The tooltip follows the behaviour

Before → after, same `Concurrent agents: 2`:

> Starts one agent in gemstack, unattended — nothing is asked mid-run.

> **Starts up to 2 agents in gemstack, one per open ticket, unattended.**

Verified in the running dashboard, not only in tests.

The menu item moved too: **"one agent, not the fan-out"** was true of the drain and became true of planning the moment its button stopped being one start. It reads `fansOut`, not a routine name.

## On testing

Every new test was checked by breaking the code it guards — five deliberate reversions in `auto-pm.ts`, three in `RoutineWork.tsx`, each caught by exactly the test that should catch it.

**One test failed that check and had to be rewritten.** The plans-instead-of-draining test passed with the fix reverted: it asserted the job's *name*, which stays `plan` either way, when the real difference is the *batch* — entries off the queue instead of tickets. It asserts the prompts now, and fails as it should.

`pnpm typecheck` on both packages; **1541 node tests + 796 vitest, all passing.**

## If you'd rather have Option A

Close this. A is coherent — Run now means "start this routine once", and fan-out is what unattended running adds. The code even had a comment defending it. Nothing else in the branch depends on this.


suleimansh:
🤖 *automated*

**Dogfooded on a real daemon and a real repo — the fan-out is no longer a claim resting on unit tests.**

Rebuilt, restarted the daemon on this branch, and drove the dashboard through the browser. Three throwaway tickets in `tf-1334-dogfood`, concurrency set to 3, one click on **Run now → Plan tickets**.

## One click, three agents

```
[framework] auto PM: planning "2026-08-22_add-contributing-file.md" in .../tf-1334-dogfood
[framework] auto PM: planning "2026-08-22_greet-by-name.md"        in .../tf-1334-dogfood
[framework] auto PM: planning "2026-08-22_readme-usage-section.md" in .../tf-1334-dogfood
```

The card said the same thing: *"started 3 agents: planning …; planning …; planning …"*.

**Each ticket was claimed before its agent started** — three `.lock.md` files appeared together, then released as each run finished. That is the property the whole change rests on: without it two clicks plan the same ticket twice.

All three plans landed, and they are real work, not stubs — each read the repo (all three independently noted `index.js` is one line and there is no `package.json`) and carries its own Effort/Uncertainty spike:

| ticket | plan | effort / uncertainty |
|---|---|---|
| Add a CONTRIBUTING.md | 62 lines | 1 / 1 |
| index.js greets by name | 59 lines | 0 / 1 |
| README gets a Usage section | 47 lines | 0 / 0 |

Before this branch the same click started **one** agent and the other two tickets sat untouched.

## The tooltip tracks both the setting and the project

> Starts up to 3 agents in tf-1334-dogfood, one per open ticket, unattended.

Read live off `Concurrent agents: 3` and the picked project, not hardcoded.

## Project scoping, proven without spending anything

Four projects are registered. With the routine switched off — so no agent could start either way — two clicks produced exactly two log lines:

```
standing down for .../gemstack           — the planning routine is switched off
standing down for .../tf-1334-dogfood    — the planning routine is switched off
```

One line per click, naming the project the picker showed. Unscoped, that would have been four lines each. `gemstack` has 71 plannable tickets, so a scoping bug would have been loud.

## A fix I had not claimed

That stand-down is also a behaviour change worth naming: **the old Run now called `start()` directly and ignored the routine's checkbox**, so clicking a switched-off routine still spent an agent. It now stands down and says why.

## Unrelated, confirmed in passing

The same daemon printed `✅ Up to date (v0.7.0)` — #1639's `PACKAGE_NAME` change resolving against the real registry.

suleimansh:
🤖 *automated*

**The whole chain, run end to end on a throwaway repo — three tickets in, two pull requests with real code out.**

Follow-up to the fan-out dogfood above. This one answers "where does planning fit, and who implements the plans?" by running every step for real on `tf-1334-dogfood`.

## The chain

```
Update from GitHub     GitHub issues  →  tickets/*.md
       ↓
Plan tickets           tickets/*.md   →  tickets/*.plan.md          fans out
       ↓
Add quick-win work     tickets        →  entries on TODO_AGENTS.md  one agent
       ↓
Spin up agents…        queue entries  →  code + pull requests       fans out
```

Planning never touches code. Implementation is the drain, and it works from the **queue**, not from plans directly — a triage routine is what promotes a ticket onto the queue, and the entry links back to its ticket so the agent gets the ticket and its plan.

The two ends fan out; the middle cannot, because it rewrites the whole queue document.

## What actually ran

| step | agents | result |
|---|---|---|
| Plan tickets | **3** | 3 plans, one per ticket, each with its own lock |
| Add quick-win work | 1 | 2 entries queued, priority-ordered |
| Spin up agents… | **2** | 2 PRs |

```
auto PM: draining the queue entry "[index.js greets by name](...)"   in .../tf-1334-dogfood
auto PM: draining the queue entry "[Add a CONTRIBUTING.md](...)"     in .../tf-1334-dogfood
```

The other three registered projects stood down in the same sweep — *"the queue is empty, so there is nothing to drain"*.

## The output is real work

**[#8](https://github.com/suleimansh/tf-1334-dogfood/pull/8)** — `index.js` +2/-1, exactly what its plan specified:

```diff
-console.log("hello")
+const name = process.argv[2] || "world"
+console.log(`hello ${name}`)
```

**[#7](https://github.com/suleimansh/tf-1334-dogfood/pull/7)** — `CONTRIBUTING.md` +22, and it re-verified the claim its plan flagged: *"There is no `package.json`, so there are no dependencies to install and no build step."*

## Two problems this turned up, neither caused by this branch

**The triage implemented a ticket instead of queueing it.** `tf-triage-quick` carries a 14-line README change and opened its own PR ([#6](https://github.com/suleimansh/tf-1334-dogfood/pull/6)), so only 2 of 3 tickets reached the queue. Triage is meant to promote work, not do it — a prompt-behaviour issue worth its own ticket.

**A Run now can be silently eaten by the 30-minute cooldown.** Clicking the drain right after the triage gave:

```
standing down for .../tf-1334-dogfood — a run was started for this project a moment ago
```

Nothing started. It is reported on the card, but a user who clicks and looks away sees a button that did nothing. A daemon restart clears the cooldown, which is what unblocked this run.

**And a third instance of this branch's own theme.** The first triage click aborted on `Branch tf-triage-quick already exists`. The sweep releases a stale pinned branch before starting (`auto-pm.ts:1006`); the card's plain-start path never calls `releasePinned`, so Run now cannot clear it and just dies — $0.16 for a turn that read nothing. Same shape as the asymmetry this PR fixes for the fan-out routines, still present for the pinned ones.

Issue #1643, opened 2026-08-22
Run now on a pinned routine dies on its own stale branch: the card never releases it


suleimansh:
🤖 *automated*

Found while dogfooding the rotation end to end (#1204, evidence in #1640).

**What happened.** `Run now` on `Add quick-win work to AI Queue` started an agent that read nothing and aborted:

```
✗ Branch tf-triage-quick already exists — aborting
    $ git branch -a --list '*tf-triage-quick*'
        tf-triage-quick
        remotes/origin/tf-triage-quick
  No tickets were read, no branch was created, and TODO_AGENTS.md was not modified.
```

Cost: **$0.16** for a turn that did no work. It repeats on every click until someone deletes the branch by hand — which is what unblocked the dogfood.

**The cause.** Both triage jobs declare a `pinnedBranch` (`auto-pm.ts:355,362`), and a pinned job aborts itself when that branch already exists (#1293). The sweep handles this: it releases a stale pinned branch before each start —

```ts
// auto-pm.ts:1006
if (item.pinnedBranch) await deps.releasePinned?.(project, item.pinnedBranch).catch(() => undefined)
```

`releasePinned` is called **only** from inside that sweep loop. The card's `Run now` for an ordinary rotation job goes down a different path — a plain `start()` in `RoutineWork.tsx` — which never calls it. So the daemon can recover from its own leftover branch and the button cannot.

The branch that blocked us had **zero commits unique to `main`** and no pull request had ever existed. Nothing was protecting it; nothing was there to release it either.

**Same family as #1640.** That one was: the sweep fans out, `Run now` did not. This one is: the sweep releases a stale pin, `Run now` does not. The general shape is that a card-fired routine reaches less of the sweep's preparation than the sweep does, and the difference is invisible until a click dies.

**Fix.** Have the plain-start path release a pinned branch first, the same way the sweep does — either by routing a pinned job's `Run now` through the narrowed sweep (as #1640 does for the fan-out jobs), or by calling the same seam before `start()`. The first keeps one code path and is the direction #1640 already points.

suleimansh:
🤖 *automated*

Reproduced on `main` after #1640/#1644/#1645 merged. A second quick-triage Run now, 21 minutes after the first, aborted in 19 seconds:

> `tf-triage-quick` already exists, so triage is already pending. It's checked out in another worktree … I made no changes: no branch created, no `setSessionName()` call, and TODO_AGENTS.md on tf-data is untouched.

[...]

PR #1644, opened 2026-08-22
A triage queues work and never does it (#1641)

### LINKED ISSUE #1641: A triage routine implements a ticket instead of queueing it
🤖 *automated*

Found while dogfooding the whole rotation on a throwaway repo (#1204, evidence in #1640).

**What happened.** `Add quick-win work to AI Queue` was fired on a repo with three open, planned tickets. It queued two of them — and **implemented the third itself**, committing a 14-line `README.md` change onto its own `tf-triage-quick` branch and opening [a pull request](https://github.com/suleimansh/tf-1334-dogfood/pull/6) for it.

The ticket it consumed (`README gets a Usage section`) never reached `TODO_AGENTS.md`. The queue entry it *did* write for another ticket even refers back to the work it had just done:

> **Must also update the `## Usage` output block in `README.md`** — that section landed via `tf-triage-quick` and this change invalidates it.

**Why it is wrong.** Triage promotes work; the drain does it. The split is what makes concurrency safe: a triage rewrites the shared queue document and so must stay one agent, while the drain fans out one agent per entry. A triage that also writes code:

- spends a drain's work in the one-agent lane, so the fan-out never gets it
- skips the queue, which is where a human can veto an item before an agent implements it
- pins its work to `tf-triage-quick`, a branch reused by every future triage — the next run then aborts on "branch already exists"

**Not a code bug.** Nothing in `auto-pm.ts` asks for this; the preset's prompt allowed it. The fix is in `prompts/`, and the check is that a triage's own branch carries only `TODO_AGENTS.md`.

**Repro.** Throwaway repo, three planned tickets, `Run now` on `Add quick-win work to AI Queue`.

suleimansh:
🤖 *automated*

Fixes #1641.

**What changed.** One new file, `prompts/triage_scope.md`, beside the queue format it protects:

> You only queue work, you never do it: the only file you change is `TODO_AGENTS.md`. Do not implement a ticket, however small its plan — no code changes, no pull request for it. Every ticket you pick goes on the queue, where a human can still veto it before an agent implements it.

The preset catalog appends it to both triage rows (`triage(PRESETS_TRIAGE_QUICK)` / `triage(PRESETS_TRIAGE_CONSENSUAL)`), so `triage_quick.md` and `triage_consensual.md` carry only what differs between them. A test asserts each preset's render *ends with* that file's text, so neither row can drop it. `prompts/SPEC.md` and `preset-catalog.SPEC.md` state the rule and its rationale.

**Why a prompt change and not a code one.** Nothing in `auto-pm.ts` asked the triage to write code — the prompt said "choose tickets to work on" and "add tickets to `TODO_AGENTS.md`" and stopped there. On `tf-1334-dogfood` that was enough for the agent to queue two tickets and implement the third (a 14-line README change, with its own PR), so it never reached the queue.

**Why the reason is in the prompt, not only the rule.** A bare "don't" loses to a bare "this is trivially small". The paragraph says what the queue is for, so the agent has something to weigh a three-line plan against.

**What the split protects.** A triage rewrites the whole queue document, so it runs as one agent; the drain runs one agent per entry. A ticket built inside a triage takes the single-agent lane, skips the fan-out, and skips the human veto — and pins its work to `tf-triage-quick`, which is reused by every later triage (#1643 is the branch-reuse side of that).

**Verified.** `gen-prompts` compiles 25 prompts, `preset-catalog.test` + `auto-pm.test` 108/108; with one triage row un-composed, exactly the new test fails. `tsc --noEmit` clean. Not re-dogfooded with a live agent — the check for that is a triage whose branch carries only `TODO_AGENTS.md`, and it costs a real run.


suleimansh:
Can we DRY this up?

suleimansh:
🤖 *automated*

Done in 99400c3a. The paragraph now lives once, as `prompts/triage_scope.md`, and the catalog appends it to both triage rows — the two preset files keep only what differs between them (what counts as a pick, and their session name).

It stops short of folding the session-name tail too. That line differs by name and the paramless presets render verbatim, so sharing it would mean adding substitution to `definePreset` — a wider change than this fix.

A new test asserts each triage's render ends with the shared file's text; un-composing one row fails exactly that test.

suleimansh:
🤖 *automated*

Dogfooded on `main` after merge: the quick triage queued all three planned tickets (priority-ordered, the doubtful one flagged for a human glance), and its `tf-triage-quick` branch has zero commits and zero files against `main`, no PR. Full chain with the clock on #1204: https://github.com/framework/the-framework/issues/1204#issuecomment-5382936130

Issue #1210, opened 2026-07-25
`Trigger routine now` next to `[ ] Auto-run`


brillout:
If I don't want to wait x minutes for auto-run to be triggered, I want to be able to click on a button `Trigger routine now`.

Required for demo video.

PR #1645, opened 2026-08-22
A routine's Run now is not held back by the sweep's cooldown (#1642)

### LINKED ISSUE #1642: Run now can be swallowed by the 30-minute cooldown, and the button looks like it did nothing
🤖 *automated*

Found while dogfooding the rotation end to end (#1204, evidence in #1640).

**What happened.** A triage `Run now` was fired, and a drain `Run now` clicked a minute later. The second click started nothing:

```
[framework] auto PM: standing down for .../tf-1334-dogfood — a run was started for this project a moment ago
```

That is `DEFAULT_AUTO_PM_COOLDOWN_MS` — **30 minutes** (`auto-pm.ts:29`) — doing its job. The cooldown exists so the unattended sweep does not stack runs on one project every tick.

**Why it is a problem for a click.** The cooldown is a policy for work *nobody asked for*. A click is someone asking. Today the same gate applies to both, so for half an hour after any run, a routine's `Run now` is a button that cannot start anything.

The outcome line does say so, under `Concurrent agents`. But it is one line of small grey text below the fold of the card, it appears after the click rather than before, and the button itself gives no hint — it is not disabled and its tooltip still promises "Starts up to N agents". A user who clicks and looks away sees nothing happen and no reason why.

Restarting the daemon clears it, which is what unblocked the dogfood — a workaround nobody should need to know.

**Worth deciding, not obviously one way:**

- **Option A** — a click ignores the cooldown, the way `onDemand` already outranks the master switch (`auto-pm.ts:637` doc): "the click is the consent the preference exists to record". Consistent with how every other gate treats an explicit ask.
- **Option B** — keep the gate, but say so *before* the click: disable the button and let the tooltip read "on cooldown until HH:MM" instead of promising agents it cannot start.

A is the smaller change and matches the existing precedent. B is safer if the cooldown is also protecting quota rather than only pacing the sweep.

**Related:** the same "a click reaches less than the sweep does" shape as #1640 and the pinned-branch gap filed alongside this.

suleimansh:
🤖 *automated*

Fixes #1642 — option A.

**What changed.** `autoPmDecision` takes `onDemand`, and skips the 30-minute cooldown when it is set. The sweep passes the flag through from `tick({ onDemand })`. One condition, one flag, no new state.

**Why A.** The cooldown paces work nobody asked for. A click is someone asking — the same reading that already lets an on-demand sweep outrank the master switch (`auto-pm.ts` `tick` doc, #1210). Before this, a Run now fired within half an hour of any run on that project stood down with *"a run was started for this project a moment ago"* — reported in small grey text under the fold of the card, after the click, while the button stayed enabled and its tooltip still promised agents. The only way past it was restarting the daemon.

**What is deliberately not waived: the concurrency cap.** The #685 double-tick test names the cooldown as what stops a tick that lands before the spawn registers from doubling up. For a click that guard is the cap alone, and it holds: `start()` resolves after the spawn, ticks are serialized by `sweeping`, so a second click counts the first agent. The new decision test pins both halves — through the cooldown when asked, still refused at the cap.

**Spec.** `auto-pm.SPEC.md` said *"Run now skips only the master switch"*. It now says it skips the master switch and the cooldown, and why the cap stays.

**Verified.** `auto-pm.test`: 91/91. With the bypass removed from the gate, exactly the two #1642 tests fail and nothing else; restored, green. `tsc --noEmit` clean.

Independent of #1640 (different lines of `auto-pm.ts`; branched from `main`).


suleimansh:
🤖 *automated*

Dogfooded on `main` after merge: the drain's Run now clicked 17 minutes after a sweep-started run went through — `started 2 agents`, claims committed first, two PRs open within two minutes and merged by CI watch — where yesterday the same click stood down with *"a run was started for this project a moment ago"*. Full chain on #1204: https://github.com/framework/the-framework/issues/1204#issuecomment-5382936130

PR #1652, opened 2026-08-22
The routine card remembers which project Run now targets (#1647)

### LINKED ISSUE #1647: The routine card's project picker resets to the first project on every navigation
🤖 *automated*

Found dogfooding #1640/#1644/#1645 on `main` (chain write-up: https://github.com/framework/the-framework/issues/1204#issuecomment-5382936130).

**What happened.** On the Overview, the *Routine work* card has a `Run in` picker that scopes the per-project routines (#1640). I set it to `tf-1334-dogfood`, clicked Run now (correctly scoped), opened the run, came back to the Overview — and the picker read `gemstack` again, the first registered project. The next click fired a quick triage on **gemstack**, the real repo. It was stopped at 40 seconds with nothing committed, but it had already pushed an empty `tf-triage-quick` to the repo's origin — which is #1643's trap, now armed for the next gemstack triage until someone deletes the branch.

**Why it matters.** The picker decides which repo spends quota and gets branches pushed, and it silently forgets the choice on the most common navigation there is (open a run, come back). The first project in the registry is the user's main repo, so the failure mode is always "the wrong click lands on the real codebase". The tooltip does name the project (*"Starts one agent in gemstack …"*), which is how it is catchable — but only by reading the tooltip before every click.

**Fix shape.** Remember the pick across navigations — the session store the rest of the dashboard already uses is enough; persisting it as a preference would also make it survive reloads. Defaulting to the project whose page the user came from would cover the open-run-and-return case on its own.

**Repro.** Overview → set `Run in` to any project but the first → open any run from the sidebar → Overview.

suleimansh:
🤖 *automated*

Fixes #1647.

## What changed

The Routine work card's `Run in` pick was component state, so the card forgot it whenever it unmounted — open a run from the sidebar, come back — and the picker read the first registered project again: the user's real repo. The next click fired a triage there and pushed an empty `tf-triage-quick` to its origin.

The pick is now a preference, **`autoPmProject`** (a project id), beside `autoPmOptOut` and `autoPmConcurrency` — the two settings it already sits with on the card:

- `src/registry.ts`: the field on `Preferences`, sanitized like the opt-out names — a bounded free-form string, not checked against the project list, since the card validates it against the projects it shows and a project removed since should fall back there rather than be erased on read. Empty is dropped.
- `dashboard/components/RoutineWork.tsx`: no `picked` state; the picker writes `updatePreferences({ autoPmProject })` and the card reads `preferences.autoPmProject` back on every render (the store is optimistic, so the `<select>` follows the click at once). The existing validation stays: a remembered id that names no registered project falls back to the first.

So the pick holds across navigations, reloads (the hard-refresh after every daemon restart included) and tabs, the way the boxes beside it do.

## Why a preference rather than session storage

The issue offered both. The pick decides which repo spends quota and gets branches pushed; a reload is exactly the moment it used to snap back to the real repo, and the dashboard keeps the user's settings in the daemon's file with no browser storage by design (`registry.SPEC.md`). One field in the existing shape, no second store. "Default to the project whose page the user came from" is not done — it is a different feature, and with the pick remembered the open-run-and-return case no longer needs it.

## Spec changes

- `RoutineWork.SPEC.md` — a new flow sentence: the project a Run now targets is picked once and kept as a setting, so leaving the Overview and coming back, or reloading, does not put the first project back; a remembered project no longer registered falls back to the first.
- `FEATURES-SPEC.md` — "Run now" on a routine, in a picked project the card remembers across navigations and reloads.
- `RoutineWork.test.SPEC.md` updated for the new tests.

## Verified

- New tests: `registry.test.ts` — the preference round-trips trimmed, and an empty or non-string value is dropped. `RoutineWork.test.tsx` — a saved pick is what the picker shows and a change writes the setting; a remembered project that is gone falls back to the first. Two existing picker tests now set the preference instead of firing a change on the select (the mock store cannot re-render on a write; the write itself is asserted in the new test).
- Broke it on purpose (card reads no preference; sanitizer drops the key): the registry test and three card tests failed, everything else green. Restored from the commit.
- `pnpm typecheck` exit 0; `pnpm test` with the bundle built: node 1548/1548, dashboard 799/799.
- Not dogfooded; the behaviour is one preference round-trip the unit tests cover end to end. The repro in the issue (Overview → pick a non-first project → open a run → Overview) is the check to do on the rebuilt dashboard.


suleimansh:
🤖 *automated*

Checked on the real dashboard, 2026-08-23 ~03:45 IDT, from a throwaway build of `main` + this branch + #1651 (no merge; the :4200 daemon was restarted from that build and put back on `main`'s build afterwards).

The issue's repro, step by step:

1. Overview → `Run in` set to `tf-1334-dogfood` (not the first project). `~/.the-framework.json` gained `"autoPmProject": "tf-1334-dogfood-f8nvbw"` immediately.
2. Opened a run from the sidebar (the 02:20 triage run's page), then the sidebar's Overview. The card remounted (fresh element ids) and the picker read **`tf-1334-dogfood`** — before this branch it read `gemstack` here.
3. `Cmd+R` on the Overview: picker still **`tf-1334-dogfood`**.

Then set it back to `gemstack` so the preference reads as it did before the check.

Issue #1646, opened 2026-08-22
The drain fans out one agent short: a live-agent count of 1 with no agent visible


suleimansh:
🤖 *automated*

Found dogfooding #1640/#1644/#1645 on `main` (chain write-up: https://github.com/framework/the-framework/issues/1204#issuecomment-5382936130).

**What happened.** `Concurrent agents: 3`, three entries on the queue, no agent running on the project — the Agents panel said *"No agents working right now"*, the last run on the project had been DONE for ten minutes. The drain's Run now reported:

```
tf-1334-dogfood: started 2 agents: draining the queue entry "[Add a root `.gitignore` …"; draining the queue entry "[Add a `package.json` …"
```

Two of three. The third entry sat on the queue.

**Why two.** The fan-out is `picks = open.slice(0, concurrency - activeAgents)` (`auto-pm.ts:922`), so `activeAgents` read **1** at 01:20:55. `activeAgentCount` (`daemon-runtime.ts:840`) is live pids on the project plus the `starting` set:

```ts
for (const [key, pid] of activeAgents) { if (keyBelongsTo(key, targetProjectId) && isPidAlive(pid)) live++ }
return live + [...starting].filter(key => keyBelongsTo(key, targetProjectId)).length
```

The pid half is re-checked against the OS, so a dead process cannot hold a slot. The `starting` half is not: a key that is added on spawn and never removed — a spawn path that throws between the two, or a refused start that skips the cleanup — costs the project one slot for the rest of the daemon's life, invisibly. Not diagnosed which it was; by 01:27 no process of the earlier runs was alive, and the daemon does not log the count.

**Timeline on that project:** 01:03 three planners via the sweep (all DONE by 01:06, locks released); 01:06 one triage via the card's plain start (DONE 01:10); 01:20 the drain click → count 1.

**Fix shape.** Either log the breakdown on every stand-down/fan-out decision (`live=N starting=M`), so the next occurrence names itself — or make `starting` self-healing the way `activeAgents` is (a key older than the spawn timeout is dropped). Both are small; the first is what would have made this an answer instead of a question.

**Repro.** Not reliable yet. Three sweep-started runs, one card-started run, then a drain click on a project the panel shows idle.

suleimansh:
🤖 *automated*

Diagnosed from the daemon's own log of that session and the rig's `tf-data` history — the slot was a real live process, not a leaked `starting` key.

[...]

Issue #1601, opened 2026-08-19
Cloud runs land work on claude.ai's own branch, not the designated tf-agent branch — and the armed PR never opens


suleimansh:
🤖 *agent issue*

Live evidence from the post-#1595 smoke run (session `2026-08-19T16-13-35-698Z`, trivial one-file prompt, handoff armed for push + draft PR):

- The local wrapper teleported to a claude.ai cloud session and self-ended (`handoff skipped: the session committed nothing` — correct for the wrapper).
- The cloud session did the work cleanly (one commit, `81eb23a`, exactly one file, zero framework-data leakage — the trimmed data-branch protocol held), **but pushed it to `claude/create-test24-md-i40ifl`**, the branch the claude.ai harness designates, not the TF-designated `tf-agent-2026-08-19T16-13-35-698Z` — which now sits empty on origin at main's tip. The agent flagged the conflict itself: *"the session's branch requirement overrides that"* — two harnesses both claiming branch authority, cloud wins.
- Consequence: the run's dashboard row says "Nothing committed — no PR to open" while the work exists on a branch TF doesn't know about; the armed draft PR never opened (the agent also read the prompt's "nothing else" as no-PR-wanted, so intent-vs-armed-handoff is a second, smaller seam here).

Fallout: any TF surface keyed to the designated branch (branches/ checkout links, CI watch, merge handoff, stale-branch sweeps) is blind to cloud work, and empty `tf-agent-*` branches accrete on origin.

Direction candidates, deliberately not chosen here: teach the wrapper to adopt the cloud session's actual branch at settle (rename or track), or stop pre-pushing/designating a `tf-agent` branch for cloud runs entirely and record the claude/ branch as the session branch. Related: #1328 (extension-driven web runs) would sidestep the second harness altogether.

suleimansh:
🤖 *agent comment*

Housekeeping: the two branches from this smoke run are now deleted on origin (`claude/create-test24-md-i40ifl` and the empty `tf-agent-2026-08-19T16-13-35-698Z`). Commit `81eb23a` (one file, test24.md) is therefore unreferenced — the evidence for this issue is the description above.

suleimansh:
🤖 *agent comment*

Direction proposal after mapping the code. The description's two candidates turn out to be complementary, not alternatives — and neither alone works:

[...]

PR #1653, opened 2026-08-23
An agent that committed nothing leaves no branch behind (#1650)

### LINKED ISSUE #1650: A pinned routine's commitless branch is never released on a repo where the name never had a PR
🤖 *automated*

Residue of #1643, found while dogfooding #1649 (write-up: https://github.com/framework/the-framework/pull/1649#issuecomment-5383150770).

**What #1649 fixed.** A triage's Run now now goes through the sweep, so the stale-branch release (`releaseStalePinnedBranch`, #1293) runs before the start. On the rig that worked: the leftover `tf-triage-quick` was deleted and the agent proceeded.

**What it cannot fix.** The release is deliberately conservative (`src/stale-branch.ts`): it deletes the branch only when its name has PR history and no PR is open. A branch with **no PR history at all** comes back `unproven` and is kept — the seam cannot tell "an agent still working toward its handoff" from "a leftover that never had a PR".

The triage routines make that leftover every time they finish:

- `triage-quick` / `triage-consensual` write only `TODO_AGENTS.md`, which lives on `tf-data`, so they push there directly and **commit nothing on their own branch**.
- A commitless branch skips the handoff (`handoff skipped: the session committed nothing`) — no push, no PR.
- The run's worktree is reclaimed, but the branch `tf-triage-quick` stays, at `main`, with zero commits.
- The next firing (scheduled or clicked) finds it, the prompt's guard says "branch already exists — triage already pending", and the agent aborts in seconds (~$0.18 each time).

So on any repo where the name has ever had a PR (the rig: PR #6), the loop is self-healing since #1649. On a repo where it never had one — a fresh project, or any project after the triage prompt changed to queue-only (#1644), which is what took the PR away — it jams forever on the second triage, exactly as in #1643, and the only way out is still `git branch -D tf-triage-quick` by hand. Seen on gemstack's origin on 2026-08-23 (an empty `tf-triage-quick` pushed by a misfired click, deleted by hand).

**Options** — this is a data-safety policy, so it wants a human decision rather than a fix-by-default:

1. **Count "no commits past `main`" as stale.** Extend the seam: a branch with no PR history *and* no commits unique to its base holds nothing, so release it. Risk: an agent that started seconds ago and has not committed yet looks identical — mitigated in practice by `git branch -D` refusing while that agent's worktree has the branch checked out (the in-flight guard #1293 already relies on), but not on the remote copy.
2. **Have the triage delete its own branch when it ends commitless.** The routine knows it never needed the branch; the daemon's settle path already knows `handoffSkip === 'no-commits'`. No seam policy change; the leftover never exists. Does not help with branches other tools leave behind.
3. **Drop the pinned-branch guard for queue-only triages.** Since #1644 the triage changes one file on `tf-data`; two overlapping triages cannot corrupt each other the way two queue rewrites on one branch could. Then there is nothing to release. Biggest change to the prompts' contract (#1293's reason for the pin was duplicate triage, which #1644's queue-only rule makes cheap rather than harmful).

I lean to 2 as the smallest correct change, with 3 as the question worth asking Rom: whether the pin still buys anything after #1644.

suleimansh:
🤖 *automated*

Closes #1650.

## The bug, plainly

A triage routine writes only `TODO_AGENTS.md`, which lives on the data branch, so its own branch (`tf-triage-quick`) ends with zero commits. Two things then went wrong in teardown, both from the one rule *only what is on the remote may go*:

- it **pushed** the empty branch to origin to make the checkout removable — that is the empty `tf-triage-quick` I deleted from gemstack's origin by hand on 2026-08-23, and the rig's origin had five such refs (`tf-triage-quick`, two `tf-plan-*`, two `tf-agent-*`), all at the install commit;
- it **kept** the branch, and the next triage found the name taken and stood down in seconds as "triage already pending" (#1643 again).

The stale-branch release (#1293, reached by Run now since #1649) could not fix it: it deletes only a branch whose name has PR history, and a queue-only triage (#1644) never opens a PR, so the leftover is `unproven` forever.

## What changes

This is issue #1650's option 2, placed where the leftover is actually made rather than in the auto-PM settle path: `removeProjectWorktree` is the one implementation behind teardown, the sweep and the dashboard's Remove button, so it is the one place that decides.

- **A checkout whose run branch holds nothing the remote lacks goes without a push, and its branch goes with it.** "Holds nothing" is: clean tree, and the tip is reachable from some remote-tracking branch *other than the branch's own* (`origin/main`, typically). That is the #1601 web-run carve-out generalised — the rule is already satisfied before any push, and the branch is by construction not the last copy of anything.
- Two guards, both from the dogfood (below):
  - only a branch the framework minted (`tf-…`, never `tf-data`) is ever deleted. A leftover checkout on any other branch keeps it whatever it holds — one was found sitting on the rig's `main`.
  - a branch's own remote copy does not count as "the remote has it": a pushed run branch with a PR contains its own tip and is exactly the branch that must stay.
- A clean checkout carrying a commit origin has not seen still takes the ordinary commit → push → keep-the-branch path.
- `attachWorktree` recreates a missing branch from HEAD, so *Continue* on such an agent still works instead of falling to "starting a new one". The only branch the framework deletes held nothing past a commit the remote had, so HEAD is where its work was.
- The sweep's line and the teardown say when a branch went (`removed the worktree for session … and its branch tf-triage-quick: the branch held nothing the remote lacks`), instead of the sweep's fixed "the branch and the session are kept".
- No policy change to `stale-branch.ts`; a comment there says why the commitless case never reaches it now.

Reads the local remote-tracking refs only, which are only ever behind the remote: a tip they do not cover yet answers false and falls back to the push — the status quo, never worse. Pre-existing empty copies already on origin are not touched.

## SPEC changes

Rebased past `0646a4ad` ("wipe"), which deleted every per-file `SPEC.md`; the six spec edits this PR carried are dropped with them. For the record, the business-logic change they stated — in case the specs come back:

- *worktrees*: a checkout whose run branch holds nothing the remote lacks — a clean tree whose tip is already inside some *other* remote branch — goes without a push, and its branch is deleted with it. A branch's own remote copy does not count; a clean checkout carrying an unpushed commit is pushed first and keeps its branch; only names the framework minted are ever deleted, a leftover checkout on any other branch (the user's `main`) keeps it.
- *store/worktree*: continuing an agent whose branch no longer exists recreates it from the repo's current commit; any other reason git cannot attach still fails the continue.
- *stale-branch*: a pinned branch with no commits of its own never needs the release; its checkout's removal deletes it.
- *merged-worktrees*: the removal announcement names the branch when it went.

**`FEATURES-SPEC.md`** (kept by the wipe) — one line under *Handoff and what lands in git*.

## Verification

- `pnpm test`: 1556 node green; dashboard 799 green on two runs (one unrelated flake on the first run, in a suite this backend-only diff does not touch).
- First CI run failed on `daemon.test.ts` (`ENOTEMPTY` removing the temp `.git`): the test's second run commits nothing and its tip is on origin under the first run's branch, so it now takes the no-push path and teardown's `git branch -D` was still writing when the test removed the repo. The test now waits for — and asserts — the branch's absence and that nothing was pushed; fails on exactly that assertion against main's `worktrees.ts`.
- New real-git tests: the empty run branch goes unpushed and deleted; a clean checkout with an unpushed commit is pushed and keeps its branch; a run branch pushed under its own name stays; a leftover checkout on a non-framework branch keeps it; `attachWorktree` recreating a gone branch and still refusing one git will not attach; `deleteBranch`; the sweep line naming the branch.
- Broken on purpose: with `src/worktrees.ts` reverted, exactly the one new removal test fails and the guard tests still pass.

## Dogfood on the rig (2026-08-23 14:36–14:40 IDT, Opus, from the first commit's build)

Rig state before: a commitless `tf-triage-quick` at `main`, locally and on origin. The routine card was pointed at the rig, quick-win triage ticked and its Run now clicked.

- The sweep released the stale `tf-triage-quick` first (#1649's path — the name has PR #6 history).
- The agent ran ~2 min, wrote the queue to `tf-data` (`e3c8812 Queue help-flag quick win; note merged sibling tickets`), ended `handoffSkip: no-commits`.
- Teardown: worktree gone, **`tf-triage-quick` gone locally, nothing pushed** — `git ls-remote --heads origin | grep triage` is empty. The next triage click there finds no branch.

What the same run surfaced, and the second commit fixes: the startup sweep also reclaimed the leftover checkout of the #1646 ghost-process session (`2026-08-22T22-06-41-065Z`), which was sitting on **`main`** — the first cut ran `git branch -D main`, stopped only by the primary checkout having it out. Hence the minted-name guard. The own-remote-copy exclusion came from an existing test (`a checkout already pushed is not re-pushed, and still goes`) failing against the first cut. The final code is not re-dogfooded; the rig case (`tf-` name, no remote copy) sits inside both guards, and both are covered against real git.

Residue, not in scope: the agent reached `tf-triage-quick` by `checkout -b` rather than the rename, so its run-id branch `tf-agent-2026-08-23T11-37-03-643Z` stays behind, empty, as seven `tf-agent-*` branches from yesterday's runs do on the rig. Same class (an empty minted branch), different name from the one the checkout is on; a follow-up if wanted.

## Open question for @brillout (issue option 3)

After #1644 the triages change one file on `tf-data`, so two overlapping triages cannot corrupt each other the way two queue rewrites on one branch could. Does the pinned branch still buy anything for the queue-only routines, or could the pin go for them entirely? This PR makes the pin harmless either way; that is a separate, smaller change if you want it.

### DIFF
--- a/FEATURES-SPEC.md
+++ b/FEATURES-SPEC.md
@@ -129,6 +129,7 @@ happens while nobody is at the keyboard.
 - CI watch: merge a PR once its checks pass
 - CI watch: one fix agent per red head commit, max two attempts
 - Reclaim the checkout of an agent whose work is on the remote — never by publishing what a `handoff: local` agent refused to
+- An agent that committed nothing leaves no branch behind: its empty branch goes with its checkout, never pushed, so a pinned routine is not stood down by its own last run
 - Release a pinned routine branch left behind by a closed PR — before the schedule fires the routine, and before its "Run now" does
 - The agent drains its own TODO backlog, one entry per turn
 
--- a/packages/the-framework/src/branch-names.ts
+++ b/packages/the-framework/src/branch-names.ts
@@ -60,6 +60,16 @@ export function agentIdFromWorktreeDir(name: string): string {
 }
 
 /**
+ * Whether a branch is one the framework minted for a run (#1650): the run-id spelling, a
+ * session-named `tf-<name>`, or the legacy slashed form — and never the data branch, which
+ * shares the prefix. The only branches the framework may delete on its own.
+ */
+export function isRunBranch(name: string): boolean {
+  if (name === DATA_BRANCH) return false
+  return name.startsWith(AGENT_BRANCH_PREFIX) || name.startsWith(LEGACY_AGENT_BRANCH_PREFIX)
+}
+
+/**
  * Whether a name under `branches/` is a framework checkout directory. The same directory also
  * holds the rename links (#1589) and possibly a user's own entries, and only names the framework
  * mints — the run branch spelling — are checkouts.
--- a/packages/the-framework/src/daemon-runtime.ts
+++ b/packages/the-framework/src/daemon-runtime.ts
@@ -597,6 +597,8 @@ export function createProjectRuntime({ cwd, env, binPath, retryDelayMs, driverPr
         // never depends on the caller having one.
         const outcome = await removeProjectWorktree(projectCwd, agentId ?? agentIdFromWorktreeDir(basename(worktree)))
         if (!outcome.ok) console.log(`[framework] keeping worktree ${worktree}: ${outcome.error}`)
+        else if (outcome.branchDeleted)
+          console.log(`[framework] removed worktree ${worktree} and its branch ${outcome.branchDeleted}: the branch held nothing the remote lacks`)
       } catch {
         // A worktree we could not retire is a worktree left on disk, which is the safe direction.
       }
--- a/packages/the-framework/src/daemon.test.ts
+++ b/packages/the-framework/src/daemon.test.ts
@@ -371,7 +371,27 @@ fs.appendFileSync(${JSON.stringify(join(cwd, 'started.log'))}, agentId + '\\n')
     // accumulated per failed session until a human noticed.
     const failedId = await runWith('failed', 2)
     assert.equal(await archived(failedId), true, "a failed run's history is copied too")
-    assert.equal(await worktreeGone(failedId), true, 'and its checkout goes too, since the remote has its branch')
+    assert.equal(await worktreeGone(failedId), true, 'and its checkout goes too')
+    // This run committed nothing, and its tip — the init commit — is already on origin under the
+    // first run's branch. So nothing is pushed and the branch goes with the checkout (#1650): the
+    // branch's absence is the last thing teardown does, so it is what the test waits on before
+    // pulling the repo out from under the daemon.
+    for (let i = 0; i < 600; i++) {
+      const gone = await git(['show-ref', '--verify', '--quiet', `refs/heads/tf-agent-${failedId}`], cwd).then(
+        () => false,
+        () => true,
+      )
+      if (gone) break
+      await new Promise(r => setTimeout(r, 20))
+    }
+    await assert.rejects(
+      () => git(['show-ref', '--verify', '--quiet', `refs/heads/tf-agent-${failedId}`], cwd),
+      'a run that committed nothing leaves no branch behind',
+    )
+    await assert.rejects(
+      () => git(['rev-parse', '--verify', `refs/remotes/origin/tf-agent-${failedId}`], cwd),
+      'and nothing of it was pushed',
+    )
 
     ac.abort()
     await done
--- a/packages/the-framework/src/merged-worktrees.test.ts
+++ b/packages/the-framework/src/merged-worktrees.test.ts
@@ -106,7 +106,7 @@ test('the sweep says what it removed and what it kept, per project (#1036)', asy
   const lines: string[] = []
   const results: Record<string, MergedSweepResult> = {
     '/a': { removed: [{ agentId: 'r1' }], failed: [] },
-    '/b': { removed: [{ agentId: 'r2' }], failed: [{ agentId: 'r3', error: 'not on the remote' }] },
+    '/b': { removed: [{ agentId: 'r2', branchDeleted: 'tf-triage-quick' }], failed: [{ agentId: 'r3', error: 'not on the remote' }] },
   }
   const sweep = startMergedWorktreeSweep({
     projects: async () => [{ path: '/a' }, { path: '/b' }],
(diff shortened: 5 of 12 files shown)


brillout:
@suleimansh (AI) so the lock-via-branch mechanism isn't needed anymore? Why?

suleimansh:
🤖 *curated*

The lock is still wanted — two triages at once is waste. The branch just stopped being one.

It guarded against two triages running at once, on one machine or across machines via origin. Since #1644 a triage writes only the queue on `tf-data` and commits nothing on its branch, so the branch never reaches origin — the cross-machine half is already gone. What's left is a local branch name the agent checks from the prompt; the daemon never checks whether a triage is actually running. That branch-name lock took four fixes to stop jamming (#1293, #1643, #1649, #1653), and each false abort still spends an agent.

Two overlapping triages today would cost two agents reading the same tickets and, at worst, a duplicated queue line — not a broken branch.

Proposal: keep the rule, move it to the daemon — it already knows its live runs (#1651), so "don't fire a triage while one is running" needs no branch and no aborting agent. Then the prompt's abort line and the stale-branch release go.

Want that as a ticket?

brillout:
I see, how about a new `routines/` in the `tf-data` branch, e.g. `routines/triage-quick-win.md` saves the quick-triage prompt (bonus: in the future we can allow the user to customize it) => we can use `routines/triage-quick-win.lock.md`

suleimansh:
🤖 *curated*

Yes — that's the ticket lock (#1420/#1583) applied to routines, and it's cross-machine, which the branch never was. Two things to settle:

1. Release. A triage never opens a PR, so no merge lifts the lock. The daemon that started the run drops it when the run ends — the same release that already frees a `no-commits` ticket claim.

2. Expiry. A machine that dies mid-triage would leave the lock forever, for every machine. Not a fixed deadline — a real run can outlive any number we pick. The owning daemon refreshes a `last-seen` time while the run is alive; a lock not refreshed for ~10 min counts as dead. Worst case of a wrong call is one duplicate triage, not lost data.

I'd split the prompt part out: `routines/<name>.md` as the user-editable prompt is its own change (the presets ship in the package today — which copy wins needs deciding). The lock lands first, keyed by the preset name: `routines/triage-quick.lock.md`, matching `tf-triage-quick`.

Filing both as tickets unless you object.

suleimansh:
🤖 *automated*

Filed: #1659 (the lock, lands first) and #1660 (the prompt on `tf-data`).

PR #1656, opened 2026-08-23
A branches/ directory that is not a git worktree is left alone, never acted on as the main checkout (#1654)

### LINKED ISSUE #1654: A branches/ directory that is no longer a git worktree is judged — committed, pushed, branch-linked — as the main checkout
🤖 *automated*

Found while dogfooding #1653 on the rig (`tf-1334-dogfood`), 2026-08-23.

## What happened

`.the-framework/branches/tf-agent-2026-08-22T22-06-41-065Z/` was not a git worktree any more — I had removed the worktree by hand at 02:17 (#1649 dogfood), and the 02:15 shutdown had written a "failed to start" stub into the path afterwards, so the directory held only `.the-framework/agent.json` + `events.jsonl`. Git has no idea the directory is special, so **every git command run inside it resolves to the enclosing repo** — the rig's primary checkout:

```
$ git -C .the-framework/branches/tf-agent-2026-08-22T22-06-41-065Z rev-parse --show-toplevel
/Users/suleiman/Projects/tf-1334-dogfood
$ git -C … rev-parse --abbrev-ref HEAD
main
```

The cleanup sweep lists `branches/` by directory (`listWorktreeDirs`, `src/store/agent-store.ts:803`) and then asks git about each path, so on that directory it concluded the session's branch was **`main`** and ran the ordinary recoverability rule against the user's primary checkout:

- `commitPendingWork` — `git add -A && commit` **in the rig's main checkout**. The tree happened to be clean, so nothing landed. On a dirty main it would have been the #1638 catch-all commit, on the user's branch.
- `pushAgentBranch(cwd, 'main')` — tried to push the user's `main` to origin. It failed only because `origin/main` was ahead: `kept the worktree for session 2026-08-22T22-06-41-065Z: main is not on the remote (error: failed to push some refs …)`, logged every pass since 02:15.
- Earlier, the rename-link step read the same `main` and left a `branches/main -> tf-agent-2026-08-22T22-06-41-065Z` symlink (the #1589 links directory held a link named after the user's default branch).
- With #1653's first cut, the "branch holds nothing" rule read `main` as empty (tip on the remote) and ran `git branch -D main` — stopped only by the primary checkout having it out. #1653 now deletes only names the framework minted, but that guard is downstream of the misread.

Residue removed by hand at 14:46 (the stub directory and the `main` link); the session's archive in `tf-data` is intact.

## Why it matters

Every consumer of a `branches/<run>` path assumes the path is a worktree. When it is not — a hand-removed worktree, a half-failed `git worktree add`, a stub written after teardown, a disk restore — the assumption silently turns "the agent's checkout" into "the user's checkout", and the rules built for an agent's throwaway branch (commit everything, push, delete when empty) run against the user's `main`.

## What to do

Refuse to treat a directory as a checkout unless git agrees it is one: `git rev-parse --show-toplevel` (or `--git-dir`) must resolve to that path, or `git worktree list --porcelain` must name it. A directory that fails that test is residue — report it ("not a worktree; left alone") and never run commit/push/branch commands through it. `currentBranch` (`src/store/worktree.ts:193`) is the narrowest place; the sweep's row builder is the broadest. The stub write at shutdown that created the directory in the first place (a meta written into a path whose worktree is gone) is the second half.

Not started.

suleimansh:
🤖 *automated*

Closes #1654.

## The bug, plainly

Git answers inside *any* directory of a repository. A directory under `.the-framework/branches/` that is no longer a worktree — on the rig: a checkout I removed by hand, then a failed-start marker written into the path at shutdown — therefore answers for the **enclosing repo**, i.e. the user's own checkout. Every consumer took that answer as the run's:

- the cleanup sweep read the run's branch as `main`, ran the recoverability rule on the user's checkout (`git add -A` + commit — the tree happened to be clean), and tried to push the user's `main` to origin; it failed only because origin was ahead, and logged that every pass for 12 hours
- the rename pass linked `branches/main -> tf-agent-…`
- #1653's first cut judged `main` as "holds nothing" and ran `git branch -D main`, stopped only by the primary checkout having it out

## What changes

One question tells a checkout from residue: **is git's top level this very directory** (`git rev-parse --show-toplevel`, both sides realpath'd). `isWorktreeRoot` in the store, and `worktreeBranch` = the branch read guarded by it. Applied at every place a `branches/<run>` path is acted on:

- `removeProjectWorktree` (teardown, the sweep, the dashboard's Remove button) refuses before any git runs: `session …'s directory is not a git worktree; left alone`. The sweep reports it once, like any other kept checkout. Nothing is committed, pushed, linked or deleted through it.
- the branch-links pass reads no branch from it, so no `main ->` link
- teardown records no branch from it, so the archive cannot say the run was on `main`
- `markFailedStart` no longer writes a marker where the checkout is gone — the daemon's log line is the record instead. That is how the residue directory came to exist (second half of the issue).

The directory itself is left where it is: the framework cannot know what a non-worktree directory holds, so it reports rather than deletes. Removal is a human `rm`.

## Verification

- `pnpm test`: 1560 node + 799 dashboard, green.
- Four new tests against real git, each verified by breaking (each fails on exactly its own guard with main's file swapped in):
  - `isWorktreeRoot` true for the main checkout and a linked worktree, false for a subdirectory, a residue directory, and a directory outside any repo — and the test states the bug directly: a plain `currentBranch` in the residue answers with the enclosing repo's branch, `worktreeBranch` answers nothing
  - removal of a residue directory is refused; the user's dirty main checkout is still dirty, HEAD unchanged, nothing pushed, the directory still there
  - the rename pass with default seams on a real repo: the real rename is linked, no `main` link for the residue (without the guard the test shows `'main'` appearing)
  - the failed-start marker creates no directory where the checkout is gone
- Not dogfooded: the rig residue was removed by hand before this was built, and reproducing it means removing a worktree under a live run. The real-git tests are the reproduction.

No per-file SPEC edits (post-wipe, #1655 regenerates them); one `FEATURES-SPEC.md` line.

### DIFF
--- a/FEATURES-SPEC.md
+++ b/FEATURES-SPEC.md
@@ -130,6 +130,8 @@ happens while nobody is at the keyboard.
 - CI watch: one fix agent per red head commit, max two attempts
 - Reclaim the checkout of an agent whose work is on the remote — never by publishing what a `handoff: local` agent refused to
 - An agent that committed nothing leaves no branch behind: its empty branch goes with its checkout, never pushed, so a pinned routine is not stood down by its own last run
+- A directory under `branches/` that git does not know as a worktree is never committed, pushed, linked or deleted through — it is reported and left alone, so a leftover can never stand in for your own checkout
+- A directory under `branches/` that git does not know as a worktree is never committed, pushed, linked or deleted through — it is reported and left alone, so a leftover can never stand in for your own checkout
 - Release a pinned routine branch left behind by a closed PR — before the schedule fires the routine, and before its "Run now" does
 - The agent drains its own TODO backlog, one entry per turn
 
--- a/FEATURES-SPEC.md
+++ b/FEATURES-SPEC.md
@@ -131,7 +131,6 @@ happens while nobody is at the keyboard.
 - Reclaim the checkout of an agent whose work is on the remote — never by publishing what a `handoff: local` agent refused to
 - An agent that committed nothing leaves no branch behind: its empty branch goes with its checkout, never pushed, so a pinned routine is not stood down by its own last run
 - A directory under `branches/` that git does not know as a worktree is never committed, pushed, linked or deleted through — it is reported and left alone, so a leftover can never stand in for your own checkout
-- A directory under `branches/` that git does not know as a worktree is never committed, pushed, linked or deleted through — it is reported and left alone, so a leftover can never stand in for your own checkout
 - Release a pinned routine branch left behind by a closed PR — before the schedule fires the routine, and before its "Run now" does
 - The agent drains its own TODO backlog, one entry per turn
 
--- a/packages/the-framework/src/branch-links.test.ts
+++ b/packages/the-framework/src/branch-links.test.ts
@@ -1,8 +1,11 @@
 import { strict as assert } from 'node:assert'
 import { test } from 'node:test'
 import { join } from 'node:path'
+import { tmpdir } from 'node:os'
+import { mkdir, mkdtemp, readdir, realpath, rm, writeFile } from 'node:fs/promises'
+import { nodeGitRunner } from './project.js'
 import { reconcileBranchLinks, startBranchLinksPass, type LinksFs } from './branch-links.js'
-import { FRAMEWORK_DIR, BRANCHES_DIR, type WorktreeDirEntry } from './store/index.js'
+import { FRAMEWORK_DIR, BRANCHES_DIR, addWorktree, worktreePath, type WorktreeDirEntry } from './store/index.js'
 
 const CWD = '/repo'
 const LINKS = join(CWD, FRAMEWORK_DIR, BRANCHES_DIR)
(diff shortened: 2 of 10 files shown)


suleimansh:
🤖 *automated*

`FEATURES-SPEC.md` carries the new line twice:

```
- A directory under `branches/` that git does not know as a worktree is never committed, pushed, linked or deleted through — it is reported and left alone, so a leftover can never stand in for your own checkout
- A directory under `branches/` that git does not know as a worktree is never committed, pushed, linked or deleted through — it is reported and left alone, so a leftover can never stand in for your own checkout
```

An interrupted edit had already written it once before the second edit added it again. Removing the duplicate in a follow-up commit.

suleimansh:
🤖 *automated*

**Dogfooded on the rig from this branch's build, 2026-08-23 15:53 IDT — no agent needed.** The residue from this morning was recreated exactly: `.the-framework/branches/tf-agent-2026-08-23T09-00-00-000Z/` holding only `.the-framework/agent.json` (status `failed`) + `events.jsonl`, not a worktree — `git rev-parse --abbrev-ref HEAD` inside it answers `main`. And the rig's main checkout was given an uncommitted edit to `README.md`, the thing the old path would have swept into a commit.

Daemon restarted; its startup sweep:

```
kept the worktree for session 2026-08-23T09-00-00-000Z: session 2026-08-23T09-00-00-000Z's directory is not a git worktree; left alone
```

After: `README.md` still uncommitted, HEAD still `3833ed2`, `origin/main` untouched, no `main ->` link in `branches/`, the directory still there. This morning's build on the same setup read `main`, committed nothing only because the tree was clean, and tried to push the user's `main` every pass.

The dashboard's Remove button runs the same `removeProjectWorktree`, so it is covered by the same verdict. Residue removed and the edit reverted afterwards; daemon back on `main`.

PR #1658, opened 2026-08-23
The run-id branch an agent branched away from goes with its checkout (#1657)

### LINKED ISSUE #1657: Every run leaves its empty run-id branch behind: the agent creates `tf-<name>` instead of being renamed onto it
🤖 *automated*

Residue noted in #1653 and #1650's post-merge check; the same class one step earlier.

## What happens

A run's worktree is created on `tf-agent-<id>`. The system prompt then tells the agent to *create* `tf-<SESSION_NAME>` and check it out (`prompts/system_prompt.md:39`), so the agent branches off the run-id branch, and `renameAgentBranch` finds the checkout already moved and renames nothing — by design (`src/store/worktree.ts:262`: the prompt ships verbatim from #326, so the rename only covers an agent that did not branch itself).

The run-id branch stays behind at the commit the run started from, forever. On the rig after two days:

```
tf-agent-2026-08-22T22-03-49-360Z   ← 9 local, all at main
tf-agent-2026-08-22T22-03-49-409Z   ← 2 of them also on origin (pushed by the pre-#1653 teardown)
…
tf-agent-2026-08-23T12-12-07-570Z
```

#1653 deletes the branch the *checkout is on* when it holds nothing; the run-id branch is a sibling ref the checkout is no longer on, so it was never judged.

## What to do

Same place as #1653, one more step: when the checkout's branch is not the run-id branch, and the run-id branch is an ancestor of it (`git merge-base --is-ancestor tf-agent-<id> <branch>` — everything it holds is in the branch that stays), delete it with the checkout. Not an ancestor — the agent committed on the run-id branch and then branched from somewhere else — keeps it, as it holds something the kept branch lacks. Nothing is pushed or touched on origin; the two empty copies already there are leftovers of the old push and can go by hand.

The alternative — have the daemon rename instead of the agent branching — means changing the #326 prompt, which is the bigger call.

suleimansh:
🤖 *automated*

Closes #1657.

## The bug, plainly

A run's worktree starts on `tf-agent-<id>`. The system prompt (`prompts/system_prompt.md:39`, verbatim from #326) then tells the agent to *create* `tf-<SESSION_NAME>` and check it out — so the agent branches off the run-id branch, the rename in `renameAgentBranch` correctly does nothing, and the run-id branch stays behind at the commit the run started from. One per run, forever: the rig had nine after two days, two of them also on origin (pushed by the pre-#1653 teardown). #1653 judged only the branch the checkout was *on*; this is its sibling.

## What changes

In `removeProjectWorktree`, one step more: when the checkout's branch is not the run-id branch and **contains** it (`git merge-base --is-ancestor`), the run-id branch holds nothing the kept branch lacks and goes with the checkout. A run-id branch carrying a commit the agent made before branching from somewhere else is not contained and stays. Decided before either delete — the containment reads both refs (the first cut checked after deleting the checkout's branch, and the triage case then kept the run-id branch; caught by its test).

The result is now `branchesDeleted: string[]` (was `branchDeleted: string`) and the sweep and teardown lines name every branch that went: `removed the worktree for session … and its branches tf-triage-quick and tf-agent-…: nothing on it is missing elsewhere`.

Nothing is pushed; origin is not touched. The two empty copies already on the rig's origin are leftovers of the old push, to go by hand.

## Verification

- `pnpm test`: 1563 node + 799 dashboard, green.
- New real-git tests: the run-id branch goes while the agent's work branch stays (pushed); the full rig case — commitless triage on `tf-triage-quick` cut from an empty run-id branch — leaves no `tf-*` branch and pushes nothing; a run-id branch with a commit of its own stays. The "user's `release` branch" test from #1653 now also asserts the run-id branch it was cut from goes. Without the fix, exactly those fail.
- Daemon teardown test ×3 green.
- Dogfood to follow: one triage click on the rig, where every run so far left a run-id branch.

One `FEATURES-SPEC.md` line extended; no per-file SPECs (post-wipe, #1655).

### DIFF
--- a/FEATURES-SPEC.md
+++ b/FEATURES-SPEC.md
@@ -129,7 +129,7 @@ happens while nobody is at the keyboard.
 - CI watch: merge a PR once its checks pass
 - CI watch: one fix agent per red head commit, max two attempts
 - Reclaim the checkout of an agent whose work is on the remote — never by publishing what a `handoff: local` agent refused to
-- An agent that committed nothing leaves no branch behind: its empty branch goes with its checkout, never pushed, so a pinned routine is not stood down by its own last run
+- An agent that committed nothing leaves no branch behind: its empty branch goes with its checkout, never pushed, so a pinned routine is not stood down by its own last run — and the run-id branch it started on goes too, once the branch it moved to holds everything the run-id branch did
 - A directory under `branches/` that git does not know as a worktree is never committed, pushed, linked or deleted through — it is reported and left alone, so a leftover can never stand in for your own checkout
 - Release a pinned routine branch left behind by a closed PR — before the schedule fires the routine, and before its "Run now" does
 - The agent drains its own TODO backlog, one entry per turn
--- a/packages/the-framework/src/daemon-runtime.ts
+++ b/packages/the-framework/src/daemon-runtime.ts
@@ -40,6 +40,7 @@ import { tailEvents, tailAgentEvents } from './dashboard-rpc/events-tail.js'
 import { resolveUserDir } from './agent-archive.js'
 import { withDataBranch } from './data-branch.js'
 import { removeProjectWorktree } from './worktrees.js'
+import { describeDeleted } from './merged-worktrees.js'
 import { reconcileBranchLinks } from './branch-links.js'
 import { scopedKey, parseScopedKey, keyBelongsTo } from './runtime-keys.js'
 import { addProject, listProjects, projectId } from './registry.js'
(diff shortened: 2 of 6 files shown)


suleimansh:
🤖 *automated*

**Dogfooded on the rig from this branch's build, 2026-08-23 16:09 IDT** — one quick-win triage click (Opus, ~2 min). Every previous run on the rig left its run-id branch behind; this one's teardown, verbatim:

```
removed worktree …/tf-agent-2026-08-23T13-09-49-775Z and its branches tf-triage-quick and tf-agent-2026-08-23T13-09-49-775Z: nothing on it is missing elsewhere
```

`tf-*` branch count on the rig: 13 before the run, 13 after — the first run to leave nothing. Nothing pushed. The nine historical empty run-id branches were deleted by hand afterwards (they have no checkout to be judged through); the two empty copies on origin are left for a human `push --delete`. Daemon back on `main`.

Issue #1332, opened 2026-07-28
Spike: Is it possible to make the extension headless?


brillout:
For full-fledged CC web support, see also:
- https://github.com/gemstack-land/the-framework/issues/1554

brillout:
If not possible, then I suggest we use the extension merely to cyphon the cookie.

suleimansh:
## Spike result: headless no, unattended yes

Tested live on 2026-08-24 with throwaway profiles, no login, no session spent. The question was whether the bridge extension can run without a user's browser; the answer splits in two.

| Test | Result |
|---|---|
| Branded Google Chrome 151, `--headless=new --load-extension=packages/chrome-extension` | The extension never loads. Branded Chrome ignores `--load-extension` since 137, and the `--disable-features=DisableLoadExtensionCommandLineSwitch` opt-out no longer works either. |
| Chrome for Testing 150, headless, same flags | Loads. The worker is a CDP `service_worker` target; over CDP `Runtime.evaluate` in it, `chrome.storage.local.set({daemonUrl, token})`, `chrome.tabs.create` (pinned tab) and `chrome.tabs.sendMessage` to the content script all work — the content script injects and answers `tf-create-session`. Every piece the bridge needs runs headless. |
| claude.ai inside that headless browser | Cloudflare's "Performing security verification" interstitial, and it never clears (50 s+). |
| Same binary, same fresh profile, same extension, **headed** | Passes Cloudflare at once and lands on the sign-in page. |

So the extension itself is fine headless; the blocker is claude.ai's Cloudflare gate rejecting a headless browser. Working around bot detection is not something we do, which leaves three options.

### Option A — a daemon-owned headed browser (recommended)

The daemon launches Chrome for Testing (not the user's Chrome, which cannot load an unpacked extension any more) with a persistent profile, `--load-extension=packages/chrome-extension`, `--remote-debugging-port`, and seeds the token over CDP as the spike did. The user signs in once in that window; from then on it is a background window (minimized/off-screen on macOS, Xvfb on Linux) and `extensionAlive` (`bridge-store.ts`, 3-minute window) holds for as long as the daemon runs — fan-out no longer depends on the user's Chrome being open.

- Seam: `browser.ts` `launchSharedBrowser`, which needs a sibling — persistent profile instead of throwaway, headed, extension flags, and a Chrome for Testing binary (`@puppeteer/browsers install chrome@stable`, ~150 MB, or one already on the machine).
[...]

Issue #1237, opened 2026-07-26
Bridge a Claude web session into the dashboard with a browser extension


suleimansh:
Show a cloud session's question in the dashboard, and let the user answer it there, using a browser extension running in the user's own logged in Claude session.

This is the interaction story #1235 is missing. #1234 stops a cloud run deadlocking on a question nobody can answer; this is the other half, letting it ask.

## Why an extension and not the API

There is a session API, and it does exactly what we want. Found by reading the shipped CLI binary (2.1.220):

| | endpoint |
|---|---|
| stream | `GET /v1/code/sessions/{id}/events/stream`, SSE, resumable via `from_sequence_num` |
| history | `GET /v1/code/sessions/{id}/events?limit=&sort_order=desc` |
| write | `POST /v1/code/sessions/{id}/events` with `{ session_id, events: [{ payload }] }` |

Two things rule it out for now. It is private and undocumented, so it can break on any CLI update. And it authenticates with the user's subscription credential, which would mean the framework holding that credential rather than the CLI holding its own, which is the posture we chose deliberately in #495.

The CLI will not lend us the API either, which was the hope:

- `--attach`, `--follow` and `--watch` are not Claude Code flags. They are strings from bundled dependencies. A control test with a nonsense flag is what exposed this; a `--help` based check wrongly suggests they are accepted, because `--help` short circuits option validation.
- `--teleport` is interactive only, and it does not refuse when it cannot work. Under `-p` it is **silently ignored** and a fresh local session runs instead. Confirmed twice, including under a pty, which is the trick that made `--cloud` work; the pty does not help because `-p` itself is the disqualifier. It cost 0.099 USD and 26 turns to find this out, which is worth knowing on its own: anything that shells out `--teleport` with `-p` will quietly do local work and bill for it.

An extension avoids the credential problem entirely. It runs in the user's own browser, in a session they are already signed into, with them present.

## What the spike found

Branch `spike/cc-web-extension`, directory `spike/cc-web-extension/`. It lives outside `packages/` and `examples/`, which the pnpm workspace globs do not match, so it joins no build.

It requests **no host permissions** and sends nothing anywhere. It answers only the readable half: can it find the question, and the box an answer would go into. Load it unpacked and open a parked session; the README says what a passing result looks like.

**Transport is already known to be a problem, and it is not the obvious one.** The daemon's CSRF guard rejects an extension outright. Measured against a running daemon:

```
Origin: chrome-extension://abc   -> 403
[...]

PR #1664, opened 2026-08-23
A cloud session asks its questions when the bridge is on, and they are answered like a local gate (#1554)

### LINKED ISSUE #1554: [CC web driver] Add support for choices? For full-fledged CC web driver
Can we pipe the choice macros to TF?

Labeling as high prio so that we know whether we should invest in the CC web driver.

For full-fledged CC web support, see also :
- https://github.com/gemstack-land/the-framework/issues/1332

suleimansh:
🤖 *automated*

Closes #1554. Addresses #1225 (the choice half; the mirror's per-turn split stays open there).

## What was wrong

The browser bridge has been able to carry a cloud session's question home and type an answer back since #1237 — but no Framework-started cloud session ever asked one. The hands-off protocol told every web run to *decide alone*. That rule was right when nothing could answer (#1234); with the bridge on, it is the whole reason choices "didn't work" on CC web.

## What changes

**The session may ask, when something can answer.** `hands_off.md` splits in two:
- `hands_off.md` keeps the land-everything rule (commit + open a PR) for every web run.
- `unattended.md` carries decide-alone and is injected only when the browser bridge is **off**. `cli.ts` reads `preferences.bridge` at hand-off and passes `RunAgentOptions.bridge`; `composeAgentSystem` gets `unattended` next to `handsOff`.

**The question keeps its shape.** The bridge question carries the block's own `multi`, and per option `default` and `stop`. Labels must be distinct — they double as the pick's ids.

**The answer is what a local gate hears.** The daemon composes the text from validated labels: the same `You paused to ask: "…". The user chose: …. Continue with that decision.` a local agent is re-prompted with; a `stop` pick types a take-over line instead (`takeoverPrompt`), since nothing of ours can end a session on claude.ai. `queueAnswer` takes `labels[]`; `/_bridge/answer` serves `{id, text}`; the extension types `text`.

**One gate panel.** The parked question renders through `ChoicePanel` (new `send` prop) — one click, multi-select, recommended — on the cloud run's view *and* in the Open Questions hub, which joins bridged questions to their web run through the agent archive (`OpenQuestion.bridge`). One card per question across checkouts sharing a `tf-data` archive.

Extension **0.9.0** (the daemon expects it); `check.mjs` pins the posted shape.

## Dogfood (real Opus web run on this repo, `2026-08-23T15-29-19-847Z`)

1. The run's system prompt carried the await protocol and land-everything, no decide-alone.
2. The cloud session `session_01HgamNnpzgVyp8TUoWt7fRM` **parked on its question** — `await-choices` with a `stop: true` option — the first Framework-started cloud session to do so.
3. Within a minute the question was on the run's view and in the hub as a "Your call" gate.
4. One click on the non-recommended option → answer `sent` (`filled contenteditable, clicked send button`) → the continuation prompt appeared verbatim in the session, which **resumed**, wrote `docs/NOTES.md` on the chosen path, committed and pushed `claude/dogfood-choice-bridge-yil7x4`; the daemon adopted that branch (#1601).

Two observations from the dogfood, not in scope here: the session pushed but did not open its PR despite the land-everything rule (with handoff `pr` the framework would have armed one — this run was started with `push`); and `suleimansh/tf-1334-dogfood` is not in the Claude GitHub App's repository selection, so web runs cannot be dogfooded on the rig.

## Not in this PR

Keeping the web run `running` until its PR lands (the #1231 reversal), the mirror's per-turn split (#1225 item 1), headless (#1332).


suleimansh:
@brillout merging this. One prompt change, so pasting it for you:

`prompts/protocols/hands_off.md` used to tell every web run two things in one block: *decide alone, never ask* and *commit + open a PR before you end*. The first rule is why no cloud session ever asked a question — even though the bridge could carry one since #1237.

It's now two files:

**`hands_off.md`** — every web run, unchanged rule: commit your work and open a PR before ending, since nothing here sees your workspace.

**`unattended.md`** — only when the browser bridge is *off*: nothing can answer you, so don't emit the await block; take the option you'd have marked `recommended`, say which assumption you made, and finish.

With the bridge on, a web run gets neither "decide alone" nor anything new — just the same await protocol a local agent has. It parks, the question shows in the dashboard, the answer is typed back into claude.ai.

Reads fine to me — ping me if you'd word either differently.

brillout:
> unattended.md

Can we remove this whole thing? Such "bad" features should be eagerly removed. As always: let's reduce LoC as much as possible.

Issue #1568, opened 2026-08-17
Bridge false-positive: the system-prompt template renders as a real question when the session asked nothing


suleimansh:
Found live 2026-08-17 while verifying the reloaded v0.8.0 extension (fresh install, correct token, bridge healthy: `bridge: sent`, `transcript: 1 block(s)`).

**Repro:** open any TF-started cloud session page on claude.ai — one whose transcript renders the quoted Framework instructions in the user message — with the bridge configured. The session in question (`session_017PSxQkP3qP2ugF2YJFrYFE`, the tickets-sync run `2026-08-17T13-53-39-020Z`) never asked a question and had already finished.

**Observed:** the panel reports `question found: yes (code)` with title `<what the human needs to do> (<the page you are stuck on>)` and options `Handled it | Could not handle it` — the literal placeholder example from the "Handing the browser to a human" protocol section, scraped out of the rendered system prompt. The bridge then *sends* it, and the dashboard run view renders a real, answerable question card for a question the session never asked ("Pick an answer" would type into the real composer of a finished session).

**Why the existing defense misses it:** `content.SPEC.md` says the page's rendered copy of our own protocol template is a decoy and "the last real question wins" — which works when a real question exists *after* the decoy, but when the page contains only the decoy (every TF cloud session renders the instructions; almost none ask), the decoy is the last question found and wins by default.

**Fix directions:**
1. Recognize the template's placeholder shape (`<what the human needs to do>`, options `Handled it`/`Could not handle it`, or generally an angle-bracket-placeholder title) and never report it — the template is a constant we ship, so matching it exactly is cheap and precise.
2. And/or scope extraction to agent-message containers so text quoted inside the *user* message (which is where the instructions render) is never a candidate.

Until fixed: the bogus card is showable for any watched TF session; answering it types into the live session, so the safe behavior is to ignore it.

PR #1671, opened 2026-08-23
The cloud session mirror reads the page's turns, not a page dump (#1225)

suleimansh:
🤖 *automated*

Addresses #1225 — item 1 ("it dumps the whole page, not the conversation"); item 2 (the 8000-char cap cutting from the front) goes with it, since the cap is now per turn.

## What was wrong

The content script looked for `<article>` elements to split the conversation. claude.ai renders none (`articles 0` in every hello note), so the mirror always fell to its plan B: the whole conversation region as one 8000-character tail. What the dashboard showed was the run's own prompt, the sidebar labels, and the actual exchange somewhere at the end, with no turn boundaries.

## What the page actually marks

Inspected on a real session (`session_01HgamNnpzgVyp8TUoWt7fRM`): the transcript is a `div[role="feed"]` ("Chat messages") of `div[data-testid="transcript-row"]`, one per turn, each carrying `data-index` (position), `data-perf-row` (`human` / `assistant` / `marker`) and `data-perf-row-streaming`. Row 0 is the run prompt (11.5 k chars); the exchange is rows 2–5. The list is virtual — a long session keeps only its tail rendered.

## What changes

**Extension `0.9.0 → 0.10.0`** (`packages/chrome-extension`, daemon lockstep bumped)
- `transcript()` reads the rows: `seq = data-index`, `human → user`, `assistant → agent`, markers skipped; per-turn text cleaned (icon glyphs, blank lines) and head-capped at 8000 — only the prompt turn is ever that long. Streaming replies are re-sent in place (the store already keys by position).
- **The page-dump fallback is removed.** A page with no turn rows mirrors nothing; the panel's transcript line and the hello note say `no transcript rows found`. A layout change is a named state, not 8 k of sidebar that reads like output.
- The hello note reports the row kinds seen (`rows 8 (human,marker,assistant), turns 4`), so a new kind shows up by name.
- Same selector fixes a latent gap: `findPendingChoice` scoped "the opening message" by `deepQueryAll('article')[0]` — `undefined` on every live page, so the #1568 decoy exclusion by position never ran there and `isTemplate` carried it alone. The opening message is now the `data-index="0"` row.
- `check.mjs`: fixtures move to the real row shape; three new cases — one entry per turn with roles mapped / markers skipped / prompt head-capped; positions from the page when only the tail is rendered; no rows → nothing mirrored and the panel says so. Break-checked (positions from DOM order fails).

**Dashboard** (`CloudAgentNotice.tsx`)
- Turns render by role: the user's as one muted `you ›` line (first line, whole turn on hover — the opening turn is the entire prompt), the session's in full; the box scrolls to the newest turn whenever it changes or grows. `scrubMirrorText` stays (per-message action labels still ride in row text). One new test, break-checked.

No change to `/_bridge/events` or the store — the `seq`/`role`/`text` contract already fit.

## SPEC.md changes

**`packages/chrome-extension/content.SPEC.md`** — the page half of the bridge.

TL;DR, two lines replaced:
```diff
-- **The transcript is mirrored per message** - each message block is reported under its position, and only positions whose text changed are re-sent.
-- **When the page has no message blocks, the conversation is mirrored whole** - as one block holding its most recent text, rather than guessing where messages divide.
+- **The transcript is mirrored one turn at a time** - each conversation turn the page marks is reported under the position the page gives it, as the user's or the session's, and only turns whose text changed are re-sent.
+- **A page that marks no turns mirrors nothing** - the panel and the self-report name that state, rather than mirroring whatever text is on screen.
```
"Mirroring the transcript" rewritten (rows, kinds, positions from the page, streaming replace, nothing-when-no-rows, what the self-report carries; the rationale records why the dump was dropped). The "Mirroring a page with no message blocks" section is removed. In "Telling a real question from the protocol's own examples", the position rule now names the opening turn as position zero and says what happens when a virtual list has scrolled it away. The panel's "how many message blocks" → "how many turn rows".

**`packages/chrome-extension/SPEC.md`** — one phrase: the transcript is mirrored "turn by turn, as it is written".

**`packages/chrome-extension/check.SPEC.md`** — the #1568 bullet speaks of the opening turn; a new "Mirroring the transcript" group lists the three cases above.

**`packages/framework/dashboard/components/CloudAgentNotice.SPEC.md`** — TL;DR and glossary say "turn by turn"; the mirror section gains:
```diff
+Each turn is its own block, in transcript order. The user's turns show as a single line — "you ›" followed by the turn's first line, the whole turn available on hover — because the opening turn is the run's entire prompt and what the session did is the point of the mirror, not what it was told. The session's turns are shown whole. The box keeps its newest turn in view, scrolling to the end whenever the last turn changes or grows, so a reply being written is watched rather than scrolled to.
```

**`FEATURES-SPEC.md`** — the mirror had no entry; added under Remote execution and sharing:
```diff
+- A cloud session's conversation mirrored into the run view, turn by turn, as it is written
```

## Verified

- `node check.mjs`: 20/20. `pnpm build` / `typecheck` / `test`: 1581 daemon + 804 dashboard, green.
- Dogfood on a live gemstack web run: pending the extension reload (human step) — will be reported here.


suleimansh:
🤖 *automated*

Dogfooded end to end on a live Opus gemstack web run (`2026-08-23T19-49-17-177Z` / `session_01J221xfDeFa5MNm2Rvaaf4E`), extension 0.10.0 reloaded:

- Hello note on the live page: `rows 4 (human,marker,assistant), turns 2` — the real row kinds, named.
- Mirrored turns at the daemon: `(0 user 8000)` — the prompt, head-capped — then `(2 agent)`, `(3 user)` — the answer the bridge typed — `(4 agent)`. Positions are the page's own (`seq 1`, a marker, is absent by design), markers skipped, roles correct.
- The parked question crossed whole (recommended + stop options), was answered from the dashboard, typed back, and the session resumed — the #1554 round trip is unaffected.
- Run view: 4 alternating blocks; both `you ›` turns render as one truncated line with the full turn in the tooltip, both session turns in full, `Show message actions` scrubbed, box scrolled to the newest turn.
- The session created NOTES.md with the chosen marker, committed, pushed `claude/dogfood-per-turn-mirror-r2jf50` — branch verified and deleted, along with the run's empty `tf-agent-*` branch.

Issue #182, opened 2026-07-03
Loop can't recover an empty/unbuilt workspace: 'improve' framing blocks scaffolding


suleimansh:
## Symptom
In the full-scope live run (see #181), the build phase produced **zero files** (derailed by universal-orm), and the loop then failed to recover: passes 1–2 built nothing, pass 3 finally scaffolded only a bare Vike 'Welcome' starter and ran out of passes. The user ran the result — it serves the default 'Welcome' page, none of the requested blog.

## Root cause (distinct from #181)
The full-fledged loop assumes an app already **exists** and only needs polishing:
- `improvePrompt` (packages/framework/src/steps.ts): *"Address these blockers… Make the **smallest changes** that clear them. **Do not add unrelated features**."* When the blocker is 'the workspace is empty / no app', 'smallest change' + 'no unrelated features' actively discourages scaffolding a whole app. In the run the agent read the empty dir as a possible wrong-path and refused: *"I can't clear this blocker by fabricating… the directory is genuinely empty."*
- The **build step doesn't guarantee output** — if the first build turn produces nothing (e.g. it stalls on a dependency), nothing forces a hard 'the app does not exist yet — create it from scratch'.
- `--max-passes` default 3 leaves no room once 2 passes are spent not-building.

## Fix ideas
- Detect an empty / unbuilt workspace (no source files, or build produced nothing) and switch the improve prompt into a **'scaffold the whole app from scratch'** directive instead of 'smallest change to clear a blocker'.
- Make the build step **verify it produced files** and hard-fail / re-prompt with a stronger 'you must create the app' if it didn't.
- Raise the default `--max-passes` (or make the first successful scaffold not count against the polish budget).

## Acceptance
- A from-scratch full-scope run where the build stalls once still recovers and scaffolds the real app (not just a starter), or fails loudly with a clear reason — it never silently ends at the framework's default 'Welcome' page.

Related: #181 (universal-orm derail, the trigger), #165 (presentation epic), #110 (framework epic).

PR #1683, opened 2026-08-24
The build-opening prompts live in prompts/ as markdown (#1347)

### LINKED ISSUE #1347: Four agent-facing prompts are still written in TypeScript
`prompts/README.md` says:

> Every prompt the framework sends an agent lives here as markdown (#551). Nothing agent-facing is written in TypeScript any more, so prompting can change without touching the code.

That isn't true. `steps.ts` still builds four of them as TypeScript strings:

- `buildPrompt` (greenfield)
- `extendPrompt` (existing codebase)
- `scaffoldPrompt` (empty workspace)
- `improvePrompt` (fix blockers)

So changing what we tell an agent in those four cases means editing code, and they're invisible to anyone reading `prompts/`. #1224 was a one-word prompt edit that had to go through a TypeScript file and a changeset.

Move them to `prompts/` like the rest, so all prompting lives in one place and can be reviewed as prose.

suleimansh:
Closes #1347.

`prompts/README.md` promised that nothing agent-facing is written in TypeScript, but `steps.ts` still built three prompts as string arrays (the issue counted four — `improvePrompt` was already removed with the production-grade gate). Changing what we tell an agent in those cases meant editing code (#1224 was a one-word prompt edit through a TypeScript file).

## What changed

- New `prompts/build_prompt.md`, `prompts/extend_prompt.md`, `prompts/scaffold_prompt.md` — the prompt text moved verbatim, with `${{ tf.prompt }}` for the user's intent (the same name the system prompt gives it). Each with its SPEC.md sibling carrying the rationale that lived in the code comments (#185's naming-the-workspace point, #182's don't-refuse retry, #545's stack-is-the-agent's-call).
- `steps.ts` keeps the three functions and `isWorkspaceEmpty` — the functions are now one-line `renderTemplate` fills over the generated constants, so call sites are untouched.
- **Rendered output is byte-identical to the old strings** — verified by direct comparison, and the existing `steps.test.ts` content assertions all still pass unchanged. Break-checked: editing `extend_prompt.md`'s first line fails the test.
- `prompts/README.md` table row added; `prompts/SPEC.md` and `steps.SPEC.md` updated — the prompt-content spec now lives with the prompts, steps.SPEC keeps only the choice + fill and the emptiness check.

## Tests

Full suite green: 1586 node tests + 817 dashboard tests; `pnpm typecheck` clean. No behavior change intended, and the byte-identity check says none happened.

🤖 *automated*

### DIFF
--- a/packages/framework/prompts/build_prompt.SPEC.md
+++ b/packages/framework/prompts/build_prompt.SPEC.md
@@ -1,0 +1,9 @@
+The greenfield build prompt: the opening prompt of a build agent whose workspace holds no app yet.
+
+## Business logic
+
+It frames a from-scratch, end-to-end build of the user's intent: the workspace may be empty, and if so the agent is to scaffold the whole project — package manifest with scripts, all config, every source file — install the dependencies, and make the app run, then summarize what it built in one short paragraph. The user's intent fills the first line; the stack is deliberately left to the agent.
+
+## Before modifying/creating SPEC.md files
+
+You must always read and respect https://raw.githubusercontent.com/brillout/sdd/refs/heads/main/sdd.md
--- a/packages/framework/prompts/build_prompt.SPEC.md
+++ b/packages/framework/prompts/build_prompt.SPEC.md
@@ -2,7 +2,11 @@ The greenfield build prompt: the opening prompt of a build agent whose workspace
 
 ## Business logic
 
-It frames a from-scratch, end-to-end build of the user's intent: the workspace may be empty, and if so the agent is to scaffold the whole project — package manifest with scripts, all config, every source file — install the dependencies, and make the app run, then summarize what it built in one short paragraph. The user's intent fills the first line; the stack is deliberately left to the agent.
+Sent as the opening prompt when the workspace holds no source at build time; the user's intent fills the `${{ tf.prompt }}` slot. The prompt file beside this spec is itself the prose of what the agent is told.
+
+## Rationale
+
+The stack is deliberately not prescribed — it is the agent's call.
 
 ## Before modifying/creating SPEC.md files
 
--- a/packages/framework/prompts/build_prompt.md
+++ b/packages/framework/prompts/build_prompt.md
@@ -1,0 +1,5 @@
+Build this app end to end: ${{ tf.prompt }}
+The workspace may be empty — if so, scaffold the whole project from scratch:
+create package.json with scripts, all config, and every source file, install
+the dependencies, and make the app run.
+When done, summarize what you built in one short paragraph.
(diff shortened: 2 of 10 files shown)


suleimansh:
DRY check — two duplications in this PR:

1. **The per-prompt SPEC.md files paraphrase the prompts they sit next to.** `build_prompt.SPEC.md`'s business-logic paragraph restates `build_prompt.md` nearly clause for clause (scaffold the whole project, package manifest with scripts, install dependencies, one-paragraph summary), and `scaffold_prompt.SPEC.md` does the same for its prompt. The prompt file is already prose — the reviewable artifact — so this is the same content in two places, and the copy in the SPEC will silently drift on the next prompt edit. The fix: each SPEC keeps only what the prompt itself does not say — when it is sent, and the rationale (the dropped how-to-behave rules, the stack being the agent's call, why the retry insists an empty directory is expected) — and stops re-narrating the prompt body.

2. **`steps.ts` builds the identical render context three times.** All three functions are `renderTemplate(X, { tf: { prompt: intent } })` — the context shape is repeated per function. A local `fill(template, intent)` helper would leave one place that knows how a build-opening prompt renders.

Not counting: the `prompts/SPEC.md` bullet appearing in both the TL;DR and the "What lives here" list — every other category in that file already appears in both, so that repetition is the file's existing pattern, not new.

🤖 *curated*

suleimansh:
Pushed the dedup: the three SPECs now carry only when-sent + rationale (the prompt file beside each is the prose of what the agent is told), and `steps.ts` renders through one `fill` helper. Steps tests + typecheck green.

🤖 *curated*

suleimansh:
@brillout prompt files moved (#1347) — review welcome, merging

packages/framework/prompts/build_prompt.md
```diff
@@ -0,0 +1,5 @@
+Build this app end to end: ${{ tf.prompt }}
+The workspace may be empty — if so, scaffold the whole project from scratch:
+create package.json with scripts, all config, and every source file, install
+the dependencies, and make the app run.
+When done, summarize what you built in one short paragraph.
```
brillout, line 5: I don't understand the goal of this prompt. Can we remove it and tree-shake (remove all code related to it)?

packages/framework/prompts/build_prompt.SPEC.md
```diff
@@ -0,0 +1,13 @@
+The greenfield build prompt: the opening prompt of a build agent whose workspace holds no app yet.
```
brillout, line 1: Let's remove all of the greenfield stuff for now (and tree-shake)

brillout:
@suleimansh Up for reviewing the prompts? I'll do a second review after yours.

suleimansh:
It was the ai-autopilot greenfield flow (empty-folder builds) — dead path today. Agreed — removal PR incoming. Prompts review next.

PR #1689, opened 2026-08-24
Remove the greenfield flow: a build always extends the existing codebase

suleimansh:
Rom's review on #1683 (inline on `build_prompt.md`): the greenfield prompt's goal is unclear — remove it and tree-shake. This does that.

## What's removed

- `prompts/build_prompt.md` + `prompts/scaffold_prompt.md` (and their SPECs) — the ai-autopilot-era empty-folder flow: "build this app end to end" and the #182 "create it from scratch NOW" retry. A TF project is a repo that already exists; no product path leads a build agent into an empty workspace.
- `isWorkspaceEmpty` and its helpers — the check existed only to pick greenfield vs extend.
- `agent.ts`: the #182 scaffold-retry block, and the opening-prompt fork — a build now always opens with `extendPrompt`. The fake driver (whose workspace always read empty, so it always took the greenfield path) now gets the extend framing too, so the tests exercise the one real path.
- The three greenfield tests; the fake-driver trigger matchers moved from the build wording to the extend wording, which live-asserts the new opening.

`steps.ts` is now just `extendPrompt` over the compiled template. Specs updated: `steps.SPEC.md`, `steps.test.SPEC.md`, `prompts/SPEC.md`, `prompts/README.md`.

## Tests

Full suite green: 1581 node tests (5 greenfield tests removed) + 827 dashboard tests; `pnpm typecheck` clean. Repo-wide grep finds zero remaining references to the removed symbols or files.

🤖 *automated*


suleimansh:
@brillout greenfield removed as requested

PR #1703, opened 2026-08-25
One Driver tab serves every cloud session, routed by claude.ai's session list (#1332)

suleimansh:
Part of #1332 — the Driver tab from the plan there. One pinned claude.ai tab serves every recent cloud session, instead of one tab per session capped at three.

## What changes for the user

- The extension opens one pinned background tab, **"The Framework Driver"**, covered by a full-page overlay: a heading, one line saying what the tab is for (and that closing it pauses the bridge), and a collapsed **Show debug logs** with the cycle log.
- Twice a minute the worker runs a cycle: it asks the daemon which sessions are its (every web run of the last 12 h, no cap, each flagged with whether an answer is queued), has the Driver read claude.ai's own session list for them, reports those statuses back, visits only the sessions that need it, types queued answers, and creates the session the daemon asked for — all in the one tab.
- A cloud run's row says **waiting** when claude.ai's list shows its session **Awaiting input**, even when the question was asked in prose rather than as a choice block.
- Closing the Driver tab pauses the bridge; the options page's **Open the Driver tab now** resumes it. The options page's switch and copy now describe the Driver.

## Measured on the live page (2026-08-25), which the design follows

- The sidebar rows are links to `/code/session_…` with a status span whose `aria-label` is text: `Awaiting input`, `Unread response`, `Idle`, `Running`, or `#<pr> · Merged/Closed`. The middle panel's "Needs input" rows are buttons without the session id, so the sidebar is what is read.
- Moving between sessions is in-app (a `window` marker survived row → session → "New" → list; one navigation entry).
- The list does **not** refresh on in-app navigation (tab A kept showing a session Unread after a click-away-and-back while a fresh load showed Idle), so the Driver reloads its page at most once a minute before reading — one load per minute rather than one per session.
- An in-app visit clears neither `Awaiting input` nor `Unread response`, so visits are planned on **change, age (5 min), or a queued answer** (`driver-plan.js`), not on every cycle.
- The list's "Show 20 more" button sits 1 level from its rows; the middle panel's "Show 2 more" sits 28 levels away — the reader only takes a button within 3 levels of rows.

## Protocol (lockstep bump to 0.12.0)

- `GET /_bridge/sessions` entries gain `answerQueued`.
- New `POST /_bridge/statuses` `{statuses: [{sessionId, status, label?}]}` — status ∈ awaiting/unread/idle/running/landed/missing/unknown; an unknown label travels verbatim (cut to 80), one bad entry refuses the batch, ≤ 500 entries.
[...]

PR #1704, opened 2026-08-25
Simplify: one shape for the daemon's project passes, the HTTP surfaces, and the picker menus

brillout:
Nine pieces of duplicated structure collapsed into shared ones. No feature changes: the same behaviour, the same tests, ~200 fewer lines of code (the rest of the diff is the branch's existing commits).

## The daemon's five background passes

The merged-worktree sweep, the CI watch, the branch-links pass, the cloud scratch-ref sweep and cloud work adoption each wrote out the same scaffolding: walk the registered projects, break between projects once stopped, join an overlapping tick rather than dropping it, hand back `{tick, stop}` — plus five identically-shaped interfaces under five different names, none of which anything outside its own module named.

`src/project-pass.ts` holds that shape once (`startProjectPass`, `ProjectPass`); each service is now just what it does to one project.

## The three non-RPC HTTP surfaces

The bridge, web-start and relay routes carried two near-identical capped JSON body readers (differing only in which message they rejected with, one of which the caller discarded) and three byte-identical plain-text responders.

`src/dashboard/http.ts` holds `readJsonBody`, `end` and `sendJson`, plus the POST preamble (`readPost`) and the GET-only guard (`requireGet`) that eight routes ran by hand. The bridge's two validators also share the session-id check they both rejected on with the same message.

`handleEvents` keeps its explicit method check on purpose — a GET on a daemon with events off is still a wrong method, not an un-wired one — with a comment saying so.

## The two data-branch claim modules

The ticket locks and the routine locks each resolved the same read/write/remove/list/funnel/log seams, mkdir-parents lambda included. `DataFileDeps` and `resolveDataFileDeps` now live in `data-branch.ts`, beside the funnel they write through; `RoutineLockDeps` extends it with the two seams that are its own.

## Dashboard

- The "Preferred editor" group was written out in both menus that offer it → `PreferredEditorItems`
- "Update from GitHub" was written out in all three surfaces that show it, tooltip prose and all → `UpdateTicketsButton` (the shared prompt already existed for exactly this reason; the button around it did not)
- The two ticket pages shared a frame but not a component → `TicketPageShell`
- The rail's Overview and Tickets rows were one row style copied → `NavRow`
- `OptionLabel`'s re-export through `OptionsMenu` had one straggling consumer left; it imports the real module now and the shim is gone

## Server

- `readTickets` and `readTicket` assembled a ticket row twice → `ticketRow`
- Both `gh pr create` callers parsed its output and invalidated the same two caches → `createdPr`
- Three test files rebuilt the dashboard wiring `test-context.ts` already had; it now builds the value (`testDashboardContext`) and the `startDashboard` options (`testDashboardOptions`) as well as setting the ambient context

## Dead surface

Dropped `export` from 24 names nothing outside their own module consumed, and removed one unused import.

## Specs

Three new source files get their `SPEC.md` siblings. No existing spec changed: business logic is untouched, and the specs describe behaviour rather than code shape.

## Testing

`pnpm typecheck`, `pnpm build` and `pnpm test` — the CI sequence — all pass on this tree: **1596 node tests, 854 dashboard tests, 0 failures.**

https://claude.ai/code/session_019Rk8i1CS5qbRRLszRXJNQV

---
_Generated by [Claude Code](https://claude.ai/code/session_019Rk8i1CS5qbRRLszRXJNQV)_

### DIFF
--- a/packages/framework/dashboard/components/AgentActionsMenu.tsx
+++ b/packages/framework/dashboard/components/AgentActionsMenu.tsx
@@ -1,7 +1,7 @@
 import { useEffect, useRef, useState } from 'react'
 import type { FrameworkEvent } from '../../src/index.js'
 import { sessionInfo } from '../../src/client.js'
-import { MoreVertical, Github, FolderOpen, Code, Check, Play, ExternalLink, Square, FolderX, Trash2, Copy, GitMerge } from 'lucide-react'
+import { MoreVertical, Github, FolderOpen, Code, Check, ExternalLink, Square, FolderX, Trash2, Copy, GitMerge } from 'lucide-react'
 import { onGithubUrl } from '../rpc/reads.js'
 import {
   sendOpenInApp,
@@ -10,17 +10,13 @@ import {
   sendRemoveWorktree,
   sendDeleteAgent,
 } from '../rpc/control.js'
-import type { EditorInfo } from '../rpc/preferences.js'
 import { useLoaded } from '../lib/use-async.js'
 import { useAction } from '../lib/use-action.js'
-import { usePreferences, updatePreferences } from '../lib/preferences.js'
-import { useDetectedEditors } from '../lib/editors.js'
 import { isAgentActive } from '../lib/live-state.js'
 import { describeSessionLink } from '../lib/session-link.js'
 import { buildResumeCommand } from '../lib/resume-command.js'
-import { cn } from '../lib/utils.js'
 import { buttonVariants } from './ui/button.js'
-import { OptionLabel } from './ui/option-label.js'
+import { PreferredEditorItems } from './PreferredEditorItems.js'
 import { ConfirmDialog } from './ui/confirm-dialog.js'
 import { Tooltip, TooltipTrigger, TooltipContent } from './ui/tooltip.js'
 import {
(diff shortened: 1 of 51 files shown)


suleimansh:
Post-merge review of #1704 (read the whole diff against the removed code).

**Main does not build since this merged.** #1703 (merged six minutes earlier) added `handleStatuses` on the file-local `readJsonBody`; this PR moved that reader to `src/dashboard/http.ts`. No textual conflict, so both landed, and CI on 350458bb fails:
`src/dashboard/bridge-endpoints.ts(366,18): Cannot find name 'readJsonBody'`. Fix in #1708: the route runs the same `readPost` preamble as the bridge's other routes.

**The simplification is real.** −150 lines of actual code once comments, blanks, SPECs and tests are stripped (the body's ~200 counts doc comments). Each of the nine collapses reproduces the removed behaviour: per-project catches, the join/stop rule, the head-vs-whole-file slice in `ticketRow`, the two tooltip variants, the 405-before-409 ordering in web-start. 24 exports dropped and the `OptionLabel` shim deleted — no compat paths.

Loose ends, none blocking:
- `relay-endpoints.ts:69-75, 112-118` still hand-roll the POST preamble (`invalid request body` for both failures), so `http.SPEC.md`'s "the caller learns which of the two it hit" is not true of the relay it names.
- The five pass SPECs (`ci-watch:105`, `merged-worktrees:64`, `branch-links:9`, `cloud-scratch-refs:56`, `cloud-work:45`) each still restate the join/stop paragraph now owned by `project-pass.SPEC.md`, in five wordings. One statement, five references would match sdd's one-concept-one-name rule.
- `dashboard-rpc/test-context.ts` gained `testDashboardContext`/`testDashboardOptions` and is the one file in its directory with no SPEC.

🤖 automated · Fable 5, effort xhigh

PR #1710, opened 2026-08-26
Only a merged or closed pull request is landed; unread and idle are visited once on change, awaiting again on age (#1707)

suleimansh:
Implements items 1 and 3 of #1707 as decided in https://github.com/framework/the-framework/issues/1707#issuecomment-5417327927.

**Item 1 — `content.js`.** A row showing only a pull request is `landed` only for `#N · Merged` / `#N · Closed`; `#N · Open` is `idle`. A status word on the row still wins. Any other pull-request wording stays `unknown` with the label carried, as before for labels the bridge does not know.

**Item 3 — `driver-plan.js`.** `PARKED` + one `REVISIT_MS` becomes two sets: `ON_CHANGE = {awaiting, unread, idle}` (visited when never seen or when the list's word changed) and `ON_AGE = {awaiting}` (also visited again after five minutes unchanged). A queued answer still forces a visit whatever the status. `running`, `landed`, `missing` are never visited on their own.

Note on `idle`: the decision says an open-PR session "is visited on change, never on age". On main the planner never visits `idle` at all, so honouring that puts `idle` — the list's own "Idle" rows included — in the on-change set. Cost: one visit per session when it turns idle, and on a fresh worker (empty `seen`) idle sessions join the first cycles' backlog, four visits a beat. If you'd rather keep `idle` unvisited, drop it from `ON_CHANGE` and item 1 becomes a report-only change (`idle` instead of `landed` on the dashboard).

**check.mjs** — three cases, each red on main before the fix and each verified to go red again when its rule is reverted:
- an open pull request alone reads as `idle`, a merged or closed one as `landed`
- awaiting/unread/idle are visited when never seen or on a status change; running/landed/missing are left alone
- after five minutes unchanged only `awaiting` is visited again; `unread` and `idle` are not

Specs: `content.SPEC.md`, `driver-plan.SPEC.md` (glossary "parked" → "stopped", now including idle), `check.SPEC.md`, the extension `SPEC.md`, `FEATURES-SPEC.md:170`.

Nothing in `dist/` changes — the daemon needs no rebuild; the extension needs a reload.

🤖 automated · Fable 5, effort high

PR #1712, opened 2026-08-26
The extension reloads itself when its files on disk change (#1711)

### LINKED ISSUE #1711: The extension could reload itself when its files change
Every change to the unpacked extension (`content.js`, `driver-plan.js`, `background.js`) needs a human click on chrome://extensions → reload, and nobody else can do it: `chrome://` pages are off-limits to any extension, and the daemon has no handle on Chrome. Every dogfood of a bridge change went through that click (#1703, #1709, #1710).

An extension *can* reload itself: `chrome.runtime.reload()` re-reads an unpacked extension's files from disk, the same as the button. What is missing is the trigger — something has to notice the files changed. Options:

1. **The worker watches its own files.** Each alarm beat (30 s) it fetches `chrome.runtime.getURL('content.js')` and the other scripts, hashes the text, compares with the hash taken at worker start; a difference → `chrome.runtime.reload()`. No daemon involvement, ~20 lines. To spike first: whether that fetch returns the *current* disk file for an unpacked extension or a cached copy.
2. **The daemon says so.** The worker already calls the daemon every beat and already compares a `version` field (0.12.0 expected). The daemon would serve a fingerprint of `packages/chrome-extension/` and the worker reload on a mismatch. Cleaner semantically ("the checkout moved"), but the daemon runs from `dist/` and would need to know where the checkout is.
3. **A dev-only watcher process** (`web-ext run` style, or the "Extensions Reloader" pattern): a separate process on the developing machine only.

After the reload, the existing path takes over: the Driver tab gets the new content script on its next page load (≤ 1 minute), as it does after a manual reload today.

"Update" in the Web Store sense does not apply: the extension is unpacked and dogfood-only; the store's auto-update covers it once published.

Recommendation: option 1 behind the spike; option 2 if the fetch turns out to be cached.

🤖 automated · Fable 5, effort high

suleimansh:
Closes #1711.

Every change to the unpacked extension needed a click on chrome://extensions before Chrome re-read its files (#1703, #1709, #1710 each went through that click). Now the extension reloads itself: the service worker takes a fingerprint of the extension's own files when it starts and again every beat (30 s), and calls `chrome.runtime.reload()` when any of them changed on disk — never while a cycle is running, and recorded as the last cycle's outcome naming the changed files. The Driver tab's content script is orphaned by the reload as by a manual one, and the next cycle's page load replaces it.

**The spike, measured 2026-08-26 in Chrome for Testing 150 with a throwaway extension (unattended, no click):**

- `fetch(chrome.runtime.getURL('probe.js'))` from the worker returns the **current** disk file, same worker, plain fetch — no cache to defeat. Option 1 from the issue, as recommended.
- `chrome.runtime.reload()` from the worker boots a new worker on the edited files (verified by the new worker's baseline hash being the edited file's).
- One thing found on the way: with developer mode switched **off** on chrome://extensions, Chrome (137+) *disables* an unpacked extension on reload — `DISABLE_UNSUPPORTED_DEVELOPER_EXTENSION` — instead of reloading it, and only a click there brings it back. Developer mode is on for anyone who loaded the extension unpacked; the README now says to leave it on. (A `--load-extension` command-line install is disabled the same way regardless, so the daemon-owned Chrome for Testing of #1332 will need the CDP `Extensions.loadUnpacked` install plus developer mode on for this to work there.)

**What changed**

- `fingerprint.js` (new): the watched files — every file Chrome loads for the extension: `manifest.json`, `background.js` and what it imports, `content.js`, `options.html` and `options.js` — hashed (FNV-1a) and compared; a file that cannot be read fails the fingerprint rather than counting as a change, so a transient read failure (a file mid-write, a checkout mid-switch) cannot reload in a loop.
- `background.js`: the alarm now runs a *beat* — the file check, then the cycle. A beat landing on a running cycle does nothing; the next one checks again. The reload is noted as the last cycle's outcome (`reloading the extension: content.js changed on disk`).
- `check.mjs`: two cases, written red first and break-checked three ways (an always-empty comparison, an unreadable file counted as a change, a file dropped from the watched list — each turned the right case red): the watched list equals what the manifest, `importScripts` and the options page load — nothing missing, nothing extra — and unchanged/edited/unreadable behave as above.
- Specs: `fingerprint.SPEC.md` (new), `background.SPEC.md`, the extension `SPEC.md`, `check.SPEC.md`; `FEATURES-SPEC.md` gets the feature; the README's "reload the extension on chrome://extensions" paragraph is rewritten.

**Not covered here**: the user's own claude.ai tabs still run the old content script until they reload — as after a manual reload; only the Driver tab is the extension's to reload.

**Dogfood**: one last manual reload to get this worker running, then an edit to a comment in `content.js` and the Driver's `page loaded on /code` self-report arriving with no click — posted below when done.

### DIFF
--- a/FEATURES-SPEC.md
+++ b/FEATURES-SPEC.md
@@ -168,6 +168,7 @@ happens while nobody is at the keyboard.
 - A web run's cloud session is created by the Chrome extension in your own browser, through claude.ai's repository picker — repo-bound, so it can push and open its pull request; without the extension (or with the browser bridge off) the run stops and says which is missing
 - Chrome extension bridging claude.ai questions back to the dashboard
 - One pinned "The Framework Driver" tab serves every recent cloud session: it reads claude.ai's own session list (the status beside each session), visits a session when the list's word for it changed to awaiting input, unread or idle, an awaiting one again every five minutes, and any session holding a queued answer — navigating inside the app, one page load a minute — and shows a full-page overlay naming what it is, with collapsible debug logs; closing the tab pauses the bridge until the extension's options page reopens it or the browser restarts
+- The extension reloads itself when its files change on disk: an edit in the checkout is running within half a minute, never mid-cycle, with no trip to chrome://extensions
 - A cloud run's row says "waiting" when claude.ai's session list shows its session awaiting input, even when the question was asked in prose rather than as a choice block
 - A cloud session's conversation mirrored into the run view, turn by turn, as it is written
 - Answer a cloud agent's question from the dashboard (typed back into claude.ai) — the same gate panel a local agent gets, multi-select and stop options included, listed with every other open question
--- a/packages/chrome-extension/README.md
+++ b/packages/chrome-extension/README.md
@@ -70,7 +70,11 @@ the Driver: reading the session list by label and paging it, a cycle that visits
 the question, types and waits for the send, then returns to the list under the overlay, an answer
 the page did not take, a session missing from the list, and the visit planner.
 
-After editing any file here, reload the extension on `chrome://extensions` AND reload the open
-claude.ai tabs: reloading the extension does not re-inject content scripts, and an orphaned
-script cannot hear the new worker. The panel shows the manifest version, which is how you tell
-a stale script from a current one.
+After editing any file here, nothing to click: the worker fingerprints the extension's files
+every beat (30 s) and reloads the extension itself when any changed, never mid-cycle
+(#1711). The Driver tab gets the new content script on the next cycle's page load; your own
+claude.ai tabs still need a reload, since reloading the extension does not re-inject content
+scripts and an orphaned script cannot hear the new worker. The panel shows the manifest
+version, which is how you tell a stale script from a current one. Leave developer mode on in
+`chrome://extensions`: with it off, Chrome 137+ disables an unpacked extension on reload instead
+of reloading it, and only a click there brings it back.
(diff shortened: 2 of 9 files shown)


suleimansh:
Dogfood, on the real extension in Suleiman's Chrome, after one last manual reload (the one that put this branch's worker in place). Every timestamp below is the daemon's own (`onBridgeStatus`), UTC; nobody touched chrome://extensions after the reload.

**Edit 1 — a comment in `content.js`, 02:16:09.** The next beat saw it and reloaded the extension:

- 02:16:32 `/_bridge/hello` — `rows 0, no transcript rows found`: a fresh content script in the Driver tab, loaded by the new worker's first cycle (its page refresh is due at once after a worker start)
- 02:16:36 `/_bridge/hello` — `rows 10 (human,marker,assistant), turns 6, sent 6 of 6 turn(s)`: the fresh script re-mirrors the whole session (the old one was sending increments, `sent 1 of 1`)
- 02:17:58 `/_bridge/hello` — `page loaded on /code`

That alone does not tell a reloaded *extension* from a reloaded *page* — the worker reloads the Driver page once a minute anyway — so:

**Edit 2 — the hello note itself, 02:19:08:** `page loaded on ${location.pathname}` → `page loaded on ${location.pathname} [self-reloaded #1711]`, on disk only.

- 02:19:31 / 02:19:35 — same fresh-script pair as above (the reload)
- **02:20:29 `/_bridge/hello` — `page loaded on /code [self-reloaded #1711]`**

A string that existed only on disk reached the daemon 81 seconds after it was written: one beat (≤30 s) to notice and reload, then the Driver's next page load. Both edits are reverted; the revert is itself a third reload, which is the feature working as intended.

PR #1716, opened 2026-08-26
A project whose directory is gone leaves the sidebar (#1140)

### LINKED ISSUE #1140: What happens when removing (or renaming) a directory
Seems like TF doesn't refresh the list (but should).

Low-prio, post-MVP.

### LINKED ISSUE #1142: Bug: `Files` missing
Low-prio, post-MVP.

I suspect it's related to:
- https://github.com/gemstack-land/the-framework/issues/1140

Image

suleimansh:
Fixes #1140. Fixes #1142.

## What

A registered repo whose directory was renamed or deleted leaves the Projects sidebar on the next refresh (the projects poll, 30s), and comes back when the directory does. Before, it stayed as a grey "not activated" entry with no files — the `Files` tab hides when the file list is empty, and `git ls-files` in a missing directory yields an empty list, which is #1142 — indistinguishable from a repo that merely lost its `.the-framework/` marker, and with nothing to click to make it go away. A URL naming the vanished project now lands on the existing "not registered" view.

The registration is kept: the record is skipped, not pruned, so a renamed-back folder or a remounted volume needs no re-adding.

## How

- `src/dashboard/projects.ts`: `defaultProjectsProvider` filters the registry records through a directory stat (`nodeFs().isDirectory`) before both `list()` and `resolvePath()`; a failing stat counts as gone. Readers are injectable (`ProviderDeps`) so the provider is testable off the user's registry.

Left alone: the daemon's own `listProjects` callers (boot reconcile, bridge session list, Discord summaries) — each already tolerates a missing path per project, and none of them is what the user sees.

## Tests

`projects.test.ts`: gone → not listed; gone → no path resolved (present one still resolves); back → listed again; a throwing directory check → gone, no error. Broken on purpose (filter removed): all four fail, the rest pass.

## Specs

`src/dashboard/projects.SPEC.md` (new "Vanished project" section + TL;DR line), `projects.test.SPEC.md`, `FEATURES-SPEC.md` (Projects sidebar).


suleimansh:
Dogfooded on this branch's build with an isolated daemon (`XDG_CONFIG_HOME` registry, `:4321`, two throwaway repos registered):

- **Daemon:** `onProjects` lists both → `mv dogfood-b dogfood-b-renamed` → lists `dogfood-a` only, `onProjectFiles(b)` = `[]` → `mv` back → both again, files back → `rm -rf dogfood-b` → `a` only. The registry file still holds both records throughout (skipped, not pruned). No daemon-log errors.
- **Dashboard, no reload:** sidebar showed both → renamed on disk → `dogfood-b` gone on the next 30s poll → renamed back → back on the next poll.
- **The #1142 case:** opening `/dogfood-b-9x0bv6` while the directory is renamed lands on the existing "No such project — It may have been removed" view instead of a grey, file-less project page.

PR #1722, opened 2026-08-26
The safety commit refuses an implausible sweep (#1638)

### LINKED ISSUE #1638: The catch-all safety commit will commit anything: 7,632 cache files went to main unnoticed
🤖 *automated*

Found while cleaning up #1632, and the cleanup there only removes the instance.

## What happened

The framework has a safety commit that preserves whatever is sitting in a checkout before it does something that could lose it. It runs `git add -A` and commits with a fixed message, in two places:

- `packages/the-framework/src/install.ts:48-52` — activation, so the install commit lands on a clean tree
- `packages/the-framework/src/store/worktree.ts:152-155` — `commitPendingWork`, before a worktree is folded away

On 2026-08-18 one of them ran in a checkout that happened to contain turborepo's local build cache:

```
118e6cad  2026-08-18 18:15  [The Framework] uncommitted changes
7634 files changed, 5097 insertions(+)
```

7,632 of those files were `.turbo/cache/`. The repo went to ~664 MB, of which the actual project — all source, all docs, every commit ever made — is ~115 MB. Nobody noticed for four days, and the way it surfaced was a clone taking several minutes.

## Why this is not a turbo problem

`.turbo/` is now ignored, so this exact file never repeats. The mechanism is untouched: **a commit path that adds everything it finds is only ever as safe as `.gitignore` is complete.** The list of things a modern toolchain drops into a working tree that nobody wants in git is open-ended — `.next/`, `coverage/`, `target/`, `.venv/`, `.cache/`, a stray database dump — and the framework will commit any of them, at any size, without a word.

It is also worth saying plainly that this is the framework doing it *unattended*. A human running `git add -A` sees the file count scroll past. This path has no reader.

## What to do about it

**Option A — refuse implausible sweeps.** Before committing, count what is pending; past a threshold (files or bytes) stop, commit nothing, and report what was seen. A real session adds a handful of files, never thousands, so the threshold separates the two cases cleanly and needs no knowledge of which tool made the mess. Catches the next unknown tool as well as this one.

**Option B — a built-in denylist.** Skip paths matching a known set of machine-local directories (`.turbo/`, `.next/`, `coverage/`, …). Fixes the shapes we can name today and ages badly: it is a list somebody has to keep current, and it says nothing about the artifact nobody has thought of yet.

I would take **A**. It is a property of the situation ("this is far more than a session's work") rather than of the tooling, so it does not need maintaining, and a refusal that reports what it saw is strictly more informative than a silent 262 MB commit.

Not started — this is a question about what the framework should do to a user's repo, so it wants your call first.

suleimansh:
Closes #1638. Option A, as announced on the issue.

## What

The `[The Framework] uncommitted changes` commit — at activation (`install.ts`) and when an agent's worktree is torn down (`store/worktree.ts` `commitPendingWork`) — used to `git add -A` whatever it found. Both now go through one module, `safety-commit.ts`, which counts what would be staged first:

- `git status --porcelain -uall -z` — every untracked file counted individually (`?? .turbo/` was one line for 7,632 files), raw NUL-separated paths so spaces and non-ASCII names cannot break the count, a rename counted once.
- Past **200 files or 20 MB**: nothing is staged, and the step fails with the report — `refused to commit 7,632 pending files as "[The Framework] uncommitted changes": far more than a session leaves behind (the limit is 200 files or 20 MB), mostly under .turbo/ (7,630 files), the repository root (2 files). Commit or ignore them yourself, then retry.`
- Sizes are read from disk only while the count is within its limit; past it the count alone refuses, and thousands of pending files is exactly where thousands of stats would cost.

Where the refusal lands:

- **Activation** fails with the report as its error and writes no marker, so the next attempt runs the check again.
- **Teardown** treats it as final — not a lost `index.lock` to retry (#1376) — logs `[framework] <checkout>: refused to commit …`, and returns false so the checkout is kept. The three callers' boolean contract is unchanged; the reason is in the daemon log because "could not be committed; its worktree was kept" cannot carry it.

## Why a limit and not a denylist

It is a property of the situation ("far more than a session's work"), not of the tooling, so it catches the next unknown cache directory too and there is no list to keep current. `.turbo/` is already ignored; this is about the mechanism.

## Verified

Full suite green (1,649 Node + 864 dashboard; 10 new). Broken on purpose in the compiled output: "never refuse" turns 5 tests red across all three files, including the real-repo one (a cache directory past the limit stays fully pending, the same tree within a bigger limit is committed whole); "retry the refusal like any failure" turns exactly the no-retry test red.

Specs: new `safety-commit.SPEC.md` (+ test spec); `install`, `store/worktree` and their test specs updated; FEATURES-SPEC's "Commit what the agent left uncommitted" line carries the exception.

🤖 *curated · Fable 5, effort high*


suleimansh:
Superseded: Rom chose to remove the safety commit altogether (https://github.com/framework/the-framework/issues/1638#issuecomment-5440340542), so a limit on it has nothing left to guard. The removal PR replaces this one.

Issue #1725, opened 2026-08-27
Branch management: the first skill (plan)


suleimansh:
Plan for the first skill of skills-plus (#1681): worktree management, called **branch management**. Nothing built yet; this is for reading before code.

## 1. The skill

One npm package, `@superskill/branch-management`, shipping three things: `SKILL.md`, a CLI, and the same functions as an API. Per your skills-via-npm idea: the skill is versioned and pinned like any dependency, and `SKILL.md` reaches the agent from `node_modules`.

What `SKILL.md` tells the agent (replaces the *Workspace* and *Session name* sections of `system_prompt.md`):

- You are started inside your own checkout, `.the-framework/branches/tf-agent-<id>/`. Everything you read or write is under it. The repo around it is the user's own working tree; never touch it.
- Name the session: pick `<name>` (`[a-z0-9-]+`), run `branch-management name <name>`. It renames your branch to `tf-<name>` (a rename, not a new branch) and prints the final name — it may differ if the name was taken.
- Commit as you go, on that branch. The framework publishes only what you committed and never commits for you (#1638).
- Before you finish: `branch-management status` must report a clean tree. Uncommitted work stays in your checkout and blocks its reclaim.
- Do not push and do not open the pull request yourself; emit `open-pr`. Publishing is the dashboard's, per handoff level.

Not the agent's job: creating the checkout. An agent that creates its own would start inside the user's checkout, which #736 and #997 exist to prevent, and only Claude Code can switch its cwd mid-session. So the launcher (daemon or CLI) creates it before the process starts — through the same package. The skill states that requirement the way a skill says "make sure X is running".

## 2. The package

Git operations, as a CLI for the agent and an API for the dashboard; one implementation, every surface a caller:

| command | does | today |
|---|---|---|
| `create <id> [--base <ref>]` | worktree + branch `tf-agent-<id>`, dependency dirs linked in | `addWorktree`, `linkDependencies` |
| `attach <id> <branch>` | re-checkout for a continued agent | `attachWorktree` |
| `name <name>` | rename the current branch to `tf-<name>`, refresh the `branches/tf-<name>` link | `renameAgentBranch` (dead today: the prompt has the agent branch itself) |
| `status [path]` | branch, clean?, on the remote? | `worktreeClean`, `branchPushed`, `currentBranch` |
| `list [--sizes]` | every checkout under `branches/` as JSON | `listWorktrees`, `listProjectWorktrees` (git half) |
[...]

PR #1730, opened 2026-08-27
Comment the branch-management bin as a forwarder to dist/cli.js

suleimansh:
`packages/branch-management/bin/branch-management` now carries a one-line comment right after the shebang stating that the file only forwards to `dist/cli.js`. No behavior change.

Opened from The Framework session `bin-forward-comment`.


suleimansh:
Dogfood of #1725 step 4 — the skill worked end to end; closing the artifact.

PR #1773, opened 2026-09-09
Install LDD: a LOGIC.md beside every source file and directory

brillout:
## What

Installs Logic-Driven Development per the `logic-driven-development` skill: a `LOGIC.md` describing the business logic of each file and directory containing source code, written for the technical product manager who never reads the code.

- `some-file.ext` → `some-file.LOGIC.md`, `some-dir/` → `some-dir/LOGIC.md`, across the whole monorepo: the product, the driver seam, the `agent-data` library, the four skills, the Chrome extension, the website, the prompts, the GitHub workflows.
- Written fresh from the code, not converted from any earlier documentation. Every file was read in full; a test file gets a "What the tests cover" list of the behaviors its tests pin, never how they work.
- One vocabulary across every file (agent, coding agent, driver, turn, gate, pick, handoff, checkout, location, the `agent-data` branch, sweep, routine, quota boundary, …), each file carrying a glossary of the coined terms it uses.
- A directory's `LOGIC.md` tells its subsystem's story and names every file under it exactly once. Files with no business logic (build and test-runner configuration, styling, static assets) are named there instead of getting one of their own.

## Coverage

| | |
|---|---|
| source files documented | 637 |
| directories documented | 45 |

## Notes

No source file is modified. Two other changes make the installation coherent: the prompt compiler skips `LOGIC.md` beside a prompt as it skipped `SPEC.md` before it, and the `framework` package's README points readers at `LOGIC.md` instead of files that no longer exist.

https://claude.ai/code/session_01YcSwKgtQE6ndJYjdDitVFB


brillout:
The `build` check failed once on commit 321dcb8 and is not being fixed here, because the failure is not this branch's.

- That commit's diff over its parent is 37 `LOGIC.md` files and nothing else: no source, config or test file changed, so no test behavior could change with it.
- The same `build` check passes on the current head, whose tree contains those 37 files plus more, and `pnpm build && pnpm typecheck && pnpm test` passes locally on this branch.
- The failing run reported `fail 1` alongside `cancelled 1` with a total suite duration of 60.4s against the suite's own 60s per-test timeout, which is the shape of a timeout on a loaded runner rather than an assertion.

No re-run was spent on it: the head has since moved on, and the later green run of the same check on a superset of the same files is the confirmation a re-run would have given. No fix exists to port, and none is needed. I am still watching this pull request and will act on any failure on a current head.

---

brillout:
Correcting one sentence in my earlier comment on this pull request: I wrote that `pnpm build && pnpm typecheck && pnpm test` passes locally on this branch. Build and typecheck do; the test suite does not. Two tests fail in the sandbox I ran them in, both asserting that a process was reaped:

- `browser.test.ts` — "the real sweep kills a browser process init inherited, and spares one a live agent holds"
- `daemon-runtime.test.ts` — "terminate reaps the whole process group of an agent that ignores SIGTERM, browser included"

Both fail on process-group semantics this sandbox does not reproduce, and one timing-sensitive daemon-clock test failed once more under load. They are not this branch's: apart from `LOGIC.md` files the branch changed only the prompt compiler's exclusion of `LOGIC.md` and a README pointer, and the `build` check passed on GitHub's runners for these same commits.

The conclusion of that comment is unaffected — it rested on the documentation-only diff and on the later green runs of the same check — but the evidence line was wrong as written.

---

PR #1775, opened 2026-09-10
Twelve fixes the LOGIC.md pass turned up: three defects, four dead things, five stale claims

brillout:
Writing a `LOGIC.md` beside every file meant reading every file, and reading every file turned up things that were wrong. Each is one commit.

## Defects

- **An action that succeeds with nothing is not a failure.** `useAction` reported a refusal and a void success alike as `undefined`. Three call sites worked around it by mapping their success to a stand-in `true`; the confirm dialog, which had no such wrapper, would have sat open with no message after a successful delete. `run` now says whether the action succeeded and carries its value, with the refusal branch excluded from that value's type.
- **Escape closes the suggestion menu.** Escape over an open `/`, `<`, `@` or `#` menu was passed through untouched, so the menu stayed on screen — and since an open menu is what makes Enter pick instead of send, the next Enter picked a suggestion the user thought they had dismissed.
- **A tag is a tag whether it was typed or loaded.** Typing `<await>` produced the `<AWAIT>` the agent expects; loading a preset carrying the same text left it as prose and sent it lower-case. Both paths now recognize the same set. Tests pin it, which the token catalogue had none of.

## Things nothing could reach

- **The preview event.** Declared, rendered by the terminal, pinned by a test, emitted by nothing, and hung on a serve config that does not exist.
- **The gate countdown.** A gate carried a delay after which the dashboard was to accept the recommended option, the panel took a prop saying whether that may run here, two call sites turned it off, and a mouse-move listener stood by to cancel it. Nothing ever started a countdown. Gone with it: the pick author no surface could produce, so a pick is the user's or nobody's. **Whether the dashboard should answer a gate for the user at all is a product question this leaves open rather than half-built.**
- **The `reference` token kind**, which no token has ever been.
- **The `sidebar_state` cookie**, written "so it survives a reload" and read by nothing.

## Claims that had outlived their code

- **A budget cap.** Twenty-odd comments named one as a way an agent ends; the usage event even said the agent stops itself at one. The flag that set it is gone, a test pins that it is gone, the meter only adds up, and `MEMORY.md` says a running agent is never cut short over spending.
- **The session name as a turn signal**, in the emitter's own doc, the backlog loop and a test title — while the same file's header says it is not one.
- **A triage that aborts when its branch exists.** The prompts pin a session name and abort on nothing; the guard is the routine lock the daemon takes before starting one.
- **Four smaller ones**: a per-agent guard that "exists to pause an agent", one agent at a time per project, a queue parser named after a function that no longer exists, and a design gallery upload nothing performs.
- **The vocabulary in the prose a reader meets first**: both package READMEs and the gallery's sample copy still said session, worktree and "Open TODOs".

## Checks

`pnpm build`, `pnpm typecheck` and `pnpm test` all run. One test fails in my sandbox and is not this branch's: `reclaim.test.ts` makes a directory read-only and expects removal to fail, which root ignores — this container runs as root, and the branch touches no file in that package.

Every `LOGIC.md` describing changed behavior was updated in the same commit, and the coverage and vocabulary lint over all 683 of them reports nothing.

https://claude.ai/code/session_01YcSwKgtQE6ndJYjdDitVFB

---
_Generated by [Claude Code](https://claude.ai/code/session_01YcSwKgtQE6ndJYjdDitVFB)_

### DIFF
--- a/packages/framework/README.md
+++ b/packages/framework/README.md
@@ -3,9 +3,9 @@
 **The Framework** — autonomous AI programming: humans make the important decisions
 while coding agents run unattended.
 
-You register your repos. From then on, agents work on them in sessions: each session
-gets a throwaway copy of the repo, does its work, and hands the result off as a pull
-request. Your own checkout is never touched.
+You register your repos. From then on, agents work on them: each agent gets a throwaway
+checkout of the repo, does its work, and hands the result off as a pull request. Your own
+checkout is never touched.
 
 ```bash
 npm i -g framework
@@ -18,7 +18,7 @@ the-framework      # serves the dashboard at http://127.0.0.1:4200
 
 ```
 the-framework          Serve the dashboard in the foreground. Ctrl+C closes it and
-                       every session it is running.
+                       every agent it is running.
 
   --port <n>           Dashboard port (default: 4200).
   --host <addr>        Bind address (default: 127.0.0.1). A non-loopback address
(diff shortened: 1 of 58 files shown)


brillout:
@suleimansh Good? I didn't review any of it (it's a list of issues Claude found while installing LDD).

suleimansh:
Seems good to me. The fixes were checked by my agent, commit by commit, they are real. The full test suite passes on my machine on this branch (1378 + 870 tests, the reclaim test included).

I didn't read all of the LOGIC files, I read only a few lines to check if they are easy to understand, and yes they are. Some files have very long texts, we can improve later.

